PR #15, opened 2019-11-14
Introduce tests

andreban:
- Adds Jasmine as a Developer Dependency
- Adds tests for JdkHelper


andreban:
> At a glance it all looks good to me. I can have a closer look at some parts if you want.

No need. I think it's good. Thanks for checking it out.

PR #25, opened 2019-11-15
Asks user for a suitable image when one is not available on the Web Manifest

### LINKED ISSUE #12: Building fails if the icon is less than 512x512
If you run the tool on a webapp that doesn't have an icon larger than 512x512, it won't work.

I don't think icons of that size are currently required for anything else, and I've found two major webapps that don't have such an icon (Pinterest and Instagram).

It would be good if we could either remove the need or let the user provide another image during `llama-pack init` instead of having to change their webapp manifest.

andreban:
Closes #12

### DIFF
--- a/cli/cmds/init.js
+++ b/cli/cmds/init.js
@@ -23,6 +23,7 @@ const {promisify} = require('util');
 const util = require('../../lib/util');
 const colors = require('colors/safe');
 const prompt = require('prompt');
+const validUrl = require('valid-url');
 prompt.get = promisify(prompt.get);
 
 // Regex for disallowed characters on Android Packages, as per
@@ -35,7 +36,7 @@ function _generatePackageId(host) {
   return parts.join('.').replace(DISALLOWED_ANDROID_PACKAGE_CHARS_REGEX, '_');
 }
 
-async function createTwaConfig(manifestUrl, manifest) {
+async function createTwaConfig(manifestUrl, manifest, icon, maskableIcon) {
   const fullStartUrl = new URL(manifest['start_url'], manifestUrl);
   prompt.message = colors.green('[llama-pack-init]');
   prompt.delimiter = ' ';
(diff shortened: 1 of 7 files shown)


andreban:
@NotWoods I think this affects your ongoing work on Adaptive Icons, but seems like a move in the right direction.

NotWoods:
I can probably add in an optional prompt for the maskable icon, following the same format.

andreban:
Sounds good to me.

andreban:
@NotWoods rebased with master and added a prompt for the maskable icon. I tested and things seems to work as intended. Do you mind taking a quick look if the maskable bits look right?

cli/cmds/init.js
```diff
@@ -77,6 +78,21 @@ async function createTwaConfig(manifestUrl, manifest) {
...
+        name: 'iconUrl',
+        description: 'URL to an image that is at least 512x512px',
+        message: 'Must be a well-formed http or https URL.',
+        required: true,
+        default: icon ? new URL(icon.src, manifestUrl).toString() : '',
+        conform: validUrl.isWebUri,
+      },
+      maskableIconUrl: {
+        name: 'maskableIconUrl',
+        description: 'URL to an image to be used when generating maskable icons',
```
NotWoods: It might be worth mentioning that this icon also has to be at least 512x512px

spec/lib/utilSpec.js
```diff
@@ -0,0 +1,96 @@
...
+          'type': 'image/png',
+        }],
+      }, 'any');
+      expect(result.src).toBe('/favicons/android-chrome-512x512.png');
+      expect(result.sizes).toBe('512x512');
+      expect(result.type).toBe('image/png');
+      expect(result.size).toBe(512);
+    });
+
+    it('returns null when not maskable icon is available', () => {
```
NotWoods: nit: 
```suggestion
    it('returns null when no maskable icon is available', () => {
```

Issue #11, opened 2019-11-12
Store configuration used to initialize the project on a TWA Manifest on a file


andreban:
Currently, llama-pack uses a Web Manifest to extract data, asks the user to confirm the values and generates `build.gradle` and `app/AndroidManifest` from it.

This would introduce an intermediate configuration file, with a format similar to [Web Manifest](https://www.w3.org/TR/appmanifest/).

The configuration will be used when applying updates to the template project to an existing project, as described on #3.

Developers can also use it to tweak configurations for their project an re-generate the project from it.

@FluorescentHallucinogen FYI

andreban:
Closed by #29

PR #29, opened 2019-11-23
Adds TwaManifest and enables update

andreban:
- Adds TwaManifest and test
- Adds an update CLI option, to rebuild a project based on an
  existing Twa Manifest
- Skips generating the Signing Key on Init, if the file already
  exists

### DIFF
--- a/cli/cmds/build.js
+++ b/cli/cmds/build.js
@@ -18,6 +18,7 @@
 
 const {androidSdkTools} = require('../../lib/androidSdk');
 const GradleWraper = require('../../lib/GradleWrapper');
+const TwaManifest = require('../../lib/TwaManifest');
 
 const {promisify} = require('util');
 const prompt = require('prompt');
@@ -30,6 +31,7 @@ async function build() {
     await androidSdkTools.installBuildTools();
   }
 
+  const twaManifest = await TwaManifest.fromFile('./twa-manifest.json');
   prompt.message = colors.green('[llama-pack-build]');
   prompt.delimiter = ' ';
   prompt.start();
@@ -59,9 +61,9 @@ async function build() {
   console.log('Signing...');
   const outputFile = './app-release-signed.apk';
   await androidSdkTools.apksigner(
-      './android.keystore', // the path to the keystore file
+      twaManifest.signingKey.path,
       result.password, // keystore password
-      'android', // alias
+      twaManifest.signingKey.alias, // alias
       result.password, // key password
       './app-release-unsigned-aligned.apk', // input file path
       './app-release-signed.apk', // output file path
--- a/cli/cmds/build.js
+++ b/cli/cmds/build.js
@@ -36,14 +36,27 @@ async function build() {
   prompt.delimiter = ' ';
   prompt.start();
 
+  const schema = {
+    properties: {
+      password: {
+        name: 'password',
+        required: true,
+        description: 'KeyStore password:',
+        hidden: true,
+        replace: '*',
+      },
+      keypassword: {
+        name: 'keypassword',
+        required: true,
+        description: 'Key password:',
+        hidden: true,
+        replace: '*',
+      },
+    },
+  };
+
   // Ask user for the keystore password
-  const result = await prompt.get({
-    name: 'password',
-    required: true,
-    description: 'Password for the Key Store',
-    hidden: true,
-    replace: '*',
-  });
+  const result = await prompt.get(schema);
 
   // Builds the Android Studio Project
   console.log('Building the Android App...');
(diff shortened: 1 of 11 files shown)


andreban:
Closes #11

lib/TwaManifest.js
```diff
@@ -0,0 +1,197 @@
...
+    return TwaManifest.fromWebManifestJson(webManifestUrl, webManifest);
+  }
+
+  /**
+   * Loads a TWA Manifest from the file system.
+   *
+   * @param {String} fileName the location of the TWA Manifest file
+   */
+  static async fromFile(fileName) {
+    const json = JSON.parse(await fs.promises.readFile(fileName));
```
PEConn: Just to check my understanding - if readFile throws an error (since the file doesn't exist), that error will be thrown from here?

PEConn: Also, do we want some slightly nicer error handling if the file has invalid JSON?

lib/TwaManifest.js
```diff
@@ -0,0 +1,197 @@
...
+
+  backgroundColorHex() {
+    return this._hexColor(this.backgroundColor);
+  }
+
+  /**
+   * Creates a new TwaManifest, using the URL for the Manifest as a base URL and uses the content
+   * of the Web Manifest to generate the fields for the TWA Manifest.
+   *
+   * @param {URL} webManifestUrl the URL where the webmanifest is available
```
PEConn: nit: missing full stop.

cli/cmds/build.js
```diff
@@ -30,18 +31,32 @@ async function build() {
     await androidSdkTools.installBuildTools();
   }
 
+  const twaManifest = await TwaManifest.fromFile('./twa-manifest.json');
   prompt.message = colors.green('[llama-pack-build]');
   prompt.delimiter = ' ';
   prompt.start();
 
+  const schema = {
```
PEConn: As a larger point (not for this CL), requiring the passwords to be entered as prompts makes it tricky to use this tool as part of a larger script. Filed #30 for this.

cli/cmds/init.js
```diff
@@ -16,28 +16,22 @@
...
-const util = require('../../lib/util');
 const colors = require('colors/safe');
 const prompt = require('prompt');
 const validUrl = require('valid-url');
+const fs = require('fs');
 prompt.get = promisify(prompt.get);
 
-// Regex for disallowed characters on Android Packages, as per
-// https://developer.android.com/guide/topics/manifest/manifest-element.html#package
-const DISALLOWED_ANDROID_PACKAGE_CHARS_REGEX = /[^ a-zA-Z0-9_\.]/;
```
PEConn: Where has all the data validation gone?

lib/jdk/KeyTool.js
```diff
@@ -0,0 +1,113 @@
...
+        },
+      },
+    };
+    const result = await prompt.get(schema);
+
+    // Execute Java Keytool
+    const keytoolCmd = [
+      'keytool',
+      '-genkeypair',
+      `-dname "cn=${result.cn}, ou=${result.ou}, o=${result.o}}, c=${result.c}"`,
```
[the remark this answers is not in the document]
andreban: I think we need to harden input validation globally. File #31 to track this.

cli/cmds/build.js
andreban: +1

lib/TwaManifest.js
andreban: If `readFile` fails, an error will be throw. The same thing happens if `JSON.parse` receives and invalid JSON.  Friendly errors are also something we need to look from a more general perspective. Filed #32 to track this.

cli/cmds/init.js
andreban: This bit was moved to `TwaManifest.js.`. This wasn't used to validate user input, but the replace invalid characters when generating a Package ID from the host. 

There's also a `validate()` method inside `TwaManifest`, which has the code moved from `TwaGenerator`. This also needs more work, but a subject for another PR.

andreban: Created #33 to track this.

lib/jdk/KeyTool.js
```diff
@@ -0,0 +1,113 @@
...
+    this.jdkHelper = jdkHelper;
+  }
+
+  /**
+   * Creates a new signing key.
+   *
+   * @param {path} keyPath where the generated key will be saved
+   * @param {String} alias the alias to be used for the key
+   */
+  async createSigningKey(keyPath, alias, overwrite = false) {
```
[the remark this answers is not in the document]
andreban: Renamed to `createSigningKeyIfNeeded `

lib/jdk/KeyTool.js
```diff
@@ -0,0 +1,113 @@
...
+        },
+        ou: {
+          name: 'ou',
+          required: true,
+          description: 'Organizational Unit (eg: Engineering Dept):',
+        },
+        o: {
+          name: 'o',
+          required: true,
+          description: 'Organization: (eg: Company Name)',
```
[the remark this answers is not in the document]
andreban: Fixed consistency. Added colons on lines 80 and 87.

lib/TwaManifest.js
andreban: done.

PR #40, opened 2019-11-28
Transforms Config into TypeScript

andreban:
- Also adds ConfigSpec.ts
- Adds Jasmine Types to Testing

### DIFF
--- a/src/lib/Config.ts
+++ b/src/lib/Config.ts
@@ -69,20 +69,17 @@ export class Config {
     return new Config(result.jdkPath, result.androidSdkPath);
   }
 
-  static async loadConfig(): Promise<Config | null> {
-    if (!await fileExists(CONFIG_FILE_NAME)) {
-      return null;
-    }
+  static async loadConfig(): Promise<Config> {
     const config = JSON.parse(await fs.promises.readFile(CONFIG_FILE_NAME));
     return new Config(config.jdkPath, config.androidSdkPath);
   }
 
   static async loadOrCreate(): Promise<Config> {
-    let config: Config | null = await Config.loadConfig();
-    if (config == null) {
-      config = await Config.createConfig();
-      await config.saveConfig();
+    if (await fileExists(CONFIG_FILE_NAME)) {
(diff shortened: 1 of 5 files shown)


src/spec/lib/ConfigSpec.ts
```diff
@@ -0,0 +1,28 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {Config} from '../../lib/Config';
+
+describe('config', () => {
+  describe('#constructor', () => {
```
PEConn: So you could use mock-fs to test out loading from file and such (I don't know much about it, just a cursory Google search).

src/lib/Config.ts
```diff
@@ -58,29 +65,24 @@ class Config {
...
-    const result = await prompt.get(schema);
-    Object.assign(this, result);
+    const result = await userPrompt.get(schema);
+    return new Config(result.jdkPath, result.androidSdkPath);
   }
 
-  async check() {
-    if (!await fileExists(this.jdkPath)) {
-      throw new Error(`${this.jdkPath} does not exist. Check the jdkPath on your config`);
+  static async loadConfig(): Promise<Config | null> {
```
PEConn: This is more a question - is there any reason you chose "Config | null" here instead of "Config?"? There's some info [here about it] - just curious if you had an opinion either way.

[here about it]: https://old.reddit.com/r/typescript/comments/76lk8y/interfaces_optional_property_vs_null_union_type/

andreban: Refactored to avoid having config is a nullable.

src/spec/lib/ConfigSpec.ts
andreban: I'll merge this and create another PR introducing a mock library. We need it for more thorough tests in other places too.

PR #41, opened 2019-11-29
Use custom jimp build

NotWoods:
Jimp recently released support for using parts of it independently. This PR shrinks the install size by removing all unneeded plugins.


PEConn:
Hey Tiger, thanks for this change!

Can you explain how it works? It seems you just set up some config - where does Jimp actually do stuff?

NotWoods:
This replaces the 'jimp' module that I previously added. Jimp is used for image resizing here but has many additional features, and the custom config strips those out and only downloads the resize plugin.

PR #42, opened 2019-11-29
Configure eslint to lint ts files

NotWoods:
I think ESLint only checks JS files by default. Adjusts the command to check ts files and adds the typescript parser. [@typescript-eslint/eslint-plugin](https://github.com/typescript-eslint/typescript-eslint) could be added for TS-specific rules.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,5 +1,6 @@
 {
   "extends": "google",
+  "parser": "@typescript-eslint/parser",
   "parserOptions": {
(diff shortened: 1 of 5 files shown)


PEConn:
Thanks for this! Will eslint with the TypeScript plugin do the same as TSLint or are there differences?

@andreban this LGTM, over to you.

NotWoods:
I can add the plugin in 🙂

PR #44, opened 2019-11-29
Transforms TwaManifest.js into typescript

andreban:
- Adds `@types/node-fetch`


src/lib/TwaManifest.ts
```diff
@@ -66,14 +66,29 @@ function generatePackageId(host) {
...
+  enableNotifications: boolean;
+  startUrl: string;
+  iconUrl: string;
+  maskableIconUrl: string;
+  useBrowserOnChromeOS: boolean;
+  splashScreenFadeOutDuration: number;
+  signingKey: string;
+  appVersion: string;
+
+  constructor(data: any) {
```
PEConn: Could you add a TODO comment to change `any` to something else once we've TypeScript-ified the rest of the code?

src/spec/lib/TwaManifestSpec.ts
```diff
@@ -45,9 +45,9 @@ describe('TwaManifest', () => {
       expect(twaManifest.iconUrl)
           .toBe('https://pwa-directory.com/favicons/android-chrome-512x512.png');
       expect(twaManifest.maskableIconUrl).toBeUndefined();
-      expect(twaManifest.themeColor).toBe('#7cc0ff');
-      expect(twaManifest.navigationColor).toBe('#7cc0ff');
-      expect(twaManifest.backgroundColor).toBe('#7cc0ff');
+      expect(twaManifest.themeColor.hex()).toBe('#7CC0FF');
```
PEConn: Would it be worth having each of the colours be different, so we can catch if we accidentally mix them up somewhere?

PR #46, opened 2019-12-04
Support WebManifest shortcuts when generating the TWA

rayankans:


src/cli/cmds/init.js
```diff
@@ -89,6 +89,15 @@ async function confirmTwaConfig(twaManifest) {
         default: twaManifest.maskableIconUrl || undefined,
         conform: validUrl.isWebUri,
       },
+      shortcuts: {
+        name: 'shortcuts',
+        validator: /y[es]*|n[o]?/,
+        message: 'Include app shortcuts?',
+        description: 'App shortcuts to display for users to users quickly start common or '+
```
PEConn: I think you've got one too many 'users' here.

rayankans: One might argue I have the exact amount of users needed

rayankans:
> What's the UX here? We're prompting to ask if the developer wants to include shortcuts and then what?

I had an offline conversation with @andreban about this. Currently, we just ask whether users want to include shortcuts or not. They can't set it themselves. If we do end up going down that path, we will need to explore some UI possibilities, as it's not that easy to set them up via a CLI. It's easier to pass a file manifest if you want to customize shortcuts in a way that differs from the published Web Manifest.

If they select yes and there are no shortcuts, nothing happens.

andreban:
Yes, filling a shortcuts section from the prompt might make things complicated (maybe we could make it optional later).

It's possible to add an `ask` attribute to the configuration and add a function that will only trigger the question if a `shortcuts` section exists in the Manifest.

I wonder if we should show developers the content of the existing shortcuts before asking them to confirm.

rayankans:
Makes sense, I added the `ask` condition, and the shortcut is now displayed. So if there are no shortcuts, no prompt shows up. Here's what it looks like if there are in fact shortcuts in the manifest
shortcuts

PR #47, opened 2019-12-05
Paths with space fix

### LINKED ISSUE #45: Build command fails with "The system cannot find the path specified"
Using the latest llama-pack instructions, I'm able to successfully execute the `init` command.

After completing the `init` command, I execute the `build` command and receive the following error:

```
PS C:\Dev\llama-pack> npx GoogleChromeLabs/llama-pack#0.1.0 build
(node:23244) ExperimentalWarning: The fs.promises API is experimental
Installing Android Build Tools. Please, read and accept the license agreement
Installing Build Tools
The system cannot find the path specified.
undefined
```

After the error, the tool exits.

When it says "system cannot find the path specified", it's not clear to me which path is being referred to, though I suspect it's the Android SDK path. When I first ran the `init` command, it prompted me for paths to JDK and Android SDK; my llama-pack-config.js looks like this:

```js
'use strict';
module.exports = {
  jdkPath: 'C:/Program Files/AdoptOpenJDK/jdk-8.0.232.09-hotspot/', // Path to the JDK 8
  androidSdkPath: 'C:/AndroidTools/' // Path to the Android SDK
};
```

**Any idea why I may be getting this path error?** I'm on Windows 10 Enterprise using Powershell.

andreban:
Closes #45

### DIFF
--- a/src/lib/util.js
+++ b/src/lib/util.js
@@ -70,7 +70,6 @@ async function untar(tarFile, destinationPath, deleteZipWhenDone = false) {
 
 function execInteractive(cwd, args, env) {
   const {spawn} = require('child_process');
-  console.log(cwd);
   const shell = spawn(`"${cwd}"`, args, {
     stdio: 'inherit',
     env: env,
--- a/src/lib/util.js
+++ b/src/lib/util.js
@@ -70,7 +70,8 @@ async function untar(tarFile, destinationPath, deleteZipWhenDone = false) {
 
 function execInteractive(cwd, args, env) {
   const {spawn} = require('child_process');
-  const shell = spawn(cwd, args, {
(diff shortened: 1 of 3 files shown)


src/lib/util.js
```diff
@@ -70,7 +70,8 @@ async function untar(tarFile, destinationPath, deleteZipWhenDone = false) {
 
 function execInteractive(cwd, args, env) {
   const {spawn} = require('child_process');
-  const shell = spawn(cwd, args, {
+  console.log(cwd);
```
PEConn: Did you mean to leave this in?

andreban: Nope. I actually noticed and removed already :)

PR #54, opened 2020-01-03
Typescript lint

andreban:


src/index.ts
```diff
@@ -16,14 +16,14 @@
...
 'use strict';
 
-const Cli = require('./cli');
+import Cli = require('./cli');
 
-module.exports = () => {
+module.exports = (): void => {
   const cli = new Cli();
   const args = process.argv.slice(2);
   // TODO(peconn): Tighten up the `any` below once CLI is typed.
```
PEConn: You can remove the TODO now (and thanks for doing it for me!).

src/lib/Config.ts
```diff
@@ -14,12 +14,16 @@
...
-const fs = require('fs');
-const homedir = require('os').homedir();
-const path = require('path');
-const fileExists = promisify(fs.exists);
+import {promisify} from 'util';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
PEConn: Why are we doing this?

src/lib/jdk/JdkHelper.ts
```diff
@@ -58,23 +65,21 @@ class JdkHelper {
...
-   * @returns {Object} an env object configure to run JDK commands
+   * @returns {NodeJS.ProcessEnv} an env object configure to run JDK commands
    */
-  getEnv() {
-    const env = Object.assign({}, this.process.env);
+  getEnv(): NodeJS.ProcessEnv {
+    const env: NodeJS.ProcessEnv = Object.assign({}, this.process.env);
     env['JAVA_HOME'] = this.getJavaHome();
     env[this.pathEnvironmentKey] =
         this.getJavaBin() + this.pathSeparator + env[this.pathEnvironmentKey];
```
PEConn: Why are you constructing the path manually here instead of using `joinPath`?

src/spec/lib/jdk/JdkHelperSpec.ts
```diff
@@ -16,52 +16,47 @@
...
-        jdkPath: '/home/user/jdk8',
-      };
+      const config = new Config('/home/user/jdk8', '/home/user/sdktools');
       const process = {
         platform: 'linux',
         env: {
           'PATH': '',
         },
-      };
+      } as unknown as NodeJS.Process;
```
PEConn: What's this about?

src/lib/types/WebManifest.ts
```diff
@@ -0,0 +1,36 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+export interface WebManifestIcon {
+  src: string;
+}
+
+export interface WebManifestShortcutJson {
+  name: string | undefined;
```
PEConn: You could make these optional, right?

```typescript
    name?: string;
    short_name?: string;
```

src/lib/jdk/JdkHelper.ts
andreban: This is actually editing the PATH environment variable. (eg: PATH="/jdk/bin:/example/bin"), not a file path. I'll add a comment explaining that.

src/spec/lib/jdk/JdkHelperSpec.ts
andreban: If we don't cast to `unknown` first, tsc will report the following error:
```
src/spec/lib/jdk/JdkHelperSpec.ts:40:23 - error TS2352: Conversion of type '{ platform: "darwin"; env: { 'PATH': string; }; }' to type 'Process' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```

src/lib/types/WebManifest.ts
andreban: Didn't actually know this was a thing. I wonder if there's a lint rule for that.

andreban:
We have `strict` enable, which enables `noImplicitAny`, `noImplicitThis`, `alwaysStrict`, and `strictNullChecks`

src/lib/types/WebManifest.ts
andreban: done

src/lib/jdk/JdkHelper.ts
andreban: Added the comment

src/lib/Config.ts
andreban: It's a bit of a leftover. Changed to a proper import.

src/index.ts
andreban: That's the upside of having a working linter. :)

src/lib/types/WebManifest.ts
andreban: I couldn't find a lint rules for this, which actually makes sense, since there's a difference between both approaches. Using `string | undefined` actually forces you to assign to the variable in the construction, while optional doesn't.

PR #55, opened 2020-01-07
Uses android-browser-helper

andreban:
- Replaces custom-tabs-client with the new library,
  android-browser-helper


andreban:
Do not merge yet. Will make another commit to use the official release of android-browser-helper.

andreban:
Updated to use the library from the Google Maven

PR #63, opened 2020-01-14
Move distributionUrl to the bottom

abdonrd:
When you create a new Android Project from Android Studio, it on the bottom.
Just to less diffing if you come from an Android Studio project.


PEConn:
Fine by me, Andre thoughts?

PR #64, opened 2020-01-14
Update com.android.tools.build:gradle version

abdonrd:


PEConn:
This LGTM - Andre any reason why we shouldn't update?

PR #67, opened 2020-01-15
Add new launcherName field

### LINKED ISSUE #61: Use name and short_name from the manifest
I think that right now llama-pack only use the `short_name` to the Android project.

I want to propose to get both: `name` and `short_name`. And then...

- Use the `name` for the Android project, application name, etc.
https://github.com/GoogleChromeLabs/llama-pack/blob/252a03347f6421e6b94239784ea4b740211a830a/template_project/app/src/main/AndroidManifest.xml#L28

- Use the `short_name` just for the homescreen.
https://github.com/GoogleChromeLabs/llama-pack/blob/252a03347f6421e6b94239784ea4b740211a830a/template_project/app/src/main/AndroidManifest.xml#L41

abdonrd:
Fix #61 

Implementation based on https://github.com/GoogleChromeLabs/llama-pack/issues/61#issuecomment-574596000

### DIFF
--- a/src/lib/TwaManifest.ts
+++ b/src/lib/TwaManifest.ts
@@ -121,7 +121,7 @@ export class TwaManifest {
     this.packageId = data.packageId;
     this.host = data.host;
     this.name = data.name;
-    this.launcherName = data.launcherName;
+    this.launcherName = data.launcherName || data.name;
     this.themeColor = new Color(data.themeColor);
     this.navigationColor = new Color(data.navigationColor);
     this.backgroundColor = new Color(data.backgroundColor);
@@ -257,7 +257,7 @@ export interface TwaManifestJson {
   packageId: string;
   host: string;
   name: string;
-  launcherName: string;
+  launcherName?: string; // Older Manifests may not have this field.
   themeColor: string;
   navigationColor: string;
   backgroundColor: string;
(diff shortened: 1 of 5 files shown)


andreban:
Thanks for the PR. Really appreciated!

One note: For developers updating their template via `llama-pack update`, their `twa-manifest.json` will not contain the `launcherName` field.

We could make `launcherName` in the `TwaManifestJson` optional:

```typescript
export interface TwaManifestJson {
  packageId: string;
  host: string;
  name: string;
  launcherName?: string; // Older Manifests may not have this field.
 ...
}
```

and use `name` in the constructor, if `launcherName` doesn't exist:

```typescript
  constructor(data: TwaManifestJson) {
    this.packageId = data.packageId;
    this.host = data.host;
    this.name = data.name;
    this.launcherName = data.launcherName || data.name;
```

Andre

abdonrd:
Oh! I didn't know there was an `update` option!

src/lib/TwaManifest.ts
```diff
@@ -120,6 +121,7 @@ export class TwaManifest {
     this.packageId = data.packageId;
     this.host = data.host;
     this.name = data.name;
+    this.launcherName = data.launcherName || data.name;
```
PEConn: Is `|| data.name` necessary? We already have the logic for that down on line 208.

PEConn: Oh, I see now that is for the update case.

Sorry for the noise (but do you mind adding a comment saying why we're doing it?).

abdonrd: Sure! 😄

PR #73, opened 2020-01-15
Allows Customizing Project Location on Gradle Wrapper

andreban:
@JudahGabriel drafted this to enable customizing the Gradle Wrapper.

### DIFF
--- a/src/lib/GradleWrapper.js
+++ b/src/lib/GradleWrapper.js
@@ -18,23 +18,26 @@
 
 const util = require('util');
 const exec = util.promisify(require('child_process').exec);
-const path = require('path');
 
 class GradleWrapper {
-  constructor(process, androidSdkTools, gradleProjectLocation) {
+  constructor(process, androidSdkTools, projectLocation) {
     this.process = process;
     this.androidSdkTools = androidSdkTools;
-    gradleProjectLocation = gradleProjectLocation || this.process.cwd();
+    this.projectLocation = projectLocation || this.process.cwd();
+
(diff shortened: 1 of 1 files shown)


JudahGabriel:
OK, I think the issue is we should instead be [setting the current directory of the node child process](https://stackoverflow.com/a/18898027/536). 

I think the final code would look like this:

```js
class GradleWrapper {
  constructor(process, androidSdkTools, gradleDirectory) {
    this.process = process;
    this.gradleDirectory = gradleDirectory;
    this.androidSdkTools = androidSdkTools;
    if (process.platform === 'win32') {
      this.gradleCmd = 'gradlew.bat';
    } else {
      this.gradleCmd = './gradlew';
    }
  }

  async assembleRelease() {
    const env = this.androidSdkTools.getEnv();
    await exec(`${this.gradleCmd} assembleRelease --stacktrace`, {
      env: env,
      cwd: gradleDirectory
   });
  }
```

I'll test this first and follow up shortly with the results.

JudahGabriel:
The above code seems to work - instead of appending the gradleDirectory to gradleCmd, just use the cwd option:

```js
await exec(`${this.gradleCmd} assembleRelease --stacktrace`, {
   env: androidSdkTools.getEnv(),
   cwd: gradleDirectory // set the current directory for the child process
});
```

andreban:
Ha! Sounds like I was overcomplicating things. Thanks for checking this out. I'll update the PR.

andreban:
@JudahGabriel does this sound good?

PR #75, opened 2020-01-18
Converts more Classes into Typescript

andreban:
- Converts AndroidSdkTools
- Converts GradleWrapper
- Converts AndroidSdkToolsSpec
- Updates other parts of the code to be compatible with the updated
  ones


andreban:
Improves #76

Issue #76, opened 2020-01-18
Convert Project into TypeScript


andreban:
The project was started using pure JavaScript. We're migrating the code to TypeScript, to make it more robust and easier to maintain.

Any new files added **must** use TypeScript.

- [x] `index.js`
- [x] `lib/Config.js`
- [x] `lib/GradleWrapper.js`
- [x] `lib/Log.js`
- [x] `lib/TwaGenerator.js`
- [x] `lib/TwaManifest.js`
- [x] `lib/util.js`
- [x] `lib/androidSdk/AndroidSdkTools.js`
- [x] `lib/androidSdk/AndroidSdkInstaller.js`
- [x] `lib/jdk/JdkHelper.js`
- [x] `lib/jdk/JdkInstaller.js`
- [x] `lib/jdk/KeyTool.js`
- [ ] `cli/index.js`
- [ ] `cli/cmds/init.js`
- [ ] `cli/cmds/build.js`
- [ ] `cli/cmds/update.js`
- [ ] `cli/cmds/help.js`
- [x] `spec/lib/ConfigSpec.js`
- [x] `spec/lib/TwaManifestSpec.js`
- [x] `spec/lib/utilSpec.js` 
- [x] `spec/lib/androidSdk/AndroidSdkToolsSpec.js`
- [x] `spec/lib/jdk/JdkHelperSpec.js`

PR #78, opened 2020-01-20
Converts util.js and utilSpec.js into TypesScript

andreban:
Advances towards #76


src/lib/TwaManifest.ts
```diff
@@ -64,10 +64,14 @@ class ShortcutInfo {
     this.url = shortcutInfo['url'] ?
       new URL(shortcutInfo['url'], webManifestUrl).toString() : undefined;
 
-    const suitableIcon = util.findSuitableIcon(shortcutInfo['icons'], 'any');
+    if (shortcutInfo.icons === undefined) {
```
PEConn: Would `if(!shortcutInfo.icons)` work here?

src/lib/util.ts
```diff
@@ -14,20 +14,21 @@
...
-const fs = require('fs');
-const util = require('util');
-const exec = util.promisify(require('child_process').exec);
-const tar = require('tar');
+const execPromise = util.promisify(exec);
 
-async function execute(cmd, env) {
-  await exec(cmd.join(' '), {env: env});
+export async function execute(cmd: string[], env: NodeJS.ProcessEnv): Promise<void> {
+  await execPromise(cmd.join(' '), {env: env});
```
PEConn: I feel this will fall over if any members of `cmd` have a space in them. Whether it's worth fixing or just worth documenting is up to you.

src/lib/TwaManifest.ts
andreban: Kind of. `tsc` complains about icons possibly being undefined in line 71 if we don't do an explicit test here.

src/lib/util.ts
andreban: Suppose we have something like:
`['foo', 'bar', 'foo bar]`

The output of `join` would be `foo bar foo bar`, which I think is what we expect. I'm actually using an approach with members containing spaces here: https://github.com/GoogleChromeLabs/llama-pack/blob/master/src/lib/jdk/KeyTool.js#L95-L108

src/spec/lib/utilSpec.ts
```diff
@@ -31,12 +29,27 @@ describe('util', () => {
           'sizes': '192x192',
           'type': 'image/png',
         }], 'any');
+    expect(result).not.toBeNull();
+    if (result === null) return;
```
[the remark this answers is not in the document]
andreban: The test aborts when the expectation fails. But `tsc`  doesn't compile the test, pointing out result could be null on the following tests. The test here makes `tsc` happy. I'll add a comment with a TODO item for me to check if there's a way to get tsc to understand the stuff below won't actually run if the first bit fails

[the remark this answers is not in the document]
andreban: Added the comment

src/lib/types/WebManifest.ts
```diff
@@ -19,6 +19,8 @@ export interface WebManifestIcon {
   src: string;
   sizes?: string;
   purpose?: string;
+  type?: string;
```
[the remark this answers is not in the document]
andreban: Renamed to `mymeType`

PR #81, opened 2020-01-20
Converts TwaGenerator.js and Log.js to TypeScript

andreban:
Advances #76


src/lib/Log.ts
```diff
@@ -0,0 +1,125 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/**
+ * An utility class to print nice Log messages.
+ */
+export default class Log {
```
PEConn: I'm surprised there's not a library to do this. Maybe there's just not a nice and small library that does this.

andreban: I kind of copied this from amp-toolbox: https://github.com/ampproject/amp-toolbox/blob/master/packages/core/lib/Log.js

PR #82, opened 2020-01-21
Converts `KeyTool.js` to TypeScript

andreban:
- Converts `KeyTool.js` to TypeScript
- Removes `JdkInstaller.js`, as it's not being used and is outdated.
- Removes `AndroidSdkInstaller.js`, as it's not being used and is
  outdated.
- Moves code prompting for user input to outside `KeyTool.js` and
  init `init.js`.


andreban:
Improves #76

src/lib/jdk/KeyTool.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+import * as fs from 'fs';
+import * as util from '../util';
+import {JdkHelper} from './JdkHelper';
+
+export interface CreateKeyOptions {
+  path: string;
+  alias: string;
+  cn: string; // First and Last Name
```
PEConn: Is there any reason other than brevity for these variable names? If not, can we just use `fullName`, `orgUnit`, `org` and `country`?

src/lib/jdk/KeyTool.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import * as fs from 'fs';
+import * as util from '../util';
```
PEConn: This is more of a style thing, and I'm happy to be outvoted, but I'd prefer it if we explicitly listed our imports instead of using `import *`.

src/lib/jdk/KeyTool.ts
```diff
@@ -0,0 +1,74 @@
...
+  o: string; // Organization
+  c: string; // Country
+  keypassword: string;
+  password: string;
+}
+/**
+ * A Wrapper of the Java keytool command-line tool
+ */
+export class KeyTool {
+  jdkHelper: JdkHelper;
```
PEConn: This should be `private` right?

src/lib/jdk/KeyTool.ts
```diff
@@ -0,0 +1,74 @@
...
+      } else {
+        return;
+      }
+    }
+
+    // Execute Java Keytool
+    const keytoolCmd = [
+      'keytool',
+      '-genkeypair',
+      `-dname "cn=${keyOptions.cn}, ou=${keyOptions.ou}, o=${keyOptions.o}}, c=${keyOptions.c}"`,
```
PEConn: One too many `}` (after `keyOptions.o`)?

src/lib/jdk/KeyTool.ts
```diff
@@ -0,0 +1,74 @@
...
+      `-alias ${keyOptions.alias}`,
+      `-keypass ${keyOptions.keypassword}`,
+      `-keystore ${keyOptions.path}`,
+      `-storepass ${keyOptions.password}`,
+      '-validity 20000',
+      '-keyalg RSA',
+    ];
+    const env = this.jdkHelper.getEnv();
+    await util.execute(keytoolCmd, env);
+    console.log('Signing Key created successfully');
```
PEConn: Not using our pretty little logging tool?

src/lib/jdk/KeyTool.ts
andreban: Ha, good catch!

src/lib/jdk/KeyTool.ts
andreban: SG - will change to import each individual one.

andreban: Sounds like a good one. done.

src/lib/jdk/KeyTool.ts
andreban: Those just match the parameters passed to keytool. I agree  it's better to have more descriptive names.

src/lib/jdk/KeyTool.ts
andreban: done

src/lib/jdk/KeyTool.ts
andreban: Added :D

src/lib/jdk/KeyTool.ts
andreban: Good catch. That didn't actually break and just added '}' to the output.

andreban:
Addressed the feedback. PTAL.

PR #83, opened 2020-01-22
Escaping user input to key tool

JudahGabriel:
Fixes the issue where e.g. the key password contains a hyphen.

### DIFF
--- a/src/lib/jdk/KeyTool.ts
+++ b/src/lib/jdk/KeyTool.ts
@@ -66,10 +66,10 @@ export class KeyTool {
       '-genkeypair',
       `-dname "cn=${keyOptions.fullName}, ou=${keyOptions.organizationalUnit}, ` +
           `o=${keyOptions.organization}, c=${keyOptions.country}"`,
-      `-alias ${keyOptions.alias}`,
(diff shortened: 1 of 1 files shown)


JudahGabriel:
@googlebot I signed it!

src/lib/jdk/KeyTool.ts
```diff
@@ -66,10 +66,10 @@ export class KeyTool {
       '-genkeypair',
       `-dname "cn=${keyOptions.fullName}, ou=${keyOptions.organizationalUnit}, ` +
           `o=${keyOptions.organization}, c=${keyOptions.country}"`,
-      `-alias ${keyOptions.alias}`,
-      `-keypass ${keyOptions.keypassword}`,
-      `-keystore ${keyOptions.path}`,
-      `-storepass ${keyOptions.password}`,
+      `-alias \"${keyOptions.alias}\"`,
```
PEConn: You don't need the backslashes here since we're using a template string (eg, there's just `"` instead of `\"` on line 67).

andreban: PS: Even though it's not needed, the output is correct with the escapes.

Issue #31, opened 2019-11-25
Harden Input Validation


andreban:
llama-pack relies on user input in many places. Catching errors at the moment the user is inputing data will make it easier to recover from errors and generate better error messages

andreban:
This has been closed by #86

PR #86, opened 2020-01-27
Moves `cli` to TypeScript and uses Inquirer.js

### LINKED ISSUE #76: Convert Project into TypeScript

### LINKED ISSUE #79: Replace `prompt` with `inquirer.js`
[`prompt`](https://github.com/flatiron/prompt) seems to be largely unmaintained, with the last comming being around 3 years ago and many open issues. TypeScript is also not supported (which will affect #76)

[`inquirer.js`](https://www.npmjs.com/package/inquirer) seems better maintained and has [TypesScript types](https://www.npmjs.com/package/@types/inquirer).

andreban:
- Moves the remaining JS files in cli/ to TypeScript
- Replaces 'prompt' with 'Inquirer.js'

Closes #76 
Closes #79


src/index.ts
```diff
@@ -14,9 +14,7 @@
  *  limitations under the License.
  */
 
-'use strict';
```
PEConn: how come we're removing this?

src/lib/Config.ts
```diff
@@ -18,11 +18,10 @@ import {promisify} from 'util';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as os from 'os';
-import userPrompt = require('prompt');
+import * as inquirer from 'inquirer';
```
PEConn: Not essential, but I'd still prefer we be precise with our imports.

src/lib/TwaGenerator.ts
```diff
@@ -21,6 +21,7 @@ import {template} from 'lodash';
 import {promisify} from 'util';
 import {TwaManifest} from './TwaManifest';
 import Jimp = require('jimp');
+import Log from './Log';
```
PEConn: TypeScript question - why do some of these have `{}` and some don't?

src/cli/inputHelpers.ts
```diff
@@ -14,8 +14,26 @@
...
 
-declare module 'prompt' {
-  import prompt = require('prompt');
-  // TODO(andreban): Remove usage of any
-  export function get(schema: any): any; // eslint-disable-line @typescript-eslint/no-explicit-any
+import Color = require('color');
+import {isWebUri} from 'valid-url';
+
+export function validatePassword(input: string): boolean {
+  return input.length > 0;
```
PEConn: Sure we don't want something longer?

src/cli/inputHelpers.ts
```diff
@@ -14,8 +14,26 @@
...
+}
+
+export function notEmpty(input: string): boolean {
+  return input.length > 0;
```
PEConn: `input.trim().length > 0` ?

src/cli/inputHelpers.ts
```diff
@@ -14,8 +14,26 @@
  *  limitations under the License.
  */
 
-declare module 'prompt' {
-  import prompt = require('prompt');
-  // TODO(andreban): Remove usage of any
-  export function get(schema: any): any; // eslint-disable-line @typescript-eslint/no-explicit-any
+import Color = require('color');
```
PEConn: And again with the imports questions, why is this `require`?

src/cli/cmds/help.ts
```diff
@@ -14,9 +14,6 @@
  *  limitations under the License.
  */
 
-'use strict';
-
-async function help() {
+export async function help(): Promise<void> {
+  return;
```
PEConn: help is a no-op?

src/cli/cmds/init.ts
```diff
@@ -0,0 +1,210 @@
...
+}
+
+async function confirmTwaConfig(twaManifest: TwaManifest): Promise<TwaManifest> {
+  const result = await inquirer.prompt([
+    {
+      name: 'host',
+      type: 'input',
+      message: 'Domain being opened in the TWA:',
+      default: twaManifest.host,
+      validate: notEmpty,
```
PEConn: We could check this is a valid URL (future CL though).

src/cli/cmds/init.ts
```diff
@@ -0,0 +1,210 @@
...
+    }, {
+      name: 'name',
+      type: 'input',
+      message: 'Name of the application:',
+      default: twaManifest.name,
+      validate: notEmpty,
```
PEConn: Similarly, we could check that this and `launcherName` aren't too long.

src/cli/cmds/init.ts
```diff
@@ -0,0 +1,210 @@
...
+    }, {
+      name: 'launcherName',
+      type: 'input',
+      message: 'Name to be shown on the Android Launcher:',
+      default: twaManifest.launcherName,
+      validate: notEmpty,
+    }, {
+      name: 'themeColor',
+      type: 'input',
+      message: 'Color to be used for the status bar:',
```
PEConn: It's maybe worth putting an example of the value format (is it #232323, just 232323, either, both?).

src/index.ts
andreban: We have `strict: true` on `tsconfig.json`, so TypeScript will add `'use string' "to the output JS for us.

src/lib/TwaGenerator.ts
andreban: On the Log, we do `export default class Log`. on TwaManifest, only `export class TwaManifest`.

src/cli/inputHelpers.ts
andreban: I need to look at the requirements for keytool and replicate there. Doing on a following CL.

src/cli/cmds/help.ts
andreban: For now... Need to add actual help messages (I'll do in another PR. Focusing this one on translating JS => TS)

src/cli/inputHelpers.ts
andreban: I don't fully understand the require one, but it's related to the underlying library being written in JavaScript. Using `import * as Color from 'color'`, `import {Color} from 'color'` or `import Color from 'color'` reports the error below:
```
Module '"/Users/andreban/Projects/twa-bootstrap/node_modules/@types/color/index"' can only be default-imported using the 'esModuleInterop' flagts(1259)

index.d.ts(135, 1): This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.
```

src/cli/cmds/init.ts
andreban: This one is not a full URL, host a host (I agree we should check)

andreban: #31 is the one I created for those.

src/cli/cmds/init.ts
andreban: Another good one for #31 (Adding here so I get the mentions in the issue)

src/cli/cmds/init.ts
andreban: The default value shows the correct format (`#232323`). Is this enough?

andreban: Also, we accept a wide range of formats. Including `blue`, `rgb(0, 0, 255)`, etc.

PEConn: Yeah, that's fine.

src/lib/Config.ts
andreban: I'm a bit divided in the case where a library has many related methods. 

When we import with `import * as inquirer from 'inquirer'`, then we get to use `inquirer.prompt(...)` in the code, which makes it clear where prompt is coming from.

When we import with `import {prompt} from 'inquirer'`, the call in the code becomes just `prompt`, which is less obvious where it comes from, unless you scan the imports. 

WDYT?

src/cli/inputHelpers.ts
andreban: +1. done

PR #89, opened 2020-01-29
Adds help messages to the help command

### LINKED ISSUE #87: Make the `help` command helpful
`help` is currently no-op. Add a proper help.

andreban:
Closes #87

### DIFF
--- a/src/cli/cmds/help.ts
+++ b/src/cli/cmds/help.ts
@@ -59,7 +59,7 @@ export async function help(args: ParsedArgs, log = new Log('help')): Promise<voi
   const command = args._[1];
   const message = HELP_MESSAGES.get(command) || HELP_MESSAGES.get('main');
 
-  // Should never happen, as we know we have a message for Command.main,
+  // Should never happen, as we know we have a message for main,
   // but TypeScript doesn't know it.
   if (message === undefined) {
     log.error(`Could not find help message for args: ${args}`);
--- a/src/cli/cmds/help.ts
+++ b/src/cli/cmds/help.ts
@@ -14,6 +14,57 @@
  *  limitations under the License.
  */
 
-export async function help(): Promise<void> {
+import Log from '../../lib/Log';
+import {ParsedArgs} from 'minimist';
+
+const HELP_MESSAGES = new Map<string, string>(
+    [
+      ['main', [
+        'llama-pack [command] <options>',
+        '',
+        '',
+        'build ............... generates an Android APK from a TWA Project',
+        'help ................ shows this menu',
+        'init ................ initializes a new TWA Project',
+        'update .............. updates an existing TWA Project with the latest llama-pack template',
+      ].join('\n')],
+      ['init', [
+        'Usage:',
+        '',
+        '',
+        'lama-pack init --manifest=[web-manifest-url]',
+        '',
+        '',
+        'Options:',
+        '--directory ......... path where to generate the project. Defaults to the current' +
+            ' directory',
+      ].join('\n')],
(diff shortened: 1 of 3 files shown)


src/cli/cmds/help.ts
```diff
@@ -14,6 +14,57 @@
...
  */
 
-export async function help(): Promise<void> {
+import Log from '../../lib/Log';
+import {ParsedArgs} from 'minimist';
+
+const HELP_MESSAGES = new Map<string, string>(
+    [
+      ['main', [
+        'llama-pack [command] <options>',
```
PEConn: You can use template literals for multiline strings.

src/cli/cmds/help.ts
```diff
@@ -14,6 +14,57 @@
...
+);
+
+export async function help(args: ParsedArgs, log = new Log('help')): Promise<void> {
+  // minimist uses an `_` object to store details.
+  const command = args._[1];
+  const message = HELP_MESSAGES.get(command) || HELP_MESSAGES.get('main');
+
+  // Should never happen, as we know we have a message for main,
+  // but TypeScript doesn't know it.
+  if (message === undefined) {
```
PEConn: Yeah, so what you want here is an `!` (a (Non-null assertion operator)[https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator]).

src/cli/cmds/help.ts
PEConn: (if you want, I'm also fine with this way.)

src/cli/cmds/help.ts
andreban: Interestingly, the linter configuration dislikes it:

```
 69:12  warning  Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion
```

src/cli/cmds/help.ts
andreban: Tired it out, but it kind of breaks the indentation, as otherwise the blank characters at the beginning of the line are printed.

src/cli/cmds/help.ts
andreban: Here's the rule: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-non-null-assertion.md

I can keep the current role and the rule our add an eslint-ignore-line. I kind of prefer the first.

src/cli/cmds/help.ts
PEConn: Oh yeah, I'd have it formatted something like:

```
const HELP_MESSAGES = new Map<string, string> (
    [
      ['main',
`
llama-pack [command] <options>

build ............... generates an Android APK from a TWA Project
help ................ shows this menu
`],
      ['init',
`
...
`
```

But yeah, I'm fine with your way. (Mixing the indent of code and output strings is usually horrible - I guess the "proper" way to do this would be to have the help strings nicely formatted in text files that we read in (and it would allow for easier internationalisation, if we ever get that far!).

src/cli/cmds/help.ts
PEConn: I disagree with that rule - the point of strict null checking is for the compiler to catch things for you that you would have missed. In this case, we've acknowledged the possibility of null and said that we want to crash if that's the case.

Sorry, I don't understand what you're saying with "I can keep the current role and the rule"

andreban: I agree. Changed eslint.json to remove the rule and removed the unnecessary check from help.ts.

PR #91, opened 2020-01-31
Allows automated builds

### LINKED ISSUE #30: Allow automating the build command.
Currently the `build` command asks for passwords interactively which makes it tricky for build to be run as part of an automated script.

We should allow passing the passwords as part of the command line invocation.

andreban:
- Developers can use the `LLAMA_PACK_KEYSTORE_PASSWORD` and
  `LLAMA_PACK_KEY_PASSWORD` enviroment variables to set the
  keystore and key passwords.
- When both passwords are set, they will be used instead of
  prompting the users for the passwords.

Closes #30

### DIFF
--- a/src/cli/cmds/build.ts
+++ b/src/cli/cmds/build.ts
@@ -23,6 +23,11 @@ import Log from '../../lib/Log';
 import * as inquirer from 'inquirer';
 import {validatePassword} from '../inputHelpers';
 
+interface SigningKeyPasswords {
+  keystorePassword: string;
+  keyPassword: string;
+}
+
 /**
  * Checks if the keystore password and the key password are part of the environment prompts the
  * user for a password otherwise.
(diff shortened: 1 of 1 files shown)


src/cli/cmds/build.ts
```diff
@@ -23,17 +23,22 @@ import Log from '../../lib/Log';
...
-  const jdkHelper = new JdkHelper(process, config);
-  const androidSdkTools = new AndroidSdkTools(process, config, jdkHelper);
+/**
+ * Checks if the keystore password and the key password are part of the environment prompts the
+ * user for a password otherwise.
+ *
+ * @returns {Promise<[string, string]>} A promise with a tuple where the first item is they
+ * keystore password and the second is the key password.
+ */
+async function getPasswords(): Promise<[string, string]> {
```
PEConn: Could you create a class/interface for this with `keystorePassword` and `keyPassword` as members?

src/cli/cmds/build.ts
```diff
@@ -23,17 +23,22 @@ import Log from '../../lib/Log';
...
+  // Check if passwords are set as environment variables.
+  const envKeystorePass = process.env['LLAMA_PACK_KEYSTORE_PASSWORD'];
+  const envKeyPass = process.env['LLAMA_PACK_KEY_PASSWORD'];
 
-  if (!await androidSdkTools.checkBuildTools()) {
-    console.log('Installing Android Build Tools. Please, read and accept the license agreement');
-    await androidSdkTools.installBuildTools();
+  if (envKeyPass !== undefined && envKeystorePass !== undefined) {
+    return [envKeystorePass, envKeyPass];
```
PEConn: Could you add a log line here saying something like "Using passwords set in the LLAMA_PACK_KEYSTORE_PASSWORD and LLAMA_PACK_KEY_PASSWORD environmental variables."?

src/cli/cmds/build.ts
andreban: done

src/cli/cmds/build.ts
andreban: done

PR #93, opened 2020-02-01
Report an Error when `iconUrl` is not 200

andreban:
- When status is not 200, the generation fails with an unclear
  error message.
- Added an explicit check for the status, with a message that is
  easier to read.

### DIFF
--- a/src/lib/TwaGenerator.ts
+++ b/src/lib/TwaGenerator.ts
@@ -169,7 +169,8 @@ export class TwaGenerator {
   private async fetchIcon(iconUrl: string): Promise<Icon> {
     const response = await fetch(iconUrl);
     if (response.status !== 200) {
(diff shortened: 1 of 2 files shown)


src/lib/TwaGenerator.ts
```diff
@@ -168,6 +168,9 @@ export class TwaGenerator {
    */
   private async fetchIcon(iconUrl: string): Promise<Icon> {
     const response = await fetch(iconUrl);
+    if (response.status !== 200) {
+      throw new Error(`Failed to download icon ${iconUrl}, with status ${response.status}`);
```
PEConn: Could you say "Responded with status ..." instead of just "with status"?

andreban: done!

PR #98, opened 2020-02-05
Adds web-manifest-url meta-tag and web-app-manifest.json

### LINKED ISSUE #88: Add web_alternative flag and include web manifest for Chrome OS
Change the `cros_web_alternative` flag to something not Chrome OS specific and include the web app manifest in the raw resources to allow installation when without a network request.

Chrome side bug: https://crbug.com/1026230

andreban:
- When webManifestUrl is set on twa-manifest.json, it includes a
 `web-manifest-url` meta-tag on AndroidManifest.xml with the
  correct value and creates a web-app-manifest.json at /res/raw
- When webManifestUrl ist not set, it doesn't add any meta tag

Closes #88

### DIFF
--- a/template_project/app/src/main/AndroidManifest.xml
+++ b/template_project/app/src/main/AndroidManifest.xml
@@ -33,9 +33,11 @@
             android:name="asset_statements"
             android:resource="@string/assetStatements" />
 
-        <meta-data
-            android:name="cros_web_alternative"
-            android:value="@string/crosLaunchUrl" />
+        <% if (webManifestUrl) { %>
+            <meta-data
+                android:name="web_manifest_url"
+                android:value="@string/webManifestUrl" />
+        <% } %>
 
         <meta-data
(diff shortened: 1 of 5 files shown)


template_project/app/src/main/AndroidManifest.xml
```diff
@@ -33,9 +33,11 @@
             android:name="asset_statements"
             android:resource="@string/assetStatements" />
 
-        <meta-data
-            android:name="cros_web_alternative"
-            android:value="@string/crosLaunchUrl" />
+        <% if (webManifestUrl) { %>
```
PEConn: Huh, I did not know we could use substitutions in the AndroidManifest.xml

andreban: That's lodash, the template engine we're using in llama-pack.

andreban:
@dominickng FYI

PR #99, opened 2020-02-07
Refactors TwaManifest.shortcuts

andreban:
- Change `TwaManifest.shortcuts` to be `ShortcutInfo[]` instead
  of `string`.
- Refactor code to reflect that change
- Fixes reversed if condition in `init` leading to not adding
  shortcuts.
- Test the result of `generateShortcut()`
- Uses correct `LauncherActivity` in `build.gradle`
- Truncates `short_name` to 12 characters, when unavailable and copying from `name`. Recommendation for 12 chars [here](https://developer.chrome.com/apps/manifest/name#short_name)

### DIFF
--- a/src/lib/TwaGenerator.ts
+++ b/src/lib/TwaGenerator.ts
@@ -19,7 +19,7 @@ import * as fs from 'fs';
 import fetch from 'node-fetch';
 import {template} from 'lodash';
 import {promisify} from 'util';
-import {TwaManifest} from './TwaManifest';
+import {TwaManifest, ShortcutInfo} from './TwaManifest';
 import Jimp = require('jimp');
 import Log from './Log';
 
@@ -232,10 +232,7 @@ export class TwaGenerator {
       await this.generateIcons(twaManifest.iconUrl, targetDirectory, IMAGES);
     }
 
-    // TODO(andreban): TwaManifest.shortcuts is a string, which is being parsed into an Object.
-    // Needs to be transformed into a proper Class.
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    await Promise.all(JSON.parse(twaManifest.shortcuts).map((shortcut: any, i: number) => {
+    await Promise.all(twaManifest.shortcuts.map((shortcut: ShortcutInfo, i: number) => {
       const imageDirs = SHORTCUT_IMAGES.map(
           (imageDir) => ({...imageDir, dest: `${imageDir.dest}shortcut_${i}.png`}));
       return this.generateIcons(shortcut.chosenIconUrl, targetDirectory, imageDirs);
--- a/src/lib/TwaManifest.ts
+++ b/src/lib/TwaManifest.ts
@@ -73,10 +73,6 @@ export class ShortcutInfo {
    */
   constructor(readonly name: string, readonly shortName: string, readonly url: string,
         readonly chosenIconUrl: string) {
-    this.name = name;
-    this.shortName = shortName;
-    this.url = url;
-    this.chosenIconUrl = chosenIconUrl;
   }
 }
 
--- a/src/lib/TwaManifest.ts
+++ b/src/lib/TwaManifest.ts
@@ -212,28 +212,26 @@ export class TwaManifest {
 
     const shortcuts: ShortcutInfo[] = [];
 
-    if (webManifest.shortcuts) {
-      for (const s of webManifest.shortcuts) {
-        if (!s.icons || !s.url || (!s.name && !s.short_name)) {
-          continue;
-        }
+    for (const s of webManifest.shortcuts || []) {
+      if (!s.icons || !s.url || (!s.name && !s.short_name)) {
+        continue;
+      }
 
-        const suitableIcon = util.findSuitableIcon(s.icons, 'any');
-        if (!suitableIcon) {
-          continue;
-        }
+      const suitableIcon = util.findSuitableIcon(s.icons, 'any');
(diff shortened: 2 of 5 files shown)


andreban:
@rayankans can you take a look as well?

src/lib/TwaManifest.ts
```diff
@@ -222,16 +207,39 @@ export class TwaManifest {
...
+
+        const name = s.name || s.short_name;
+        const shortName = s.short_name || s.name!.substring(0, 12);
+        const url = new URL(s.url, webManifestUrl).toString();
+        const iconUrl = new URL(suitableIcon.src, webManifestUrl).toString();
+        const shortcutInfo = new ShortcutInfo(name!, shortName!, url, iconUrl);
+
+        shortcuts.push(shortcutInfo);
+
+        if (shortcuts.length === 3) {
```
rayankans: 4

src/lib/TwaGenerator.ts
```diff
@@ -232,10 +232,7 @@ export class TwaGenerator {
       await this.generateIcons(twaManifest.iconUrl, targetDirectory, IMAGES);
     }
 
-    // TODO(andreban): TwaManifest.shortcuts is a string, which is being parsed into an Object.
-    // Needs to be transformed into a proper Class.
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
andreban: @PEConn I think this is the last `any`

template_project/app/build.gradle
```diff
@@ -148,7 +148,7 @@ task generateShorcutsFile {
                     'intent'(
                             'android:action': 'android.intent.action.MAIN',
                             'android:targetPackage': twaManifest.applicationId,
-                            'android:targetClass': 'android.support.customtabs.trusted.LauncherActivity',
+                            'android:targetClass': 'com.google.androidbrowserhelper.trusted.LauncherActivity',
```
rayankans: Are you sure about this? I'm guessing you tested it and the previous targetClass did not work.

src/lib/TwaManifest.ts
```diff
@@ -222,16 +207,39 @@ export class TwaManifest {
 
     const fullStartUrl: URL = new URL(webManifest['start_url'] || '/', webManifestUrl);
 
-    const shortcuts: ShortcutInfo[] = (webManifest.shortcuts || [])
-        .map((s: WebManifestShortcutJson) => new ShortcutInfo(s, webManifestUrl))
-        .filter((s: ShortcutInfo) => s.isValid())
-        .filter((_: ShortcutInfo, i: number) => i < 4);
+    const shortcuts: ShortcutInfo[] = [];
+
+    if (webManifest.shortcuts) {
```
rayankans: nit: `for (const s of (webManifest.shortcuts || [])) {`
to reduce indentation.

I'm not too familiar with TS but it seems to me that webManifest.shortcuts is guaranteed to be a list instead of a null or undefined though?

template_project/app/build.gradle
PEConn: You've tested this, right?

src/lib/TwaGenerator.ts
PEConn: 🎊

src/lib/TwaManifest.ts
```diff
@@ -61,34 +64,19 @@ function generatePackageId(host: string): string {
...
-    if (suitableIcon !== null) {
-      this.chosenIconUrl = new URL(suitableIcon.src, webManifestUrl).toString();
-    }
-  }
-
-  isValid(): boolean {
-    return this.name != undefined && this.url != undefined && this.chosenIconUrl != undefined;
+  constructor(readonly name: string, readonly shortName: string, readonly url: string,
+        readonly chosenIconUrl: string) {
+    this.name = name;
```
PEConn: You're using [Parameter Properties][1] here right? I don't think you need lines 76 to 79.

[1]: https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties

src/lib/TwaManifest.ts
andreban: This is a WebManifest shortcut, defined here: https://github.com/GoogleChromeLabs/llama-pack/blob/master/src/lib/types/WebManifest.ts#L40

It can be undefined.

template_project/app/build.gradle
andreban: I think this is a left-over when we migrated to `androix.browser-1.2.0`, which doesn't have a LauncherActivity anymore. This should have been changed with the other references across the code.

src/lib/TwaManifest.ts
andreban: done

src/lib/TwaManifest.ts
andreban: changed the for loop.

template_project/app/build.gradle
andreban: Yes, tested against `https://contentindex.dev/manifest.json`. It works :)

rayankans: Does the previous value not work though?  It's still working for me

src/lib/TwaManifest.ts
andreban: This is great! Updated!

template_project/app/build.gradle
andreban: Are you using the `android-browser-helper-1.1.0`? The old class doesn't exist anymore in `androidx.browser`, so I wouldn't expect it to work.

PR #101, opened 2020-02-11
Splits the project into `core` and `cli` modules

andreban:
- Uses `lerna` to manage modules
- Adds README.md for each module
- Updates version to 0.4.0


README.md
```diff
@@ -28,58 +28,11 @@ feature requests, and contribute with pull requests, if possible.
...
-When building the project for the first time, the Android Build Tools will need to be installed.
-The tool will inkove the installation process for the build tools. Make sure to read and accept
-the license agreement before proceeding.
-
-As a result of the build step, the tool will generate a signed APK (`app-release-signed.apk`)
-that can be uploaded to the Play Store. You will also need to deploy a Digital Asset Links file to
-validate your domain. The
-[TWA Quick Start Guide](https://developers.google.com/web/updates/2019/08/twas-quickstart#creating-your-asset-link-file)
-explains how to extract the information needed to generate it.
+- **[llama-pack-core](./packages/core):** a javascript library that for generating, building and
```
PEConn: nit: remove "that"

README.md
```diff
@@ -28,58 +28,11 @@ feature requests, and contribute with pull requests, if possible.
...
+updating TWA projects.
+- **[llama-pack-cli](./packages/cli):** a command-line version for llama-pack.
```
PEConn: A command-line "version" sounds a bit odd.

"llama-pack core wrapped for the command line" maybe?

package.json
```diff
@@ -24,33 +22,17 @@
...
+    "@llama-pack/core": "file:packages/core",
+    "@llama-pack/cli": "file:packages/cli"
   },
   "devDependencies": {
     "@typescript-eslint/eslint-plugin": "^2.9.0",
     "@typescript-eslint/parser": "^2.9.0",
     "eslint": "^6.6.0",
     "eslint-config-google": "^0.14.0",
-    "jasmine": "^3.5.0"
+    "jasmine": "^3.5.0",
```
PEConn: So the tests are in the parent package?

packages/core/src/index.ts
```diff
@@ -0,0 +1,34 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {AndroidSdkTools} from './lib/androidSdk/AndroidSdkTools';
+import {Config} from './lib/Config';
+import {GradleWrapper} from './lib/GradleWrapper';
+import Log from './lib/Log';
```
PEConn: Why no {}?

README.md
andreban: "command-line interface for llama-pack"?

packages/core/src/index.ts
andreban: `Log` is using `export default Log`, while others simple use `export`. Maybe I should change `Log` to just use `export`, but likely in another PR.

package.json
andreban: We don't cleaned-up the dependencies in root. We can re-add later if we end up needing.

Issue #100, opened 2020-02-11
llama-pack init fails for squoosh.app


andreban:
```
init Fetching Manifest:  https://squoosh.app/manifest.json
? Domain being opened in the TWA: squoosh.app
? Name of the application: Squoosh
? Name to be shown on the Android Launcher: Squoosh
? Color to be used for the status bar: #F78F21
? Color to be used for the splash screen background: #FFFFFF
? Relative path to open the TWA: /
? URL to an image that is at least 512x512px: https://squoosh.app/assets/icon-large.png
? URL to an image that is at least 512x512px to be used when generating maskable icons (https://squoosh.app/assets/icon-large-maskable.png) 1
>> TypeError: uri.match is not a function
```

andreban:
Closed via #102

andreban:
@naveedahmed1 the issue you described in here https://github.com/GoogleChromeLabs/svgomg-twa/issues/48#issuecomment-585159035 must be fixed now. Can you confirm?

naveedahmed1:
Yes it has been fixed, thank you so much @andreban :)

But the issue with push notifications still exists i.e. when I click Push Notification it opens it in website and not in the app.

PR #102, opened 2020-02-12
Fixes handling of maskable images

andreban:
- [cli] Fixes validation for maskable images
- [core] Checks for content-type when downloading images and
  throws an error when the content-type is invalid.

### DIFF
--- a/packages/core/src/lib/TwaGenerator.ts
+++ b/packages/core/src/lib/TwaGenerator.ts
@@ -201,6 +201,13 @@ export class TwaGenerator {
           ` Responded with Content-Type "${contentType}"`);
     }
 
+    // TODO(andreban): Support for image/svg being tracked in
+    // https://github.com/GoogleChromeLabs/llama-pack/issues/103
+    if (contentType.startsWith('image/svg')) {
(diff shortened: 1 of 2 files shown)


andreban:
Fixes #100

packages/core/src/lib/TwaGenerator.ts
```diff
@@ -194,6 +194,13 @@ export class TwaGenerator {
       throw new Error(
           `Failed to download icon ${iconUrl}. Responded with status ${response.status}`);
     }
+
+    const contentType = response.headers.get('content-type');
+    if (!contentType?.startsWith('image/')) {
```
PEConn: Can Android handle any type of image (can it do SVGs?)?

andreban: The image that is read here is converted to PNG by `jimp` when saved. To your point, `jimp` doesn't support reading SVGs. So, we should reject it for now and create an issue to add support for enabling reading SVGs.

andreban: Throwing a specific error for image/svg. I also created https://github.com/GoogleChromeLabs/llama-pack/issues/103

PR #108, opened 2020-02-22
Reduce package size by omitting tests & build tools

NotWoods:
- Moved some dependencies to devDependencies
- Removed dist/spec from shipped files in core
- Used promisify with extractZip
- Deleted unused copy of bin/llama-pack


packages/cli/src/bin/llama-pack.js
```diff
@@ -1,21 +0,0 @@
-#!/usr/bin/env node
```
andreban: Removing this file will break running the CLI from the command-line. The reasons it exists is the shebang on the first line. It is also rerferenced in `/packages/cli/package.json`.

NotWoods: package.json references `bin/llama-pack.js`, which is still present. This is a duplicate file in the `src` directory.

andreban:
Ohhh, makes sense. Good catch! All seems good then.

PR #109, opened 2020-02-22
Move Config read/write into cli

NotWoods:
Moves functions for reading/writing Config into the cli, since none of them are used in core. check() has been removed since its no longer used.

Also added a `path` parameter so a prompt could be added to allow custom config paths.

Core Config is now just a data class, so perhaps it should be changed to an interface.

### DIFF
--- a/packages/cli/src/lib/Cli.ts
+++ b/packages/cli/src/lib/Cli.ts
@@ -15,15 +15,15 @@
  */
 
 import * as minimist from 'minimist';
-import {Config} from '@llama-pack/core';
 import {update} from './cmds/update';
 import {help} from './cmds/help';
 import {build} from './cmds/build';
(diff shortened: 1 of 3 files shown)


PEConn:
I'm happy with adding the `path`, unsure about whether methods for reading/writing the config should be part of the `core` library even if they're not used there. @andreban thoughts?

andreban:
Functions to load and save the config would ideally be part of the library. 

The section of code that tries to load a config from a file and asks users for config information if it fails  (`loadOrCreateConfig`) and `createConfig` should be part of the CLI.

We may also want to add `Config#serialize(): string` and `Config#deserialise(config: string): Config` to make it easier for developers who'd like to save a Config to a database, for instance.

PR #114, opened 2020-02-29
Fix packageName generation

### LINKED ISSUE #113: Replace all occurrences of the dashes to underscores in package id
Command `llama-pack init --manifest https://maskable-transparent-bg-bug.web.app/manifest.json` produce `"packageId": "app.web.maskable_transparent-bg-bug.twa"`, which is invalid. Should be `"packageId": "app.web.maskable_transparent_bg_bug.twa"`.

andreban:
- Fixes regex to handle multiple invalid characters
- Moves generatePackageId to util.
- Adds a test for multiple invalid characters

Closes #113


packages/core/src/lib/util.ts
```diff
@@ -25,6 +25,10 @@ import {WebManifestIcon} from './types/WebManifest';
 const execPromise = promisify(exec);
 const extractZipPromise = promisify(extractZip);
 
+// Regex for disallowed characters on Android Packages, as per
+// https://developer.android.com/guide/topics/manifest/manifest-element.html#package
+const DISALLOWED_ANDROID_PACKAGE_CHARS_REGEX = /[^ a-zA-Z0-9_\.]/g;
```
PEConn: I don't think there should be a space in here.

andreban: Yes, that shouldn't be there. Removed and added a test for input with spaces.

PR #117, opened 2020-02-29
Adaptative Icons Improvements

### LINKED ISSUE #112: Icons with transparent background are displayed with black background
I have a PWA with maskable icon with transparent background.

After I pack it using `llama-pack` the resulting TWA icon displays with black background.

I expected the white background (as WebAPK) or transparent (as in Chrome on desktop).

Is it ok?

I've checked the all icons in all `/app/src/main/res/` subfolders like `drawable-xxxhdpi`. All of them have the transparent background.

I'm not sure if the fact that my icon is maskable important or not.

### LINKED ISSUE #115: Generated Adaptative Icon doesn't have parallax effect
**Describe the bug**
the resulting TWA icon looks adaptive (if I change the mask shape in Android setting, it changes correctly too). But for some reason there is no parallax visual effect (see https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive for more info).

**To Reproduce**
1. Run llama-pack init on `llama-pack init --manifest https://maskable-transparent-bg-bug.web.app/manifest.json`
2. Run llama-pack build
3. Install the APK do an Android device

**Expected behavior**
The Icon should have parallax effect.

**Additional context**
The WebAPK icon is adaptive and has the parallax visual effect. 
https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive for more info

andreban:
- Background for transparent icons is white.
- ~~Moves icon to the foreground layer, instead of background.~~
- Adds `android.useAndroidX=true` to `build.gradle` to make importing the project into Android Studio more seamless (unrelated to the icon).

Closes #112

### DIFF
--- a/packages/core/template_project/app/src/main/res/mipmap-anydpi-v26/ic_launcher.xml
+++ b/packages/core/template_project/app/src/main/res/mipmap-anydpi-v26/ic_launcher.xml
@@ -15,10 +15,11 @@
 -->
 <adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:aapt="http://schemas.android.com/aapt">
-    <background>
+    <background android:drawable="@android:color/white" />
+    <foreground>
         <aapt:attr name="android:drawable">
             <layer-list>
-                <item android:drawable="@android:color/black" />
+                <item android:drawable="@android:color/white" />
                 <item android:drawable="@mipmap/ic_maskable"
 					android:top="12.75dp"
 					android:right="12.75dp"
@@ -26,6 +27,5 @@
 					android:bottom="12.75dp" />
             </layer-list>
         </aapt:attr>
-    </background>
-    <foreground android:drawable="@android:color/transparent" />
+    </foreground>
 </adaptive-icon>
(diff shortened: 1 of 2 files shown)


andreban:
@NotWoods would mind taking a quick look?

FluorescentHallucinogen:
I believe we should align with WebAPK behavior (fill the transparent space with white color, not black).

See also my https://github.com/GoogleChromeLabs/llama-pack/issues/112#issuecomment-592962889 and https://github.com/GoogleChromeLabs/llama-pack/issues/112#issuecomment-592963645.

The first is WebAPK, the second is TWA:

webapk-twa

andreban:
Thanks for the input, @NotWoods 

#### Regarding the white background:
The [spec](https://www.w3.org/TR/appmanifest/#icon-masks) does **advise** against use transparent icons for the maskable icon, but does provide guidance on how user agents should handle it:

> If the icon contains transparent pixels, the user agent MUST composite the icon onto a solid color (eg. white) of the user agent's choice.

It does seem that we could switch the fill color from black to white and remain true to the spec. This would also replicate the behaviour for WebAPKs and the Android Platform in general.

#### Regarding the parallax effect
Would you have more details on that discussion? If this is likely to make it to the spec, it may be worth keeping it as it is. Otherwise, I think making it compatible with WebAPKs and with what Android expects as a platform worth it.  I'll double check how it works in WebAPKs and understand the design decisions on their side too.

NotWoods:
https://github.com/w3c/manifest/pull/657#issuecomment-426848210

andreban:
I have discussed with the WebAPK team and reverted the icon back to the background layer, as it's currently being considered to be changed there.

Issue #116, opened 2020-02-29
Shortcut items with no icon URL are ignored


FluorescentHallucinogen:
Repro: `llama-pack init --manifest https://shortcut-bugs.web.app/manifest.json`.

Shortcut items with no icon URL are just ignored. `llama-pack` should provide some default icon in this case instead of ignoring them.

andreban:
Looking at [this section](https://github.com/GoogleChromeLabs/llama-pack/blob/master/packages/core/template_project/app/build.gradle#L127-L132) of the template project, it seems this is by design. 

The icon is somewhat required on Android, as if the developer doesn't set it in `shortcuts.xml` Android will generate a default icon, an Android head, that is not meaningful to the action.

It doesn't sound to me that llama-pack would be better to provide a default icon that would be able to convey any better meaning to the actions. 

I do think we should alert developers of missing data in the shortcuts, so they know it's not being included and know how to fix it. 

@rayankans will know more about the current design.

FluorescentHallucinogen:
> if the developer doesn't set it in shortcuts.xml Android will generate a default icon, an Android head

Just checked, can't confirm the case of item with no icon, the situation with WebAPK is very sad. I've opened https://crbug.com/1057380.

rayankans:
Hi, I explained on the bug how the case with WebAPKs is related to caching and the fact that this hasn't been enabled on all channels.

As for llama-pack, as @andreban mentioned, this was a design decision. I could be convinced otherwise. Maybe a warning that it's being skipped is best for now? I believe that we we show developers a summary of the shortcuts included though when they are generating the TWA.

andreban:
+1 to a warning of with what is being skipped and a summary of the shortcuts being included.

FluorescentHallucinogen:
The web app manifest spec doesn't require icon for shortcut items, it's optional (see https://w3c.github.io/manifest/#example-11).

If the developer declared 5 shortcut items (and 2 of them have no icons) in the web app manifest for PWA, then he expects to see these 5 shortcut items for TWA (and 2 of them should have default icon — android heads).

I.e. we should skip items only if

* the property is required by spec (e.g. `"url"` or `"name"`), but for some reason is missing

or

* the property is present, but invalid

Since #103 is not fixed yet, if `"icons"`/`"src"` is SVG file, it breaks the build. As an idea, we could temporarily skip shortcut items like this until SVG will be supported.

[...]

PR #122, opened 2020-03-11
Add console warnings if a shortcut has been skipped.

rayankans:
Resolves https://github.com/GoogleChromeLabs/bubblewrap/issues/116


rayankans:
@andreban Do you mind taking a look

andreban:
Have you considered using [`Log.ts`](https://github.com/GoogleChromeLabs/bubblewrap/blob/master/packages/core/src/lib/Log.ts) to output the warnings? There's a `Log.warn()` method that will format it as a warning (yellow) and with a tag.

rayankans:
Changed this to use Log.ts

packages/core/src/lib/TwaManifest.ts
```diff
@@ -100,6 +101,8 @@ export class TwaManifest {
   generatorApp: string;
   webManifestUrl?: URL;
 
+  private static log: Log = new Log('twa-manifest');
```
andreban: nit: I've been passing a log object in the constructor, but initialising with a default value. This helps with tests so those don't pollute the console. But I'm ok to keep like this too.

rayankans: It's being used from within a static function though. I could make it a module-level constant but that won't help with the pollution issue.

Issue #105, opened 2020-02-13
Please add support for TWA WebView Fallback


naveedahmed1:
When a user starts a Trusted Web Activity (TWA) based application, a browser that supports TWAs may not be available on the device. In those situations, the default implementation will fallback to opening the URL in a Custom Tab or the default browser, if a browser that supports Custom Tabs is not found. This approach may be undesirable as it breaks the full-screen experience for users.

I would therefore suggest you to please add an option to the CLI that allow user to enable `WebView Fallback` with a single switch.

This could be based on 
https://github.com/GoogleChrome/android-browser-helper/tree/5a0accde2341108fcc7eb7d0b0e0d7ce45a29962/demos/twa-webview-fallback

andreban:
Thanks @naveedahmed1 

The goal is to add the WebView fallback directly to android-browser-helper. The issue we're tracking this is here: https://github.com/GoogleChrome/android-browser-helper/issues/15, and the demo is the first step in that direction. We will start working on that soon. 

Once that is done, llama-pack will automatically get it.

naveedahmed1:
That's great, thanks for the update @andreban :)

andreban:
#105 adds this

1. Update `twa-manifest.json` and include / update the `fallbackStrategy` field to `webview`. 
2. Run `npx @bubblewrap/cli update` to apply the changes to the template
3.  Run `npx @bubblewrap/cli build` to generate the APK.

We still default to `customtab`, as it provides better compatibility for Web APIs. We strongly recommend  to comprehensively test the PWA inside the WebView when using it.

naveedahmed1:
`bubblewrap` is the new name for `@llama-pack cli` right?

naveedahmed1:
Can you please also add the basic instruction about this library to project read me file https://github.com/GoogleChromeLabs/bubblewrap/blob/master/README.md covering at least these two topics:

- Creating new project
- Updating existing project
- Different available settings

andreban:
Work on more comprehensive docs will start soon.

naveedahmed1:
Should a new project generated through `@bubblewrap/cli` already have the option `fallbackStrategy`?

I just installed `@bubblewrap/cli` and generated a new app, but `twa-manifest` doesn't contain `fallbackStrategy`.

andreban:
We havent' published a version with the feature yet. Will publish it this week.

PR #126, opened 2020-03-18
Updates android-browser-helper and enables WebView fallback on config

andreban:
- Updates to using android-browser-helper-1.2.0
- Adds an option to enable the WebView fallback on the twa-manifest.json

### DIFF
--- a/packages/core/src/spec/lib/TwaManifestSpec.ts
+++ b/packages/core/src/spec/lib/TwaManifestSpec.ts
@@ -134,6 +134,7 @@ describe('TwaManifest', () => {
         shortcuts: [{name: 'name', shortName: 'shortName', url: '/', chosenIconUrl: 'icon.png'}],
         webManifestUrl: 'https://pwa-directory.com/manifest.json',
         generatorApp: 'test',
+        fallbackType: 'webview',
       } as TwaManifestJson;
       const twaManifest = new TwaManifest(twaManifestJson);
       expect(twaManifest.packageId).toEqual(twaManifestJson.packageId);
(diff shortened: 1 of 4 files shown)


packages/core/src/spec/lib/TwaManifestSpec.ts
```diff
@@ -72,6 +72,7 @@ describe('TwaManifest', () => {
       expect(twaManifest.generateShortcuts())
           .toBe('[[name:\'shortcut name\', short_name:\'short\',' +
             ' url:\'https://pwa-directory.com/launch\', icon:\'shortcut_0\']]');
+      expect(twaManifest.fallbackType).toBe('customtabs');
```
PEConn: Can we add a test as well to check that `webview` works?

packages/core/template_project/app/build.gradle
```diff
@@ -36,7 +36,10 @@ def twaManifest = [
     shortcuts: <%= generateShortcuts() %>,
     // The duration of fade out animation in milliseconds to be played when removing splash screen.
     splashScreenFadeOutDuration: <%= splashScreenFadeOutDuration %>,
-    generatorApp: '<%= generatorApp %>' // Apllication that generated the Android Project
+    generatorApp: '<%= generatorApp %>', // Apllication that generated the Android Project
```
PEConn: nit: type in "Apllication"

packages/core/template_project/app/src/main/AndroidManifest.xml
```diff
@@ -21,6 +21,7 @@
      the abbreviated format. -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="<%= packageId %>">
+    <uses-permission android:name="android.permission.INTERNET"/>
```
PEConn: Did this get added in by Android Studio or do we need it for WebView? It's seems a little ugly including it even if we don't need WebView, but on the other hand it's not that much overhead.

andreban: This needs to be added for the WebView to work on Android 9 and below. We conditionally add it, only when the fallback is type `webview`

andreban: Missed the "could" on the text. Fixed that.

PEConn: If it's easy enough let's do it, I don't think it's blocking though.

andreban: Done

packages/core/template_project/app/build.gradle
andreban: done

packages/core/src/spec/lib/TwaManifestSpec.ts
andreban: done

andreban:
Closes #105

PR #133, opened 2020-04-03
Adds instructions to update the CLI config

andreban:


packages/cli/README.md
```diff
@@ -47,6 +47,10 @@ Create a folder and extract the downloaded file into it.
 When running `bubblewrap` for the first time, it will ask where it can find the JDK and Android command
 line tools. So, take note of the location where both were decompressed.
 
+### Updating the location of the JDK and / or the Android command line tools.
+If the location for the JDK or the Android command line tools have been setup with the wrong path or
+if their location has changed after the initial configuration, the location for either of those can
+be changed by editing the configuration file at `${USER_HOME}/.llama-pack/llama-pack-config.json`.
```
abdonrd, line 53: We should change the name to `.bubblewrap/bubblewrap-config.json`?

andreban: There isn't a huge benefit on updating the file name and it will either cause current configs to be lost or need extra code to handle both cases.

PR #135, opened 2020-04-09
Implements input validation for the AplicationId

andreban:
- Implements validation and tests
- Improves some error messages on the CLI

### DIFF
--- a/packages/cli/src/lib/cmds/init.ts
+++ b/packages/cli/src/lib/cmds/init.ts
@@ -36,19 +36,19 @@ async function confirmTwaConfig(twaManifest: TwaManifest): Promise<TwaManifest> 
       type: 'input',
       message: 'Domain being opened in the TWA:',
       default: twaManifest.host,
-      validate: async (input): Promise<boolean> => notEmpty(input, 'host cannot be empty'),
+      validate: async (input): Promise<boolean> => notEmpty(input, 'host'),
     }, {
       name: 'name',
       type: 'input',
       message: 'Name of the application:',
       default: twaManifest.name,
-      validate: async (input): Promise<boolean> => notEmpty(input, 'name cannot be empty'),
+      validate: async (input): Promise<boolean> => notEmpty(input, 'name'),
     }, {
       name: 'launcherName',
       type: 'input',
       message: 'Name to be shown on the Android Launcher:',
       default: twaManifest.launcherName,
-      validate: async (input): Promise<boolean> => notEmpty(input, 'Launcher name cannot be empty'),
+      validate: async (input): Promise<boolean> => notEmpty(input, 'Launcher name'),
     }, {
       name: 'themeColor',
       type: 'input',
@@ -66,7 +66,7 @@ async function confirmTwaConfig(twaManifest: TwaManifest): Promise<TwaManifest> 
       type: 'input',
       message: 'Relative path to open the TWA:',
       default: twaManifest.startUrl,
-      validate: async (input): Promise<boolean> => notEmpty(input, 'URL cannot be empty'),
+      validate: async (input): Promise<boolean> => notEmpty(input, 'URL'),
     }, {
       name: 'iconUrl',
       type: 'input',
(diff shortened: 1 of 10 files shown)


packages/core/src/spec/lib/utilSpec.ts
```diff
@@ -137,4 +137,34 @@ describe('util', () => {
...
+  describe('#validatePackageId', () => {
+    it('returns true for valid packages', () => {
+      expect(util.validatePackageId('com.pwa_directory.appspot.com')).toBeTrue();
+      expect(util.validatePackageId('com.pwa1directory.appspot.com')).toBeTrue();
+    });
+
+    it('returns false for packages with invalid characters', () => {
+      expect(util.validatePackageId('com.pwa-directory.appspot.com')).toBeFalse();
+      expect(util.validatePackageId('com.pwa@directory.appspot.com')).toBeFalse();
+      expect(util.validatePackageId('com.pwa*directory.appspot.com')).toBeFalse();
```
PEConn, line 150: Maybe it's worth adding an example with some weird unicode character (maybe a unicode character that looks like a full stop but isn't).

packages/core/src/spec/lib/utilSpec.ts
```diff
@@ -137,4 +137,34 @@ describe('util', () => {
...
+
+    it('packages with less than 2 sections return false', () => {
+      expect(util.validatePackageId('com')).toBeFalse();
+      expect(util.validatePackageId('')).toBeFalse();
+    });
+
+    it('packages starting with non-letters return false', () => {
+      expect(util.validatePackageId('com.1char.twa')).toBeFalse();
+      expect(util.validatePackageId('1com.char.twa')).toBeFalse();
+      expect(util.validatePackageId('com.char.1twa')).toBeFalse();
```
PEConn, line 167: One that starts with an underscore?

packages/cli/src/lib/cmds/init.ts
```diff
@@ -91,19 +91,26 @@ async function confirmTwaConfig(twaManifest: TwaManifest): Promise<TwaManifest>
...
-      validate: notEmpty,
+      validate: async (input): Promise<boolean> =>
+        notEmpty(input, 'KeyStore location cannot be empty'),
     }, {
       name: 'keyAlias',
       type: 'input',
       message: 'Key name:',
       default: twaManifest.signingKey.alias,
-      validate: notEmpty,
+      validate: async (input): Promise<boolean> => notEmpty(input, 'Key alias cannot be empty'),
```
PEConn, line 113: Could you just pass the name to `notEmpty` and have `notEmpty` take care of the "cannot be empty" part? (As it is, for half of these you say "cannot be empty" and the other half you have "can't be empty".

packages/cli/src/spec/inputHelpersSpec.ts
```diff
@@ -0,0 +1,64 @@
...
+    });
+  });
+
+  describe('#validateKeyPassword', () => {
+    it('throws Error for empty string', async () => {
+      await expectAsync(inputHelpers.validateKeyPassword('')).toBeRejectedWithError();
+      await expectAsync(inputHelpers.validateKeyPassword(' ')).toBeRejectedWithError();
+    });
+
+    it('throws Error for empty string', async () => {
```
PEConn, line 37: description is wrong.

packages/cli/src/spec/inputHelpersSpec.ts
```diff
@@ -0,0 +1,64 @@
...
+      await expectAsync(inputHelpers.validateKeyPassword('a')).toBeRejectedWithError();
+      await expectAsync(inputHelpers.validateKeyPassword('abc')).toBeRejectedWithError();
+      await expectAsync(inputHelpers.validateKeyPassword('abcde')).toBeRejectedWithError();
+      await expectAsync(inputHelpers.validateKeyPassword('abcde ')).toBeRejectedWithError();
+    });
+
+    it('throws Error for empty string', async () => {
```
PEConn, line 44: here as well

packages/core/src/spec/lib/utilSpec.ts
andreban: `\u2024` (`․`) should do it.

packages/cli/src/spec/inputHelpersSpec.ts
andreban: done

packages/cli/src/spec/inputHelpersSpec.ts
andreban: done

packages/cli/src/lib/cmds/init.ts
andreban: done

packages/core/src/spec/lib/utilSpec.ts
andreban: done

packages/core/src/spec/lib/utilSpec.ts
andreban: done

PR #148, opened 2020-05-01
Fixes --sdk_root parameter on Windows

andreban:
- on Windows, escaping --sdk_root with `"` causes the sdkmanager
to error. Removing the escape char on Windows.
- We'll need to throw an error when a path with spaces is passed
in the config, but working on that on another PR.


andreban:
Closes #149

Issue #155, opened 2020-05-03
Update quick documentation for android CLI tools


amanintech:
Guide to suggest adding android-CLI tools to the tools folder

Which occurred in issue #147 

Fix
add a step by step more explained procedures to a quick start guide

amanintech:
I can do this

andreban:
@amanintech, I just found another issue with the current version of the Android SDK Tools:

This [issue](https://issuetracker.google.com/issues/150888434) is preventing it to output a signed APK to be uploaded to the Play Store.

I don't think there's a fix for it right now, and there's not much I can do on the Bubblewrap side. The workaround seems to be download an use an older version of the SDK Tools.
 
https://dl.google.com/android/repository/sdk-tools-windows-4333796.zip do it.

This seems to be an issue on Windows only and seems to work fine on Linux and MacOS

amanintech:
Also tried installing the apk it failed 
@andreban  do you think these issues are connected somehow

andreban:
Yes. The APK needs to signed for it to be installed on the device. `app-release-unsigned-aligned.apk` is not signed and can't be uploaded to the store or installed on a device. The correct file should be `app-release-signed.apk`.

andreban:
It seems #4 is getting more critical.

PR #158, opened 2020-05-03
Updated Documentation for more detailed steps on requirements

amanintech:
This update to documentation includes the fix for #155 that is to make to simpler for a person to install and specify the requirements.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -41,17 +41,27 @@ versions are incompatible with the Android command line tools.
 ### Get the Android command line tools
 Download a version of Android command line tools that is compatible with your OS from
 [https://developer.android.com/studio#command-tools](https://developer.android.com/studio#command-tools).
-Create a folder and extract the downloaded file into it.
+Create a folder and extract the downloaded file into it. This will further install the androidSdk and android SDK manager without needing to install the whole Android IDE.
 
 ### Tell Bubblewrap where the JDK and Android command line tools are
 When running `bubblewrap` for the first time, it will ask where it can find the JDK and Android command
 line tools. So, take note of the location where both were decompressed.
+- The **OpenJDK** path should contain `bin,com,include ..`
+- The **AndroidSDK** path should contain `tools` which should have `bin, cli`
 
 ### Updating the location of the JDK and / or the Android command line tools.
 If the location for the JDK or the Android command line tools have been setup with the wrong path or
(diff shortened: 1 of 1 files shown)


packages/cli/README.md
```diff
@@ -80,7 +90,7 @@ bubblewrap build
 
 When building the project for the first time, the Android Build Tools will need to be installed.
 The tool will inkove the installation process for the build tools. Make sure to read and accept
-the license agreement before proceeding.
+the license agreement before proceeding. This process will install the other required files inside the root decompressed directory of the android CLI package.
```
andreban, line 93: s/root decompressed directory/decompressed root directory

packages/cli/README.md
```diff
@@ -41,17 +41,27 @@ versions are incompatible with the Android command line tools.
...
+- The **AndroidSDK** path should contain `tools` which should have `bin, cli`
 
 ### Updating the location of the JDK and / or the Android command line tools.
 If the location for the JDK or the Android command line tools have been setup with the wrong path or
 if their location has changed after the initial configuration, the location for either of those can
 be changed by editing the configuration file at `${USER_HOME}/.llama-pack/llama-pack-config.json`.
 
+#### Sample llama-pack-config.json
+```
+{
```
andreban, line 59: Can you fix indentation for the example. I also think it may be better to add the full path. eg: `/user/home/bubblewrap-user/open-jdk`.

packages/cli/README.md
```diff
@@ -41,17 +41,27 @@ versions are incompatible with the Android command line tools.
 ### Get the Android command line tools
 Download a version of Android command line tools that is compatible with your OS from
 [https://developer.android.com/studio#command-tools](https://developer.android.com/studio#command-tools).
-Create a folder and extract the downloaded file into it.
+Create a folder and extract the downloaded file into it. This will further install the androidSdk and android SDK manager without needing to install the whole Android IDE.
 
 ### Tell Bubblewrap where the JDK and Android command line tools are
 When running `bubblewrap` for the first time, it will ask where it can find the JDK and Android command
 line tools. So, take note of the location where both were decompressed.
+- The **OpenJDK** path should contain `bin,com,include ..`
```
andreban, line 49: Those two examples are a bit disconnected. Can we add something like:
"To ensure if you are taking note of the correct location, check if each directory contains the following files:"

Issue #151, opened 2020-05-03
bubblewrap init command - ERROR Cannot read property 'readFile' of undefined


JotinhaBR:
**Describe the bug**
ERROR Cannot read property 'readFile' of undefined

**Screenshots**
![image](https://user-images.githubusercontent.com/6248317/80908302-22fb6b80-8cf5-11ea-90b6-e3974d350aa2.png)

**Desktop**
 - OS: Windows 10
 - Browser chrome

andreban:
Can you check what happens if you run with npx?
```
npx @bubblewrap/cli init --manifest=https://stopstop.planetsweb.com.br/jogar/manifest.json
```

Which NodeJs version are you using?

Obrigado,
Andre

JotinhaBR:
Acredito que você é Brasileiro, bom obrigado, estou me sentindo sentindo estupido na cabeça toda e não percebido que um pacote de vocês requer que tenha o NodeJS 10, atualize o pacote e funcionou!

Obrigado :)

EN:
Good thanks, i'm feeling stupid in the head and don't realize that a package of yours requires that you have NodeJS 10, update the package and it worked!

Thank you :)

andreban:
Correto 😄 . 

I have filed #152 so we display a better error message when an unsupported NodeJs version is being used.

PR #160, opened 2020-05-05
CLI checks minimum supported Node.js version

### LINKED ISSUE #152: Check for valid NodeJs version on startup
The project requires NodeJs 10 and above. Check for the NodeJs version on start-up and throw error if an unsupported version is being used.

andreban:
Uses `semver` to check the Node.js version and throws an error if an unsupported version is used.

Capture

Closes #152


andreban:
@JotinhaBR FYI, as this should help developers understand what's happening on #151

PR #161, opened 2020-05-11
Adds a `bubblewrap install` command

### LINKED ISSUE #49: Create a `llama-pack debug` command
The debug command would correctly setup the device to debug the TWA and launch it. 

- Create the debug configuration, as described [here](https://developers.google.com/web/updates/2019/02/using-twa#debugging). 
- Run `./gradlew installDebug`
- We may also want to add a command (or an option) to clean the debug config, by wiping out `/data/local/tmp/chrome-command-line` on the device.
- Caveat: Developers still need to enable the relevant flag on `chrome://flags`

andreban:
`bubblewrap install` allows developers to install an APK to a
connected development device to test the application without
the need to find the correct adb commands.

Closes #49

### DIFF
--- a/packages/cli/src/lib/cmds/help.ts
+++ b/packages/cli/src/lib/cmds/help.ts
@@ -29,6 +29,7 @@ const HELP_MESSAGES = new Map<string, string>(
         'update .............. updates an existing TWA Project with the latest bubblewrap template',
         'validate ............ validates if an URL matches the PWA Quality Criteria for Trusted' +
             ' Web Activity',
+        'install ............. installs the output application to a connected device',
       ].join('\n')],
       ['init', [
         'Usage:',
@@ -69,6 +70,17 @@ const HELP_MESSAGES = new Map<string, string>(
         '',
         '',
         'bubblewrap validate --url=[pwa-url]',
+      ].join('\n')],
+      ['install', [
+        'Usage:',
+        '',
+        '',
+        'bubblewrap install',
+        '',
+        '',
+        'Options: ',
+        '--apkFile ................. path to the APK file to be isntalled. Defaults to ' +
+            '"./app-release-signed.apk"',
       ].join('\n')],
     ],
 );
--- a/packages/cli/src/lib/cmds/help.ts
+++ b/packages/cli/src/lib/cmds/help.ts
@@ -81,6 +81,7 @@ const HELP_MESSAGES = new Map<string, string>(
         'Options: ',
         '--apkFile ................. path to the APK file to be isntalled. Defaults to ' +
             '"./app-release-signed.apk"',
+        '--verbose ................. prints the adb command being executed',
       ].join('\n')],
     ],
 );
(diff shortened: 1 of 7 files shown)


packages/cli/src/lib/cmds/install.ts
```diff
@@ -0,0 +1,35 @@
...
+export async function install(
+    args: ParsedArgs, config: Config, log = new Log('build')): Promise<boolean> {
+  const jdkHelper = new JdkHelper(process, config);
+  const androidSdkTools = new AndroidSdkTools(process, config, jdkHelper, log);
+  const apkFile = args.apkFile || DEFAULT_APK_FILE;
+
+  // parameter 0 would be the path to 'node', followed by `bubblewrap.js` at 1, then `install` at
+  // 2. So, we want to start collecting args from parameter 3 and ignore any a possible
+  // `--apkFile`, which is specific to install. Extra parameters are passed through to `adb`.
+  const originalArgs = process.argv.slice(3).filter((v) => !v.startsWith(APK_FILE_PARAM));
```
PEConn, line 32: I think this is clever, but I'm not 100% sure what we're trying to do. We're making a `bubblewrap install` behave almost exactly like `adb install`, but letting the user enter `--apkFile` anywhere (although that does seem to be optional). It seems that we're trying to be very helpful but I'm wondering if we could end up being confusing.

packages/cli/src/lib/cmds/help.ts
```diff
@@ -70,6 +71,17 @@ const HELP_MESSAGES = new Map<string, string>(
...
       ].join('\n')],
+      ['install', [
+        'Usage:',
+        '',
+        '',
+        'bubblewrap install',
+        '',
+        '',
+        'Options: ',
+        '--apkFile ................. path to the APK file to be isntalled. Defaults to ' +
```
PEConn, line 82: What do you think about adding some option (maybe like `debug` or `verbose`) that prints out the adb command we're executing?

packages/core/src/spec/lib/androidSdk/AndroidSdkToolsSpec.ts
```diff
@@ -146,4 +146,52 @@ describe('AndroidSdkTools', () => {
...
+      {platform: 'win32',
+        expectedCwd: [
+          '"C:\\Users\\user\\android-sdk\\platform-tools\\adb"',
+          'install',
+          'app-release-signed.apk',
+        ]},
+    ];
+
+    tests.forEach((test) => {
+      it(`Build the correct install commandon ${test.platform}`, async () => {
```
PEConn, line 173: nit: typo

packages/cli/src/lib/cmds/install.ts
andreban: Someone using Bubblewrap may not have `adb` in ther `PATH`. The goal is just to make it easier to call `adb install`. It defaults to using the default output file name, so `bubblewrap install` should be enough in most cases. But if, for instance, you have more than one device connected, you may want to pass extra parameters to ADB, and this allows that.

packages/cli/src/lib/cmds/help.ts
andreban: done

PR #162, opened 2020-05-13
 Extract information from signing keys

andreban:
- Implements keytool -list and extracts fingerprints from the
  response.

### DIFF
--- a/packages/core/src/lib/jdk/KeyTool.ts
+++ b/packages/core/src/lib/jdk/KeyTool.ts
@@ -14,7 +14,7 @@
  *  limitations under the License.
  */
 
-import {existsSync, promises} from 'fs';
+import {existsSync, promises, fstat} from 'fs';
 import {execute} from '../util';
 import {JdkHelper} from './JdkHelper';
 import Log from '../Log';
@@ -93,6 +93,9 @@ export class KeyTool {
    * @returns {Promise<string>} the raw output of the `keytool --list` command
    */
   async list(keyOptions: KeyOptions): Promise<string> {
+    if (!existsSync(keyOptions.path)) {
+      throw new Error(`Couldn't find signing key at "${keyOptions.path}"`);
+    }
     const keyListCmd = [
       'keytool',
       '-list',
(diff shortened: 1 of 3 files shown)


packages/core/src/lib/jdk/KeyTool.ts
```diff
@@ -77,4 +85,65 @@ export class KeyTool {
...
+    lines.forEach((line) => {
+      line = line.trim();
+      fingerprintTags.forEach((tag) => {
+        if (line.startsWith(tag)) {
+          // a fingerprint line has the format <tag>: <value>. So, we account for the extra colon
+          // when substringing and then trim to remove whitespaces.
+          const value = line.substring(tag.length + 1, line.length).trim();
+          fingerprints.set(tag, value);
+        }
+      });
```
PEConn, line 143: This could just be:

    if (line.startsWith("SHA")) {
        const parts = str.split(":")
        fingerprints.set(parts[0], parts[1].trim());
    }

Although I think your version is probably nicer.

packages/core/src/spec/lib/jdk/KeyToolSpec.ts
```diff
@@ -0,0 +1,144 @@
...
+Creation date: 28 Jan 2019
+Entry type: PrivateKeyEntry
+Certificate chain length: 1
+Certificate[1]:
+Owner: CN=Test Test, OU=Test, O=Test, L=London, ST=London, C=GB
+Issuer: CN=Test Test, OU=Test, O=Test, L=London, ST=London, C=GB
+Serial number: ea67d3d
+Valid from: Mon Jan 28 14:58:00 GMT 2019 until: Fri Jan 22 14:58:00 GMT 2044
+Certificate fingerprints:
+   SHA1: 38:03:D6:95:91:7C:9C:EE:4A:A0:58:43:A7:43:A5:D2:76:52:EF:9B
```
PEConn, line 34: You could do:

    const SHA1_KEY = "....";
    const SHA256_KEY = "....";
    const LIST_OUTPUT = `
    ....
        SHA1: ${SHA1_KEY}
    ....`

So you don't have to repeat yourself in the test. Up to you though.

andreban: done.

PEConn:
Still GTLM

PR #179, opened 2020-05-20
Generates assetlinks.json when building the App

### LINKED ISSUE #5: Generate /.well-know/assetlinks.json
Helps developers to generate a correct assetlinks.json file, with the correct Fingerprint and package name.

andreban:
- Uses information from they keystore and the applicationId to
  generate the Digital Asset Links file that matches the key used
  to build the signed APK.

- If the developer opted into Play Signing by Play the fingerprint
  won't match and the Digital Asset Links validation will not
  validate

Closes #5

### DIFF
--- a/packages/cli/src/lib/cmds/build.ts
+++ b/packages/cli/src/lib/cmds/build.ts
@@ -14,8 +14,8 @@
  *  limitations under the License.
  */
 
-import {AndroidSdkTools, Config, DigitalAssetLinks, GradleWrapper, JdkHelper, Log, TwaManifest,
-  KeyTool} from '@bubblewrap/core';
+import {AndroidSdkTools, Config, DigitalAssetLinks, GradleWrapper, JdkHelper, KeyTool, Log,
+  TwaManifest} from '@bubblewrap/core';
 import * as inquirer from 'inquirer';
 import * as path from 'path';
 import * as fs from 'fs';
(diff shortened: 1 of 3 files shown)


packages/cli/src/lib/cmds/build.ts
```diff
@@ -14,10 +14,11 @@
  *  limitations under the License.
  */
 
-import {AndroidSdkTools, Config, GradleWrapper, JdkHelper, Log, TwaManifest}
-  from '@bubblewrap/core';
+import {AndroidSdkTools, Config, DigitalAssetLinks, GradleWrapper, JdkHelper, Log, TwaManifest,
+  KeyTool} from '@bubblewrap/core';
```
PEConn, line 18: Is there an ordering to these imports? If they're alphabetical, `KeyTool` is in the wrong place.

packages/cli/src/lib/cmds/build.ts
```diff
@@ -137,5 +139,31 @@ export async function build(
   );
 
   log.info(`Signed Android App generated at "${outputFile}"`);
+
+  try {
+    const digitalAssetLinksFile = './assetlinks.json';
```
PEConn, line 144: Is this going to be relative to the directory the tool was run from? Is that what we want?

packages/cli/src/lib/cmds/build.ts
```diff
@@ -137,5 +139,31 @@ export async function build(
   );
 
   log.info(`Signed Android App generated at "${outputFile}"`);
+
+  try {
```
PEConn, line 143: I'd say you probably want to move this into its own function, that way we can make the `if` on line 153 a guard clause and return early afterwards.

packages/cli/src/lib/cmds/build.ts
```diff
@@ -137,5 +139,31 @@ export async function build(
...
+    if (!sha256Fingerprint) {
+      log.warn('Could not find SHA256 fingerprint. Skipping generating "assetlinks.json"');
+    } else {
+      const digitalAssetLinks =
+      DigitalAssetLinks.generateAssetLinks(twaManifest.packageId, sha256Fingerprint);
+
+      await fs.promises.writeFile(digitalAssetLinksFile, digitalAssetLinks);
+
+      log.info(`Digital Asset Links file generated at ${digitalAssetLinksFile}`);
+      log.warn('If opting into "App Signing by Google Play" when uploading the Android app for ' +
```
PEConn, line 162: I understand why you're doing this, but I don't like the idea of the tool outputting a warning that the user can't get rid of. I'm not sure there's a better way though.

andreban: We can do info and just hightlight it with colors. How does that sound?

packages/cli/src/lib/cmds/build.ts
andreban: +1. I thought about it when writing the if statement but ended up forgetting. 🤦

packages/cli/src/lib/cmds/build.ts
andreban: I noticed that, but for now I'm following what we're doing for `./app-release-signed.apk`. I do think a flag to output somewhere different in the future will be useful (will be a larger change than this, as it'll touch other steps).

packages/cli/src/lib/cmds/build.ts
andreban: VS Code doesn't automatically sort it, do I've been doing this manually after auto-imports. Forgot about this one.

packages/cli/src/lib/cmds/build.ts
PEConn: I think it's more that even if the user does everything right, they're going to have a big highlighted warning every time they run the tool.

I think longer term we should have something in twa-config for Play Signing - either a "isUsingPlaySigning: false" the user can set or a "playSigningSignatures: '...'" they can use to make things work right. It's probably outside of the scope for this change though.

packages/cli/src/lib/cmds/build.ts
```diff
@@ -78,6 +79,36 @@ async function startValidation(): Promise<PwaValidationResult> {
...
+    });
+
+    const sha256Fingerprint = keyInfo.fingerprints.get('SHA256');
+    if (!sha256Fingerprint) {
+      log.warn('Could not find SHA256 fingerprint. Skipping generating "assetlinks.json"');
+      return;
+    }
+
+    const digitalAssetLinks =
+    DigitalAssetLinks.generateAssetLinks(twaManifest.packageId, sha256Fingerprint);
```
PEConn, line 100: nit: indent.

andreban: done

PR #189, opened 2020-06-01
Adds Web Vitals metrics to validation

andreban:
Annotation 2020-06-01 203839

### DIFF
--- a/packages/cli/src/lib/pwaValidationHelper.ts
+++ b/packages/cli/src/lib/pwaValidationHelper.ts
@@ -25,25 +25,23 @@ export function printValidationResult(validationResult: PwaValidationResult, log
 
   const accessibilityValue = validationResult.scores.accessibility.printValue;
 
-  const fcpValue = getColor(validationResult.scores.firstContentfulPaint);
   const lcpValue = getColor(validationResult.scores.largestContentfulPaint);
   const fidValue = getColor(validationResult.scores.firstInputDelay);
   const clsValue = getColor(validationResult.scores.cumulativeLayoutShift);
 
   log.info('');
   log.info(underline('Quality Criteria scores'));
-  log.info(`Lighthouse Performance score: ......... ${performanceValue}`);
-  log.info(`Lighthouse PWA check: ................. ${pwaValue}`);
+  log.info(`Lighthouse Performance score: ................... ${performanceValue}`);
+  log.info(`Lighthouse PWA check: ........................... ${pwaValue}`);
   log.info('');
   log.info(underline('Web Vitals'));
-  log.info(`First Contentful Paint (FCP) .......... ${fcpValue}`);
-  log.info(`Largest Contentful Paint (LCP) ........ ${lcpValue}`);
-  log.info(`First Input Delay (FID) ............... ${fidValue}`);
-  log.info(`Cumulative Layout Shift (CLS) ......... ${clsValue}`);
+  log.info(`Largest Contentful Paint (LCP) .................. ${lcpValue}`);
+  log.info(`Maximum Potential First Input Delay (Max FID) ... ${fidValue}`);
+  log.info(`Cumulative Layout Shift (CLS) ................... ${clsValue}`);
   log.info('');
   log.info(underline('Other scores'));
-  log.info(`Lighthouse Accessibility score......... ${accessibilityValue}`);
+  log.info(`Lighthouse Accessibility score................... ${accessibilityValue}`);
   log.info('');
   log.info(underline('Summary'));
-  log.info(bold(`Overall result: ....................... ${overallStatus}`));
+  log.info(bold(`Overall result: ................................. ${overallStatus}`));
 }
--- a/packages/cli/src/lib/pwaValidationHelper.ts
+++ b/packages/cli/src/lib/pwaValidationHelper.ts
@@ -1,6 +1,15 @@
-import {PwaValidationResult} from '@bubblewrap/validator';
-import {red, green, bold, underline, gray} from 'colors';
+import {PwaValidationResult, ScoreResult} from '@bubblewrap/validator';
+import {red, green, bold, underline, yellow} from 'colors';
 import {Log} from '@bubblewrap/core';
+
+function getColor(score: ScoreResult): string {
+  switch (score.status) {
+    case 'PASS': return green(score.printValue);
+    case 'WARN': return yellow(score.printValue);
+    case 'FAIL': return red(score.printValue);
+    default: return score.printValue;
+  }
+}
 
 export function printValidationResult(validationResult: PwaValidationResult, log: Log): void {
   log.info('');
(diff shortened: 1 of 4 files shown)


packages/cli/src/lib/pwaValidationHelper.ts
```diff
@@ -1,31 +1,46 @@
...
 
   log.info('');
   log.info(underline('Quality Criteria scores'));
   log.info(`Lighthouse Performance score: ......... ${performanceValue}`);
   log.info(`Lighthouse PWA check: ................. ${pwaValue}`);
   log.info('');
+  log.info(underline('Web Vitals'));
+  log.info(`First Contentful Paint (FCP) .......... ${fcpValue}`);
+  log.info(`Largest Contentful Paint (LCP) ........ ${lcpValue}`);
+  log.info(`First Input Delay (FID) ............... ${fidValue}`);
```
PEConn, line 41: Technically isn't this Total Blocking Time, not First Input Delay?

https://web.dev/vitals/#lab-tools-to-measure-core-web-vitals

packages/validator/src/lib/PwaValidator.ts
```diff
@@ -80,6 +93,26 @@ export class PwaValidator {
...
+
+    // Web Vitals Scores
+    const fcpScore =
+      psiResult.lighthouseResult.audits.metrics.details.items[0]['firstContentfulPaint'];
+    const fcpStatus = 'PASS';
+
+    const lcpScore =
+      psiResult.lighthouseResult.audits.metrics.details.items[0]['largestContentfulPaint'];
+    const lcpStatus =
+        this.getStatus(Math.round(lcpScore / 100) * 100, MIN_LCP_PASS_SCORE, MIN_LCP_WARN_SCORE);
```
PEConn, line 105: What are we doing here? Are we trying to make the output number end with two zeros (eg `1234` -> `1200`)?

packages/validator/src/lib/psi/PsiResult.ts
```diff
@@ -30,6 +30,20 @@ export type LighthouseCategories = {
...
+  [key: string]: number;
+}
+
+export type LighthouseAudit = {
+  id: string;
+  title: string;
+  description: string;
+  details: {
+    type: string;
+    items: LighthouseAuditDetailsItem[];
```
PEConn, line 43: Could you add a comment that PSI returns a list here but we only ever use the first item?

packages/validator/src/lib/psi/PsiResult.ts
```diff
@@ -30,6 +30,20 @@ export type LighthouseCategories = {
   [key in LighthouseCategoryName]: LighthouseCategory;
 };
 
+export type LighthouseAuditDetailsItem = {
+  [key: string]: number;
```
PEConn, line 34: Can we not just hard code the keys we care about like `firstContentfulPaint` and `largestContentfulPaint`? I feel that would make the code easier to understand.

packages/validator/src/lib/psi/PsiResult.ts
```diff
@@ -30,6 +30,20 @@ export type LighthouseCategories = {
   [key in LighthouseCategoryName]: LighthouseCategory;
 };
 
+export type LighthouseAuditDetailsItem = {
```
PEConn, line 33: I'm wondering if there's a better name for this object - something like `AuditResults` maybe? "Item" implies that there's only one thing so it's a bit odd to have a map which is called XxxItem.

packages/validator/src/spec/PwaValidatorSpec.ts
```diff
@@ -99,5 +114,193 @@ describe('PwaValidator', () => {
...
               '?url=https%3A%2F%2Fexample.com%2F');
     });
+
+    // FCP Tests
+    it('FCP is PASS when 0', async () => {
+      const psiResult = mockPsiResult(0.8, 1.0, {
+        firstContentfulPaint: 0,
+        largestContentfulPaint: 0,
+        maxPotentialFID: 0,
+        cumulativeLayoutShift: 0,
```
PEConn, line 124: Could you use the `{ ... oldArray }` syntax here? I'm not 100%, but I think this bit could be rewritten as:

```
const psiResult = mockPsiResult(0.8, 1.0, {
  firstContentfulPaint: 0,
  ...WEB_VITALS_SCORES,
});
```
So it makes it a bit more obvious which lines we care about and which ones are just the default.

(Maybe `firstContentfulPaint` needs to come after the `WEB_VITALS_SCORES`, I'm not 100% sure. Do tell me if you try it.)

packages/validator/src/spec/PwaValidatorSpec.ts
```diff
@@ -99,5 +114,193 @@ describe('PwaValidator', () => {
...
+        firstContentfulPaint: 10000,
+        largestContentfulPaint: 0,
+        maxPotentialFID: 0,
+        cumulativeLayoutShift: 0,
+      });
+      const pwaValidator = mockPwaValidator(psiResult);
+      const result = await pwaValidator.validate(new URL('https://example.com'));
+      expect(result.scores.firstContentfulPaint.status).toBe('PASS');
+      expect(result.scores.firstContentfulPaint.printValue).toBe('10.0 s');
+    });
```
PEConn, line 142: Don't we want a test for FCP is `WARN` or `FAIL` to complete the set?

packages/cli/src/lib/pwaValidationHelper.ts
andreban: Lighthouse calls this `maxPotentialFID`. Maybe rename this to "Maximum Potential First Input Delay (Max FID)"?

packages/validator/src/lib/PwaValidator.ts
andreban: Yes.

packages/validator/src/spec/PwaValidatorSpec.ts
andreban: It does work (though `WEB_VITALS_SCORES` needs to come first). Thanks for the suggestion!

packages/validator/src/spec/PwaValidatorSpec.ts
andreban: FCP isn't actually on https://web.dev/vitals/, so we don't have thresholds defined. I'm wondering it's better not to show it at all (not sure why I added in the first place)

PEConn: Oh right. Yeah, I'd vote for getting rid of it just for code (and output) simplicity.

packages/validator/src/lib/PwaValidator.ts
PEConn: OK - that's a little bit non-obvious at first glance - could you add some nicely named convenience method?

packages/cli/src/lib/pwaValidationHelper.ts
PEConn: Yeah, that sounds good - hope it fits though.

packages/validator/src/lib/PwaValidator.ts
andreban: Moved this to a function.

packages/validator/src/lib/psi/PsiResult.ts
andreban: changed.

packages/validator/src/lib/psi/PsiResult.ts
andreban: Added
```typescript
export type LighthouseMetricAuditNames = 'firstContentfulPaint' | 'largestContentfulPaint' |
    'maxPotentialFID' | 'cumulativeLayoutShift';
```

packages/validator/src/lib/psi/PsiResult.ts
andreban: added

packages/validator/src/spec/PwaValidatorSpec.ts
andreban: done

packages/validator/src/lib/psi/PsiResult.ts
```diff
@@ -30,17 +33,18 @@ export type LighthouseCategories = {
   [key in LighthouseCategoryName]: LighthouseCategory;
 };
 
-export type LighthouseAuditDetailsItem = {
-  [key: string]: number;
+export type LighthouseMetricAudit = {
```
PEConn, line 36: Is there some technical reason this can't just be:

```
export type LighthouseMetricAudit = {
  firstContentfulPaint: number;
  largestContentfulPaint: number;
  maxPotentialFID: number;
  cumulativeLayoutShift: number:
}
```

PEConn: (also, if we were to go this route, I'm wondering whether we want to type the numbers - eg "percent", "milliseconds" or something)

packages/validator/src/lib/PwaValidator.ts
```diff
@@ -95,22 +94,20 @@ export class PwaValidator {
     const accessibilityScore = psiResult.lighthouseResult.categories.accessibility.score;
 
     // Web Vitals Scores
-    const fcpScore =
-      psiResult.lighthouseResult.audits.metrics.details.items[0]['firstContentfulPaint'];
+    const lighthouseAuditMetrics = psiResult.lighthouseResult.audits.metrics.details.items[0];
+    const fcpScore = lighthouseAuditMetrics.firstContentfulPaint;
+    const roundedFcpScore = PwaValidator.roundToNearestMultiple(fcpScore, 100);
```
PEConn, line 99: Is "`PwaValidator.`" needed here?

packages/validator/src/lib/psi/PsiResult.ts
andreban: Oh, gotcha - I thought the concern was the unbounded string (this approach does give us nice autocomplete).

The reason I went this route was two things: 1 - there are dozens of keys on this object on the JSON. 2 - The same keys are used in other places in the JSON, and this would allow us to re-use if we need in the future.

I'll move to your suggestion for now and we can review in the future if we need to change.

packages/validator/src/lib/PwaValidator.ts
andreban: It is, as it's a static method in PwaValidator.

packages/cli/src/lib/pwaValidationHelper.ts
andreban: Done. it fits - we're around 65 columns with that.

Screenshot 2020-06-03 at 10 26 04

packages/validator/src/lib/psi/PsiResult.ts
andreban: done

PR #191, opened 2020-06-02
Runs `apksigner.jar` instead of `apksigner.bat` on win32

andreban:
- apksigner.bat depends on find_java.bat, which is broken in the
  Android CLI tools.
- This implements a workaround that runs `apksigner.jar` directly.

### DIFF
--- a/packages/core/src/lib/jdk/JdkHelper.ts
+++ b/packages/core/src/lib/jdk/JdkHelper.ts
@@ -18,6 +18,7 @@
 
 import {Config} from '../Config';
 import * as path from 'path';
+import {execute} from '../util';
 
 /**
  * Helps getting information relevant to the JDK installed, including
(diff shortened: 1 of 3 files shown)


packages/core/src/lib/jdk/JdkHelper.ts
```diff
@@ -50,6 +51,18 @@ export class JdkHelper {
     }
   }
 
+  /**
+   * Runs the `java` command, passing args as parameters.
+   */
+  async runJava(args: string[]): Promise<{stdout: string; stderr: string}> {
+    const java = this.process.platform === 'win32' ? '/bin/java.exe' : '/bin/java';
```
PEConn, line 58: Tangential question - does this cover 64 bit versions of Windows too?

andreban: It does: https://nodejs.org/api/os.html#os_os_platform

`win32` is a bad name. To check for 32bit / 64bit one would use `os.arch()`(https://nodejs.org/api/os.html#os_os_arch)

PEConn: Ah cool - this wasn't blocking, I was just curious.

Feel free to merge!

PR #193, opened 2020-06-04
Changed the images library back to sharp

chenlevy24:
now bubblewrap works with svg images

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,6 @@
     "typescript": "^3.9.2"
   },
   "dependencies": {
-    "sharp": "^0.25.3"
+    "@types/sharp": "^0.25.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -33,8 +33,5 @@
     "lerna-changelog": "^1.0.1",
     "npm": "^6.14.5",
     "typescript": "^3.9.2"
-  },
-  "dependencies": {
-    "@types/sharp": "^0.25.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -33,5 +33,8 @@
     "lerna-changelog": "^1.0.1",
     "npm": "^6.14.5",
     "typescript": "^3.9.2"
+  },
+  "dependencies": {
+    "sharp": "^0.25.3"
   }
 }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -38,6 +38,7 @@
     "@types/node-fetch": "^2.5.7",
     "@types/tar": "^4.0.3",
     "@types/valid-url": "^1.0.3",
+    "@types/sharp": "^0.25.0",
     "color": "^3.1.2",
     "extract-zip": "^1.7.0",
     "inquirer": "^7.0.4",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -33,5 +33,8 @@
     "lerna-changelog": "^1.0.1",
     "npm": "^6.14.5",
     "typescript": "^3.9.2"
+  },
+  "dependencies": {
```
andreban, line 37: We don't need sharp as a dependency for the root `package.json`. We should remove it here.

packages/core/package.json
```diff
@@ -41,7 +41,7 @@
     "color": "^3.1.2",
     "extract-zip": "^1.7.0",
     "inquirer": "^7.0.4",
-    "jimp": "^0.9.8",
+    "sharp": "^0.23.2",
```
andreban, line 44: We should also remove the dependency on `jimp` from the `packages/core/package.json`

packages/core/src/lib/TwaGenerator.ts
```diff
@@ -20,8 +20,8 @@ import fetch from 'node-fetch';
 import {template} from 'lodash';
 import {promisify} from 'util';
 import {TwaManifest, ShortcutInfo} from './TwaManifest';
-import Jimp = require('jimp');
 import Log from './Log';
+const sharp = require('sharp');
```
andreban, line 24: Sharp has Typescript types: https://www.npmjs.com/package/@types/sharp. We can install it with `npm install --save @types/sharp` in the packages/core folder.

The, we can use a proper import statement here (and also get type checking in the code)

packages/core/src/lib/TwaGenerator.ts
```diff
@@ -142,9 +142,12 @@ export class TwaGenerator {
   }
 
   private async saveIcon(data: Buffer, size: number, fileName: string): Promise<void> {
-    const image = await Jimp.read(data);
-    await image.resize(size, size);
-    await image.writeAsync(fileName);
+    console.log("ss");
```
andreban, line 145: Remove the dangling `console.log`

packages/core/src/lib/TwaGenerator.ts
```diff
@@ -142,9 +142,12 @@ export class TwaGenerator {
...
+    const image =await sharp(data)
```
andreban, line 146: nit: missing space after the `=` sign.

packages/core/src/lib/TwaGenerator.ts
```diff
@@ -142,9 +142,12 @@ export class TwaGenerator {
...
+    .resize(size)
```
andreban, line 147: nit: ident the lines below.

packages/core/src/lib/TwaGenerator.ts
```diff
@@ -16,12 +16,13 @@
...
 import * as path from 'path';
 import * as fs from 'fs';
+import * as sharp from 'sharp';
 import fetch from 'node-fetch';
 import {template} from 'lodash';
 import {promisify} from 'util';
 import {TwaManifest, ShortcutInfo} from './TwaManifest';
 import Log from './Log';
-const sharp = require('sharp');
+
```
PEConn, line 25: nit: Remove empty line.

PR #197, opened 2020-06-09
Fixed the bug that converted svg's in low quality

chenlevy24:


PEConn:
Looks good to me too.

PR #198, opened 2020-06-11
Change configuration file name

chenlevy24:
Change configuration file name from `llama-pack-config.json` to `bubblewrap-config.json`

### DIFF
--- a/packages/cli/src/lib/config.ts
+++ b/packages/cli/src/lib/config.ts
@@ -14,13 +14,14 @@
  *  limitations under the License.
  */
 
-import {existsSync} from 'fs';
+
 import {join} from 'path';
 import {homedir} from 'os';
 import {Config} from '@bubblewrap/core';
 import * as inquirer from 'inquirer';
+import {existsSync, promises} from 'fs';
 
-const DEFAULT_CONFIG_PATH = join(homedir(), '/.llama-pack/llama-pack-config.json');
+const DEFAULT_CONFIG_PATH = join(homedir(), '/.bubblewrap-config/bubblewrap-config.json');
 
 async function createConfig(): Promise<Config> {
   const result = await inquirer.prompt([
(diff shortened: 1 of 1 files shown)


packages/cli/src/lib/config.ts
```diff
@@ -14,13 +14,14 @@
  *  limitations under the License.
  */
 
-import {existsSync} from 'fs';
+
```
PEConn, line 17: nit: random newline.

packages/cli/src/lib/config.ts
```diff
@@ -14,13 +14,14 @@
...
 import {join} from 'path';
 import {homedir} from 'os';
 import {Config} from '@bubblewrap/core';
 import * as inquirer from 'inquirer';
+import {existsSync, promises} from 'fs';
 
-const DEFAULT_CONFIG_PATH = join(homedir(), '/.llama-pack/llama-pack-config.json');
+const DEFAULT_CONFIG_PATH = join(homedir(), '/.bubblewrap-config/bubblewrap-config.json');
```
PEConn, line 24: Do you mind making 4 variables:

`DEFAULT_CONFIG_FOLDER` - ~/.bubblewrap-config
`DEFAULT_CONFIG_NAME` - bubblewrap-config.json
`LEGACY_CONFIG_FOLDER` - ~/.llama-pack
`LEGACY_CONFIG_NAME` - llama-pack-config.json

And then using them below?

packages/cli/src/lib/config.ts
```diff
@@ -38,6 +39,14 @@ async function createConfig(): Promise<Config> {
 }
 
 export async function loadOrCreateConfig(path = DEFAULT_CONFIG_PATH): Promise<Config> {
+  if (!existsSync(path)) {
+    // no new named config file found
+    if (existsSync(join(homedir(), '/.llama-pack/llama-pack-config.json'))) {
+      // old named config file found - rename it and it's folder
+      promises.rename((join(homedir(), '/.llama-pack/llama-pack-config.json')), path);
```
PEConn, line 46: What we're doing here is a bit odd:

1. Rename/move `~/.llama-pack/llama-pack-config.json` to `~/.bubblewrap-config/bubblewrap-config.json`.
2. Rename `~/.llama-pack` to `~/.bubblewrap-config`.

So in step #2, are we not overwriting the `.bubblewrap-config` directory with the now empty `.llama-pack-config` directory?

packages/cli/src/lib/config.ts
```diff
@@ -38,6 +39,14 @@ async function createConfig(): Promise<Config> {
 }
 
 export async function loadOrCreateConfig(path = DEFAULT_CONFIG_PATH): Promise<Config> {
+  if (!existsSync(path)) {
+    // no new named config file found
+    if (existsSync(join(homedir(), '/.llama-pack/llama-pack-config.json'))) {
```
PEConn, line 44: It would be nice to log something to the console here, just to let the user know what's happening.

packages/cli/src/lib/config.ts
```diff
@@ -38,6 +39,14 @@ async function createConfig(): Promise<Config> {
...
+      // old named config file found - rename it and it's folder
```
PEConn, line 45: nit: It should be "its", not "it's" (I only know this by getting it wrong so many times).

PR #201, opened 2020-06-18
Added test for config.ts 

chenlevy24:


packages/cli/src/spec/configSpec.ts
```diff
@@ -0,0 +1,38 @@
...
+import {homedir} from 'os';
+import {existsSync} from 'fs';
+import {loadOrCreateConfig} from '../lib/config';
+import * as mock from 'mock-fs';
+
+describe('config', () => {
+  describe('#loadOrCreateConfig', () => {
+    it('checks if the file\'s name was changed in case it has the old name', async () => {
+      const dir = join(homedir(), '.llama-pack');
+      // creates a mock file systes
```
andreban, line 27: nit: s/systes/system

andreban:
This PR seems to incorporate the changes from #198, but doesn't address the comments on that issue. Can you address the feedback from #198 in this issue? We can probably just close the other PR once done.

package.json
```diff
@@ -32,6 +33,11 @@
     "lerna": "^3.21.0",
     "lerna-changelog": "^1.0.1",
     "npm": "^6.14.5",
+    "ts-node": "^8.10.2",
     "typescript": "^3.9.2"
+  },
+  "dependencies": {
+    "@bubblewrap/cli": "^1.3.1",
```
PEConn, line 40: @andreban I'm not 100% on how our dependencies work, but should this be here? If so it's a bit weird that it wasn't here before.

packages/cli/src/lib/config.ts
```diff
@@ -14,13 +14,14 @@
...
-import {existsSync} from 'fs';
+
 import {join} from 'path';
 import {homedir} from 'os';
 import {Config} from '@bubblewrap/core';
 import * as inquirer from 'inquirer';
+import {existsSync, promises} from 'fs';
 
-const DEFAULT_CONFIG_PATH = join(homedir(), '/.llama-pack/llama-pack-config.json');
+const DEFAULT_CONFIG_PATH = join(homedir(), '/.bubblewrap-config/bubblewrap-config.json');
```
PEConn, line 24: As Andre mentioned, I left a few comments on the previous PR for this file.

packages/cli/src/spec/configSpec.ts
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright 2019 Google Inc. All Rights Reserved.
```
PEConn, line 2: 2020, surely?

packages/cli/src/spec/configSpec.ts
```diff
@@ -0,0 +1,38 @@
...
+ *  limitations under the License.
+ */
+
+import {join} from 'path';
+import {homedir} from 'os';
+import {existsSync} from 'fs';
+import {loadOrCreateConfig} from '../lib/config';
+import * as mock from 'mock-fs';
+
+describe('config', () => {
```
PEConn, line 23: We are messing with the user's file system, so I do reckon we could probably do with a few more tests (sorry, I know I said something different offline, but I've had a bit more thought about it now). Primarily, we want to check that:

* If there's no old config, we don't touch anything.
* If there's only an old config in `./llama-pack`, we should move the config and remove `./llama-pack`.
* If there are other files in `./llama-pack`, we should not delete that directory after moving the config.
* If there exists both an old config and a new config, we shouldn't overwrite the new `bubblewrap-config.json` with the old one.

(Sorry all of these cases weren't present in the original bug.)

package.json
andreban: This is the root `package.json`. No need for this dependency here.

packages/cli/src/lib/config.ts
```diff
@@ -38,15 +42,30 @@ async function createConfig(): Promise<Config> {
...
-    }
+async function renameConfigIfNeeded(): Promise<void> {
+  if (existsSync(join(DEFAULT_CONFIG_FOLDER, DEFAULT_CONFIG_NAME))) return;
+  // no new named config file found
+  if (!existsSync(join(LEGACY_CONFIG_FOLDER, LEGACY_CONFIG_NAME))) return;
+  // old named config file found - rename it and its folder
+  console.log('An old named config file was found, changing it now');
+  const files = await fsPromises.readdir(LEGACY_CONFIG_FOLDER);
+  const numOfFiles = files.length;
+  if (numOfFiles != 1) {
```
andreban, line 53: nit: Can we add this comment to the numOfFiles check:

```suggestion
  // At this point, we know that's at least one file in the folder, `LEGACY_CONFIG_NAME`, so `numOfFiles`
  // will be at least `1`. We avoid destroying / moving other files in this folder.
  if (numOfFiles != 1) {
```

packages/cli/src/lib/config.ts
```diff
@@ -38,15 +42,30 @@ async function createConfig(): Promise<Config> {
...
-  if (!existsSync(path)) {
-    // no new named config file found
-    if (existsSync(join(homedir(), '/.llama-pack/llama-pack-config.json'))) {
-      // old named config file found - rename it and it's folder
-      promises.rename((join(homedir(), '/.llama-pack')), (join(homedir(), '/.bubblewrap-config')));
-      promises.rename((join(homedir(), '/.bubblewrap-config/llama-pack-config.json')), path);
-    }
+async function renameConfigIfNeeded(): Promise<void> {
+  if (existsSync(join(DEFAULT_CONFIG_FOLDER, DEFAULT_CONFIG_NAME))) return;
+  // no new named config file found
```
PEConn, line 47: Nit: Could you make the comments into proper sentences (capital letter and ending with a full stop).

packages/cli/src/lib/config.ts
```diff
@@ -38,15 +42,30 @@ async function createConfig(): Promise<Config> {
...
+  if (!existsSync(join(LEGACY_CONFIG_FOLDER, LEGACY_CONFIG_NAME))) return;
+  // old named config file found - rename it and its folder
+  console.log('An old named config file was found, changing it now');
```
PEConn, line 50: Can you use the log objects we have instead of just console.log?

packages/cli/src/lib/config.ts
```diff
@@ -38,15 +42,30 @@ async function createConfig(): Promise<Config> {
...
-export async function loadOrCreateConfig(path = DEFAULT_CONFIG_PATH): Promise<Config> {
-  if (!existsSync(path)) {
-    // no new named config file found
-    if (existsSync(join(homedir(), '/.llama-pack/llama-pack-config.json'))) {
-      // old named config file found - rename it and it's folder
-      promises.rename((join(homedir(), '/.llama-pack')), (join(homedir(), '/.bubblewrap-config')));
-      promises.rename((join(homedir(), '/.bubblewrap-config/llama-pack-config.json')), path);
-    }
+async function renameConfigIfNeeded(): Promise<void> {
+  if (existsSync(join(DEFAULT_CONFIG_FOLDER, DEFAULT_CONFIG_NAME))) return;
```
andreban, line 46: nit: We're calling `join(DEFAULT_CONFIG_FOLDER, DEFAULT_CONFIG_NAME)` and `join(LEGACY_CONFIG_FOLDER, LEGACY_CONFIG_NAME)` a couple of times - should we move those into variables?

packages/cli/src/lib/config.ts
```diff
@@ -38,15 +42,30 @@ async function createConfig(): Promise<Config> {
...
+  // no new named config file found
+  if (!existsSync(join(LEGACY_CONFIG_FOLDER, LEGACY_CONFIG_NAME))) return;
+  // old named config file found - rename it and its folder
+  console.log('An old named config file was found, changing it now');
+  const files = await fsPromises.readdir(LEGACY_CONFIG_FOLDER);
+  const numOfFiles = files.length;
+  if (numOfFiles != 1) {
+    // there are other files at the old config folder. we leave it and create a new one
+    await fsPromises.mkdir(DEFAULT_CONFIG_FOLDER);
+    fsPromises.rename(join(LEGACY_CONFIG_FOLDER, LEGACY_CONFIG_NAME),
```
PEConn, line 56: Do we not need `await` here as well? @andreban ?

packages/cli/src/spec/configSpec.ts
```diff
@@ -17,22 +17,91 @@
...
       mock({
-        [dir]: {
+        [LEGACY_CONFIG_FOLDER]: {
           'llama-pack-config.json': '{}',
         }});
       await loadOrCreateConfig();
-      expect(existsSync(join(homedir(), '/.bubblewrap-config/bubblewrap-config.json'))).toBeTrue();
-      expect(existsSync(join(homedir(), '/.llama-pack/llama-pack-config.json'))).toBeFalse();
+      // checks if the file name was changed
+      expect(existsSync(join(DEFAULT_CONFIG_FOLDER, DEFAULT_CONFIG_NAME))).toBeTrue();
```
PEConn, line 49: We shouldn't have lines that are longer than 100 characters, could you break them up and make them neat please?

packages/cli/src/spec/configSpec.ts
```diff
@@ -17,22 +17,91 @@
...
+              'llama-pack-config.json': '{"content":"some old content"}',
+            },
+            [DEFAULT_CONFIG_FOLDER]: {
+              'bubblewrap-config.json': '{"content":"some new content"}',
+            }});
+          await loadOrCreateConfig();
+          // checks if the file name was changed
+          expect(existsSync(join(DEFAULT_CONFIG_FOLDER, DEFAULT_CONFIG_NAME))).toBeTrue();
+          expect(existsSync(join(LEGACY_CONFIG_FOLDER, LEGACY_CONFIG_NAME))).toBeTrue();
+          const file1 =await fsPromises.readFile(
```
PEConn, line 98: nit: space before await.

packages/cli/src/lib/config.ts
andreban: yes

PR #203, opened 2020-06-19
Closes #180: Use monochrome icon for notifications

### LINKED ISSUE #180: Provide a notification icon for delegated notifications
If you create a TWA using bubblewrap with notification delegation turned on, the TWA doesn't provide a small icon for the notification, this means that although the notifications are shown by the TWA, they still use the small icon from the browser, for example:

2020-05-20_14-48-50-Notifications

We should check what the requirements are for the small icon and if we can automatically generate it (eg from the TWA icon), or at the least allow the user to specify it in their `twa-manifest.json`.

NotWoods:
Adds `monochromeIconUrl` following the same format as the generic & maskable icons. Android deals with transforming notification images so no additional work is needed to follow the spec! Additionally fixed up some indentation.

### DIFF
--- a/packages/cli/src/lib/cmds/init.ts
+++ b/packages/cli/src/lib/cmds/init.ts
@@ -76,16 +76,20 @@ async function confirmTwaConfig(twaManifest: TwaManifest): Promise<TwaManifest> 
     }, {
       name: 'maskableIconUrl',
       type: 'input',
-      message: 'URL to an image that is at least 512x512px to be used when generating ' +
-          'maskable icons',
+      message:
+        'URL to an image that is at least 512x512px to be used when generating maskable icons.' +
+        '\n\nMaskable icons should look good when their edges are removed by an icon mask. ' +
+        'They will be used to display adaptive launcher icons on the Android home screen.',
       default: twaManifest.maskableIconUrl,
       filter: (input): string | undefined => input.length === 0 ? undefined : input,
       validate: async (input): Promise<boolean> => input === undefined || await validateUrl(input),
     }, {
       name: 'monochromeIconUrl',
       type: 'input',
-      message: 'URL to an image that is at least 48x48px to be used when generating ' +
-          'monochrome icons',
+      message:
+        'URL to an image that is at least 48x48px to be used when generating monochrome icons.' +
+        '\n\nMonochrome icons should look good when displayed with a single color,' +
+        'the PWA\' theme_color. They will be used for notification icons.',
       default: twaManifest.monochromeIconUrl,
       filter: (input): string | undefined => input.length === 0 ? undefined : input,
       validate: async (input): Promise<boolean> => input === undefined || await validateUrl(input),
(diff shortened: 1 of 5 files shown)


packages/core/src/spec/lib/TwaManifestSpec.ts
```diff
@@ -87,6 +88,7 @@ describe('TwaManifest', () => {
       expect(twaManifest.startUrl).toBe('/');
       expect(twaManifest.iconUrl).toBeUndefined();
       expect(twaManifest.maskableIconUrl).toBeUndefined();
+      expect(twaManifest.monochromeIconUrl).toBeUndefined();
```
PEConn, line 91: Could you add a testcase (or change an existing one) where this is not undefined?

packages/core/src/lib/TwaGenerator.ts
```diff
@@ -256,6 +264,11 @@ export class TwaGenerator {
       await this.generateIcons(twaManifest.maskableIconUrl, targetDirectory, ADAPTIVE_IMAGES);
     }
 
+    // Generate notification images
+    if (twaManifest.monochromeIconUrl) {
+      await this.generateIcons(twaManifest.monochromeIconUrl, targetDirectory, NOTIFICATION_IMAGES);
```
PEConn, line 269: I'm wondering if it would be worth parallelizing any of this. I don't think we have an issue with performance at the moment though. @andreban thoughts?

packages/cli/src/lib/cmds/init.ts
```diff
@@ -81,6 +81,14 @@ async function confirmTwaConfig(twaManifest: TwaManifest): Promise<TwaManifest>
       default: twaManifest.maskableIconUrl,
       filter: (input): string | undefined => input.length === 0 ? undefined : input,
       validate: async (input): Promise<boolean> => input === undefined || await validateUrl(input),
+    }, {
+      name: 'monochromeIconUrl',
+      type: 'input',
+      message: 'URL to an image that is at least 48x48px to be used when generating ' +
+          'monochrome icons',
```
PEConn, line 88: Do you reckon we need a little more explanation here as to what the monochrome icon is? Specifically that it doesn't actually need to be monochrome, it just needs to look good when turned monochrome.

Also, who determines which colour it will be displayed.

(I know these are in the spec, but should we maybe link to that or just include a sentence of explanation?)

packages/core/src/lib/TwaGenerator.ts
andreban: I don't think we have perf issues for now. We can parallelize things, but I think this will require change other pieces of code and might be better suited for it's own pull-request.

packages/cli/src/lib/cmds/init.ts
andreban: My understanding is that we are using monochrome icons in the Web Manifest to generate notification icons in the Android App. It sounds to me that we inside the Bubblewrap we should call this "notification icon"?

PEConn:
LGTM, @andreban ?

PR #207, opened 2020-06-23
Clarifies project can be opened in Android Studio

andreban:


packages/cli/README.md
```diff
@@ -85,6 +85,9 @@ the values that should be used when building the Android project.
 It will also ask you for the details needed to generate a signing key, used to sign the
 app before uploading to the Play Store.
 
+> :grey_exclamation: The output from the `init` command is a regular Android project that can be opened
+and built using [Android Studio](https://developer.android.com/studio/).
```
PEConn, line 89: It's probably worth pointing out that if you want to upload the project built from Android Studio, you should use "Generate signed APK", not just build

Or maybe that you can modify it in Android Studio, but it's best to do the final build with bubblewrap.

andreban: Maybe link to the Android Studio documentation on how to sign an APK for production: https://developer.android.com/studio/publish/app-signing#sign-apk - we do recommend using Bubblewrap unless the developer has good Android development knowledge

PR #209, opened 2020-06-25
Add enableLocation for enabling location delegation

EiraGe:


packages/core/template_project/app/src/main/AndroidManifest.xml
```diff
@@ -96,6 +100,10 @@
         <activity android:name="com.google.androidbrowserhelper.trusted.WebViewFallbackActivity"
             android:configChanges="orientation|screenSize" />
 
+        <activity android:name="com.google.androidbrowserhelper.trusted.PermissionRequestActivity"
```
andreban, line 103: IIRC, this is part of an unreleased version of android-browser-helper, am I right? If that's the case, we also need to update the dependency on `app/build.gradle` when that version comes out.

packages/core/template_project/app/build.gradle
```diff
@@ -28,6 +28,7 @@ def twaManifest = [
     navigationColor: '<%= navigationColor.hex() %>', // The color used for the navigation bar.
     backgroundColor: '<%= backgroundColor.hex() %>', // The color used for the splash screen background.
     enableNotifications: <%= enableNotifications %>, // Set to true to enable notification delegation.
+    enableLocation: <%= enableLocation %>, // Set to the location permission type if enabled.
```
PEConn, line 31: Sorry, should this be a bool or a location permission type?

packages/core/template_project/app/src/main/AndroidManifest.xml
PEConn, line 103: We could also only include this Activity if enableLocation is true.

EiraGe:
I updated this to import the locationdelegation module and use LocationDelegationExtraCommandHandler if enableLocation is true (as we discussed in [the android-browser-helper change](https://github.com/GoogleChrome/android-browser-helper/pull/117#discussion_r447539863])

This change is depending on the alpha branch of android-browser-helper, so we don't need/shouldn't to merge it before the depending change released. (I'll update the version in `app/build.gradle` accordingly once we have the release)

@andreban and @PEConn , could you please take a look if the approach in this change looks ok?

packages/core/template_project/app/build.gradle
```diff
@@ -46,7 +47,7 @@ android {
     compileSdkVersion 29
     defaultConfig {
         applicationId "<%= packageId %>"
-        minSdkVersion 19
+        minSdkVersion 21
```
andreban, line 50: Is `minSdkVersion 21` a requirement for location delegation? If that's the case and it's not possible to maintain it, I'd only use 21 when notification delegation is enabled and user 19 otherwise. Sounds worth warning developers that enabling notification delegation will make the app incompatible with some devices too. According to Android Studio, we'd be dropping from supporting 98.1% of devices to 94.1%.

Screenshot 2020-07-30 at 13 52 41

packages/core/template_project/app/src/main/java/TwaDelegationService.java
```diff
@@ -0,0 +1,15 @@
+package <%= packageId %>;
```
andreban, line 1: We don't strictly need to template the TwaDelegationService at this point. eg: we could ship this implemenation with the location library and just swap which implementation we reference on the AndroidManifest.xml.

This approach does make it easier for us to mix and match other extensions, but I wanted to check if that's indeed what we have in mind with this approach.

EiraGe:
Updated to use the FeatureManager. 

@andreban PTAL Thanks :)

andreban (changes requested):
Left a note on the feature implementation. Other than that, looking good.

packages/core/src/lib/TwaManifest.ts
```diff
@@ -56,14 +56,20 @@ const DEFAULT_APP_VERSION_NAME = DEFAULT_APP_VERSION_CODE.toString();
...
 const DEFAULT_SIGNING_KEY_ALIAS = 'android';
 const DEFAULT_ENABLE_NOTIFICATIONS = false;
+const DEFAULT_ENABLE_LOCATIONS = false;
 const DEFAULT_GENERATOR_APP_NAME = 'unknown';
 
 export type FallbackType = 'customtabs' | 'webview';
 
 type Features = {
   appsFlyer?: AppsFlyerConfig;
+  locationDelegation: boolean;
```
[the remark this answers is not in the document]
EiraGe: (sorry for the delay) Added the empty LocationDelegationConfig

packages/core/src/lib/TwaManifest.ts
```diff
@@ -56,12 +57,14 @@ const DEFAULT_APP_VERSION_NAME = DEFAULT_APP_VERSION_CODE.toString();
 const DEFAULT_SIGNING_KEY_PATH = './android.keystore';
 const DEFAULT_SIGNING_KEY_ALIAS = 'android';
 const DEFAULT_ENABLE_NOTIFICATIONS = false;
+const DEFAULT_ENABLE_LOCATIONS = false;
```
PEConn, line 60: Shouldn't this be `DEFAULT_ENABLE_LOCATION` or `DEFAULT_ENABLE_LOCATION_DELEGATION`?

packages/core/src/lib/TwaManifest.ts
```diff
@@ -282,6 +285,7 @@ export class TwaManifest {
       enableNotifications: DEFAULT_ENABLE_NOTIFICATIONS,
       shortcuts: shortcuts,
       webManifestUrl: webManifestUrl.toString(),
+      features: {locationDelegation: DEFAULT_ENABLE_LOCATIONS ? {} : undefined},
```
PEConn, line 288: I'm not quite sure what you're doing here. How come we don't need to do the same for the appsFlyer feature?

PEConn (approved):
This looks good.

I'm wondering if we should have a test that checks the output DelegationService against some golden version

andreban:
Closing this, as #366 was based on it and recently merged.

PR #212, opened 2020-06-27
Add support for fullscreen display mode

### LINKED ISSUE #175: Add support for fullscreen/immersive mode
`android-browser-helper` is making the Immersive Mode API easier to use:  is making the Immersive Mode API easier to use: https://github.com/GoogleChrome/android-browser-helper/issues/95

1. During the `init` command, Bubblewrap should use read `display` attribute from the Web Manifest.
2. If the `display` is fullscreen, default the configuration to use immersive mode. Otherwise, default to regular mode.
3. Ask the developer to confirm the choice.

christianliebel:
Closes #175

### DIFF
--- a/packages/cli/src/lib/cmds/init.ts
+++ b/packages/cli/src/lib/cmds/init.ts
@@ -17,9 +17,9 @@
 import * as fs from 'fs';
 import Color = require('color');
 import * as inquirer from 'inquirer';
-import {Config, JdkHelper, KeyTool, Log, TwaGenerator, TwaManifest, util} from '@bubblewrap/core';
-import {validateColor, validateDisplayMode, validateKeyPassword, validateUrl,
-  notEmpty} from '../inputHelpers';
+import {Config, DISPLAY_MODES, JdkHelper, KeyTool, Log, TwaGenerator, TwaManifest,
+  util} from '@bubblewrap/core';
+import {validateColor, validateKeyPassword, validateUrl, notEmpty} from '../inputHelpers';
 import {ParsedArgs} from 'minimist';
 import {APP_NAME} from '../constants';
 
@@ -52,10 +52,10 @@ async function confirmTwaConfig(twaManifest: TwaManifest): Promise<TwaManifest> 
       validate: async (input): Promise<boolean> => notEmpty(input, 'Launcher name'),
     }, {
       name: 'display',
-      type: 'input',
+      type: 'list',
       message: 'Display mode to be used:',
       default: twaManifest.display,
-      validate: validateDisplayMode,
+      choices: DISPLAY_MODES,
     }, {
       name: 'themeColor',
       type: 'input',
--- a/packages/cli/src/lib/cmds/init.ts
+++ b/packages/cli/src/lib/cmds/init.ts
@@ -17,7 +17,7 @@
 import * as fs from 'fs';
 import Color = require('color');
 import * as inquirer from 'inquirer';
-import {Config, DISPLAY_MODES, JdkHelper, KeyTool, Log, TwaGenerator, TwaManifest,
+import {Config, DisplayModes, JdkHelper, KeyTool, Log, TwaGenerator, TwaManifest,
   util} from '@bubblewrap/core';
 import {validateColor, validateKeyPassword, validateUrl, notEmpty} from '../inputHelpers';
 import {ParsedArgs} from 'minimist';
@@ -55,7 +55,7 @@ async function confirmTwaConfig(twaManifest: TwaManifest): Promise<TwaManifest> 
       type: 'list',
       message: 'Display mode to be used:',
       default: twaManifest.display,
-      choices: DISPLAY_MODES,
+      choices: DisplayModes,
     }, {
       name: 'themeColor',
       type: 'input',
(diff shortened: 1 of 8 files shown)


christianliebel:
@andreban What should we do if the display mode is `browser` or `minimal-ui`?

packages/cli/src/lib/inputHelpers.ts
```diff
@@ -50,3 +50,11 @@ export async function validateUrl(url: string): Promise<boolean> {
   }
   return true;
 }
+
+export async function validateDisplayMode(displayMode: string): Promise<boolean> {
+  const validModes = ['browser', 'minimal-ui', 'standalone', 'fullscreen'];
```
andreban, line 55: Can we move this to TwaManifest.ts? We can define it like the following:
```typescript
const DISPLAY_MODES = ['standalone', 'fullscreen'];
type DisplayMode = typeof DISPLAY_MODES[number];

function asDisplayMode(input: string): DisplayMode | null {
    if (DISPLAY_MODES.indexOf(input) > -1) {
        return input as DisplayMode;
    }
    return null;
}

console.log(asDisplayMode('browser'));
console.log(asDisplayMode('invalid'));

```

packages/core/src/lib/TwaManifest.ts
```diff
@@ -94,6 +96,7 @@ export class TwaManifest {
   host: string;
   name: string;
   launcherName: string;
+  display: string;
```
andreban, line 99: with the `DisplayMode` type created, this becomes

```typescript
display: DisplayMode;
```

packages/core/src/lib/TwaManifest.ts
```diff
@@ -254,6 +258,7 @@ export class TwaManifest {
       name: webManifest['name'] || webManifest['short_name'] || DEFAULT_APP_NAME,
       launcherName: webManifest['short_name'] ||
         webManifest['name']?.substring(0, SHORT_NAME_MAX_SIZE) || DEFAULT_APP_NAME,
+      display: webManifest['display'] || DEFAULT_DISPLAY_MODE,
```
andreban, line 261: this then becomes:

```typescript
display: asDisplayMode(webManifest['display']) || DEFAULT_DISPLAY_MODE,
```

packages/core/src/spec/lib/TwaManifestSpec.ts
```diff
@@ -24,6 +24,7 @@ describe('TwaManifest', () => {
         'name': 'PWA Directory',
         'short_name': 'PwaDirectory',
         'start_url': '/?utm_source=homescreen',
+        'display': 'standalone',
```
andreban, line 27: Can we add a tests for a Web Manifest that doesn't have a display mode, using browser, minimal-ui and fullscreen?

packages/core/src/lib/types/WebManifest.ts
```diff
@@ -34,6 +34,7 @@ export interface WebManifestJson {
   name?: string;
   short_name?: string;
   start_url?: string;
+  display?: string;
```
andreban, line 37: Web could define a WebManifestDisplayMode:

```typescript
```

then

```typescript
  display?: WebManifestDisplayMode;
```

packages/cli/src/lib/cmds/init.ts
```diff
@@ -49,6 +50,12 @@ async function confirmTwaConfig(twaManifest: TwaManifest): Promise<TwaManifest>
       message: 'Name to be shown on the Android Launcher:',
       default: twaManifest.launcherName,
       validate: async (input): Promise<boolean> => notEmpty(input, 'Launcher name'),
+    }, {
+      name: 'display',
+      type: 'input',
```
andreban, line 55: Since we have a small set of choices, can we use the `type: 'list'` prompt here? More info on https://github.com/SBoudrias/Inquirer.js#list---type-list

packages/cli/src/spec/inputHelpersSpec.ts
```diff
@@ -61,4 +61,17 @@ describe('inputHelpers', () => {
...
+  describe('#validateDisplayMode', () => {
+    it('returns true for valid colors', async () => {
+      expect(await inputHelpers.validateDisplayMode('browser'));
+      expect(await inputHelpers.validateDisplayMode('minimal-ui'));
+      expect(await inputHelpers.validateDisplayMode('standalone'));
+      expect(await inputHelpers.validateDisplayMode('fullscreen'));
+    });
+
+    it('throws Error for invalid display modes', async () => {
+      await expectAsync(inputHelpers.validateColor('bogus')).toBeRejectedWithError();
```
andreban, line 74: probably work adding a validation for an empty string too

packages/core/src/spec/lib/TwaManifestSpec.ts
andreban: Can we also add tests for the constructor? We want to check
 - The default value is set when reading a TwaManifestJson that doesn't contain a value for `display`
 - The value is correctly set when reading a TwaManifestJson that contains a `display` value of `fullscreen`

packages/cli/src/spec/inputHelpersSpec.ts
christianliebel: This function is now removed, but I added checking the empty string to the new `asDisplayMode` tests.

PR #215, opened 2020-06-30
Put password in quotes to allow more characters in password

gugu:


andreban:
@gugu Thanks for the pull request! Much appreciated. The tests on the CI are failing though, can you check please? On the PR itself, it seems the case might be allowing spaces in the password instead of longer ones?

andreban:
@gugu the tests are not passing still. Can you check again? Also, can you let me know which OS are you testing on? I want to ensure it works across Windows, Linux and MacOS.

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
```diff
@@ -148,8 +148,8 @@ export class AndroidSdkTools {
     const apkSignerParams = [
       `sign --ks ${keystore}`,
       `--ks-key-alias ${alias}`,
-      `--ks-pass pass:${ksPass}`,
-      `--key-pass pass:${keyPass}`,
+      `--ks-pass "pass:${ksPass}"`,
```
PEConn, line 151: Shouldn't we be using single quotes instead of double quotes? Double quotes means that the password can contain spaces, but single quotes will escape other characters like `*` and `.`.

gugu:
Fixed. Still can be an issue if password includes ', but it is better than before

PEConn:
This is definitely a step in the right direction, but we probably should be properly escaping things that we pass to the command line.

gugu:
yes, no join in "exec" function in util.ts

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
```diff
@@ -148,8 +148,8 @@ export class AndroidSdkTools {
     const apkSignerParams = [
       `sign --ks ${keystore}`,
       `--ks-key-alias ${alias}`,
-      `--ks-pass pass:${ksPass}`,
-      `--key-pass pass:${keyPass}`,
+      `--ks-pass 'pass:${ksPass}'`,
```
andreban, line 151: re: @PEConn's comment on escaping single quotes, we'd need something like this. Worth adding some tests:

```suggestion
      `--ks-pass 'pass:${ksPass.replace(/'/g,'\\\'')}'`,
```

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
```diff
@@ -148,8 +148,8 @@ export class AndroidSdkTools {
...
+      `--key-pass 'pass:${keyPass}'`,
```
andreban, line 152: And same here:

```suggestion
      `--key-pass 'pass:${keyPass.replace(/'/g,'\\\'')}'`,
```

gugu:
@PEConn I did the right thing - added execFile function and started to run apksigner and java using this function. Tested on linux, works fine. exec function should not be used to run user-provided input

PEConn:
Thanks for that. This looks good to me.

PR #217, opened 2020-06-30
Fixes Unhandled Promise Rejections

andreban:
- Added missing `await` calls to fix Unhandled Promis Rejections
  when renaming the old config name to the new config name.
- Defaults CLI to the `help` command
- Uses `.bubblewrap` as the name for the config folder, in case we need to store other files in the future.

@chenlevy24 FYI

### DIFF
--- a/packages/cli/src/spec/configSpec.ts
+++ b/packages/cli/src/spec/configSpec.ts
@@ -23,7 +23,7 @@ import * as mock from 'mock-fs';
 import * as inquirer from 'inquirer';
 
 const DEFAULT_CONFIG_FOLDER = join(homedir(), '.bubblewrap');
-const DEFAULT_CONFIG_NAME = 'bubblewrap-config.json';
+const DEFAULT_CONFIG_NAME = 'config.json';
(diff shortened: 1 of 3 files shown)


packages/cli/src/spec/configSpec.ts
```diff
@@ -22,7 +22,7 @@ import {loadOrCreateConfig} from '../lib/config';
 import * as mock from 'mock-fs';
 import * as inquirer from 'inquirer';
 
-const DEFAULT_CONFIG_FOLDER = join(homedir(), '.bubblewrap-config');
+const DEFAULT_CONFIG_FOLDER = join(homedir(), '.bubblewrap');
 const DEFAULT_CONFIG_NAME = 'bubblewrap-config.json';
```
PEConn, line 26: Maybe we should just change the file name to `~/.bubblewrap/config.json`, `~/.bubblewrap/bubblewrap-config.json` is a bit tautological.

andreban: Yeah, makes sense. Renamed.

PR #220, opened 2020-07-02
Work in progress - added doctor command

### LINKED ISSUE #181: Create a `doctor` command
`bubblewrap doctor` should check the configuration file and if settings for the location of the Java JDK and Android SDK are valid, along with information on how to download and setup.

See https://flutter.dev/docs/get-started/install/windows#run-flutter-doctor for an example of what the command could look like.

chenlevy24:
Closes #181

### DIFF
--- a/packages/cli/src/lib/Cli.ts
+++ b/packages/cli/src/lib/Cli.ts
@@ -25,6 +25,7 @@ import {loadOrCreateConfig} from './config';
 import {major} from 'semver';
 import {version} from './cmds/version';
 import {BUBBLEWRAP_LOGO} from './constants';
+import {updateConfig} from './cmds/updateConfig';
 
 export class Cli {
   async run(args: string[]): Promise<boolean> {
@@ -70,6 +71,8 @@ export class Cli {
       }
       case 'install':
         return await install(parsedArgs, config);
+      case 'updateConfig':
+        return await updateConfig(parsedArgs);
       default:
         throw new Error(
             `"${command}" is not a valid command! Use 'bubblewrap help' for a list of commands`);
--- a/packages/cli/src/lib/Cli.ts
+++ b/packages/cli/src/lib/Cli.ts
@@ -26,6 +26,7 @@ import {major} from 'semver';
 import {version} from './cmds/version';
 import {BUBBLEWRAP_LOGO} from './constants';
 import {updateConfig} from './cmds/updateConfig';
+import {doctor} from './cmds/doctor';
 
 export class Cli {
   async run(args: string[]): Promise<boolean> {
@@ -73,6 +74,8 @@ export class Cli {
         return await install(parsedArgs, config);
       case 'updateConfig':
         return await updateConfig(parsedArgs);
+      case 'doctor':
+        return await doctor();
       default:
         throw new Error(
             `"${command}" is not a valid command! Use 'bubblewrap help' for a list of commands`);
--- a/packages/cli/src/lib/cmds/doctor.ts
+++ b/packages/cli/src/lib/cmds/doctor.ts
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2020 Google Inc. All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {ConsoleLog, Log} from '@bubblewrap/core';
+import {join} from 'path';
+import {existsSync, promises as fsPromises} from 'fs';
+import {loadOrCreateConfig} from '../config';
+import {enUS as messages} from '../strings';
+
+async function jdkDoctor(log: Log): Promise<boolean> {
+  const config = await loadOrCreateConfig();
+  const jdkPath = config.jdkPath;
+  if (!existsSync(jdkPath)) { // Checks if the given path is a real path.
+    log.error(messages.jdkPathIsNotValid);
+    return false;
+  }
+  const file = await fsPromises.readFile(join(jdkPath, 'release'), 'utf-8');
+  if (!file) {
+    log.error(messages.jdkPathIsNotCorrect);
+    return false;
+  };
+  if (file.indexOf('JAVA_VERSION="1.8') < 0) { // Checks if the jdk's version is 8 as needed.
+    log.error(messages.jdkIsNotsupported);
+    return false;
+  }
+  return true;
+}
+
+async function androidSdkDoctor(log: Log): Promise<boolean> {
+  const config = await loadOrCreateConfig();
+  const androidSdkPath = config.androidSdkPath;
+  if (!existsSync(androidSdkPath)) { // Checks if the given path is a real path.
+    log.error(messages.androidSdkPathIsNotValid);
+    return false;
+  }
+  if (!existsSync(join(androidSdkPath, 'build-tools'))) { // Checks if the path given is indeed.
+    log.error(messages.androidSdkPathIsNotCorrect);
+    return false;
+  };
+  return true;
+}
+
+export async function doctor(log: Log = new ConsoleLog('doctor')): Promise<boolean> {
+  const jdkResult = await jdkDoctor(log);
+  const androidSdkResult = await androidSdkDoctor(log);
+  if (jdkResult && androidSdkResult) {
+    log.info('Your jdkpath and androidSdkPath are valid.');
+  }
+  return jdkResult && androidSdkResult;
+}
--- a/packages/cli/src/lib/cmds/updateConfig.ts
+++ b/packages/cli/src/lib/cmds/updateConfig.ts
@@ -16,22 +16,19 @@
 
 import {Config, Log, ConsoleLog} from '@bubblewrap/core';
 import {ParsedArgs} from 'minimist';
-import {join} from 'path';
-import {homedir} from 'os';
 import {existsSync} from 'fs';
 import {loadOrCreateConfig} from '../config';
-
-const CONFIG_FILE_PATH = join(homedir(), '.bubblewrap-config/bubblewrap-config.json');
+import {DEFAULT_CONFIG_FILE_PATH} from '../config';
 
 async function updateAndroidSdkPath(path: string, log: Log): Promise<boolean> {
   if (!existsSync(path)) {
     log.error('Please enter a valid path.');
     return false;
   }
-  const config = loadOrCreateConfig();
-  const jdkPath = (await config).jdkPath;
+  const config = await loadOrCreateConfig();
+  const jdkPath = config.jdkPath;
   const newConfig = new Config(jdkPath, path);
-  newConfig.saveConfig(CONFIG_FILE_PATH);
+  newConfig.saveConfig(DEFAULT_CONFIG_FILE_PATH);
   return true;
 }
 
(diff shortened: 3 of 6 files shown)


packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,76 @@
...
+  return true;
+}
+
+async function androidSdkDoctor(log: Log): Promise<boolean> {
+  const config = loadOrCreateConfig();
+  const androidSdkPath = (await config).androidSdkPath;
+  // Checks if the given path is a real path.
+  if (!existsSync(androidSdkPath)) {
+    log.error('The androidSdkPath doesn\'t exist, please run the' +
+    'following command to update it:\n bubblewrap updateConfig' +
```
PEConn, line 56: You need a space at the start of the string on this line or at the end of the last one (otherwise you'll get `rune thefollowing command`).

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+  return true;
+}
+
+async function androidSdkDoctor(log: Log): Promise<boolean> {
+  const config = loadOrCreateConfig();
+  const androidSdkPath = (await config).androidSdkPath;
+  // Checks if the given path is a real path.
+  if (!existsSync(androidSdkPath)) {
+    log.error('The androidSdkPath doesn\'t exist, please run the' +
```
PEConn, line 55: Maybe it would be nice to do this:

```
log.error('The androidSdkPath doesn't exist, please run the following command:');
log.info('    bubblewrap updateConfig --androidPath path/to/sdk');
log.error('Then run bubblewrap doctor again.');
```

That way the command to run is somewhat highlighted (well, it's white instead of red).

What do you think? (If it turns out looking bad, we won't do it.)

@andreban since you've been working on making the UI pretty.
```

File: packages/cli/src/lib/cmds/doctor.ts [hunk 3]
```diff
@@ -0,0 +1,76 @@
...
+    ' 8(LTS)" at the link below: \n' +
+    'https://adoptopenjdk.net/releases.html?variant=openjdk8&jvmVariant=hotspot');
+    return false;
+  }
+  return true;
+}
+
+async function androidSdkDoctor(log: Log): Promise<boolean> {
+  const config = loadOrCreateConfig();
+  const androidSdkPath = (await config).androidSdkPath;
```
PEConn, line 52: Can we just do:

```
const config = await config;
const androidSdkPath = config.androidSdkPath;
```

PEConn: (same in the above function.)

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,76 @@
...
+    'following command to update it:\n' +
+    'bubblewrap updateConfig --jdkPath path/here. after thet run bubblewrap doctor again');
+    return false;
+  }
+  const file = await fsPromises.readFile(join(jdkPath, 'release'), 'utf-8');
+  if (!file) {
+    log.error('The jdkPath isn\'t correct, please run the' +
+    'following command to update it:' +
+    '\nbubblewrap updateConfig --jdkPath path/here, such that the folder "here" ' +
+    'contains the file "release.txt". after thet run bubblewrap doctor again');
```
PEConn, line 37: The file should be `release` not `release.txt`.

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,76 @@
...
+    return false;
+  };
+  // Checks if the jdk's version is 8 as needed.
+  if (file.indexOf('JAVA_VERSION="1.8') < 0) {
+    log.error('Wrong jdk version. You should download "OpenJDK' +
```
PEConn, line 42: You're a little inconsistent with where you're breaking the newlines into strings. Whichever IDE you're using, there should be a setting to enable a "vertical ruler" - set this up for 100 characters. It will put a line down the page/screen that tells you what the character limit is.

(eg, for VSCode, instructions: https://stackoverflow.com/a/29972073 )

packages/cli/src/lib/cmds/updateConfig.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+import {Config, Log} from '@bubblewrap/core';
+import {ParsedArgs} from 'minimist';
+import {join} from 'path';
+import {homedir} from 'os';
+import {existsSync, promises as fsPromises} from 'fs';
+import {loadOrCreateConfig} from '../config';
+
+const CONFIG_FILE_PATH = join(join(homedir(), '.bubblewrap-config/bubblewrap-config.json'));
```
PEConn, line 24: We already have this value defined in `config.ts`, we should export it use that instead otherwise we're in danger of someone changing one and not the other and things going wrong.

PEConn: Also, @andreban made a change to where the config file is stored (from `~/.bubblewrap-config/bubblewrap-config.json` to just `~/.bubblewrap/config.json`). You'll need to pull from master and merge/rebase your branch to get his changes.

https://github.com/GoogleChromeLabs/bubblewrap/commit/e38ea223a5f2b2b6a29ddc875502fb5275116460

packages/cli/src/lib/cmds/updateConfig.ts
```diff
@@ -0,0 +1,69 @@
...
+
+async function updateAndroidSdkPath(path: string, log: Log): Promise<boolean> {
+  if (!existsSync(path)) {
+    log.error('Please enter a valid path.');
+    return false;
+  }
+  const config = loadOrCreateConfig();
+  const jdkPath = (await config).jdkPath;
```
PEConn, line 32: Again, you can move the `await` up one line and the code becomes a bit cleaner.

packages/cli/src/lib/cmds/updateConfig.ts
```diff
@@ -0,0 +1,69 @@
...
+  const newConfig = new Config(jdkPath, path);
+  fsPromises.unlink(CONFIG_FILE_PATH);
```
PEConn, line 34: You shouldn't need to do this - newConfig.saveConfig should overwrite the existing one.

packages/cli/src/lib/cmds/updateConfig.ts
```diff
@@ -0,0 +1,69 @@
...
+  newConfig.saveConfig(CONFIG_FILE_PATH);
```
PEConn, line 35: Do we not need an `await` here?

packages/cli/src/lib/cmds/updateConfig.ts
```diff
@@ -0,0 +1,69 @@
...
+  return true;
+}
+
+async function updateJdkPath(path: string, log: Log): Promise<boolean> {
```
PEConn, line 39: You should be able to merge this function and the one above together - most of it is the same.

packages/cli/src/lib/cmds/updateConfig.ts
```diff
@@ -0,0 +1,69 @@
...
+  const newConfig = new Config(path, androidSdkPath);
+  fsPromises.unlink(CONFIG_FILE_PATH);
+  newConfig.saveConfig(CONFIG_FILE_PATH);
+  return true;
+}
+
+async function invoke(args: ParsedArgs, log: Log): Promise<boolean> {
+  if (args.jdkPath) {
+    await updateJdkPath(args.JdkPath, log);
+  } else if (args.androidSdkPath) {
```
PEConn, line 55: These don't need to be exclusive. Someone could run

```
bubblewrap updateConfig --jdkPath blah/blah --androidSdkPath blah2/blah
```

packages/cli/src/lib/Cli.ts
```diff
@@ -48,6 +50,10 @@ export class Cli {
     }
 
     switch (command) {
+      case 'updateConfig':
+        return await updateConfig(parsedArgs);
+      case 'doctor':
```
PEConn, line 55: You'll want to add these new commands to the help output (see line 58).

packages/cli/src/lib/cmds/doctor.ts
chenlevy24: Do we get the current config file like that? Don't we need to do loadConfig at least?

packages/cli/src/lib/Cli.ts
chenlevy24: I didn't understand exactly what did you mean, the user will need to type bubblewrap help doctor?

packages/cli/src/lib/cmds/doctor.ts
PEConn: Oh sorry, what I meant to type:

```
const config = await loadOrCreateConfig();
const androidSdkPath = config.androidSdkPath
```

packages/cli/src/lib/Cli.ts
PEConn: If you run `bubblewrap help`, it prints out all the `bubblewrap` commands and tells you how to use them. Since you're adding new bubblewrap commands, you should add them to the output of `bubblewrap help`.

packages/cli/src/lib/cmds/doctor.ts
chenlevy24: It doesn't let me remove the await. it says that androidSdkPath is from type Promise<config> and I have to use await

packages/cli/src/lib/cmds/doctor.ts
andreban: The thing about mixing `log.error` and `log.info` is that I think one goes to `stdout` and the other to `stderr`, which would lead to some weird behaviour if we redirect one of them to a file, for instance.

We are using the `colors` module for the `validator` already. Maybe something like:

```typescript
import {bold} from 'colors';
```
then
```typescript
log.error(`The androidSdkPath doesn't exist, please run the following command:
       ${colors.bold('bubblewrap updateConfig --androidPath path/to/sdk'})
Then run ${colors.bold('bubblewrap doctor') again.`);
```

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,70 @@
...
+  if (file.indexOf('JAVA_VERSION="1.8') < 0) {
+    log.error('Wrong jdk version. You should download "OpenJDK 8(LTS)" at the link below: \n' +
+    'https://adoptopenjdk.net/releases.html?variant=openjdk8&jvmVariant=hotspot');
+    return false;
+  }
+  return true;
+}
+
+async function androidSdkDoctor(log: Log): Promise<boolean> {
+  const config = loadOrCreateConfig();
```
andreban, line 48: It's more clear if you invoke `await` when calling the function:
```suggestion
  const config = await loadOrCreateConfig();
  const androidSdkPath = config.androidSdkPath;
```

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,70 @@
...
+  const androidSdkPath = (await config).androidSdkPath;
```
andreban, line 49: The this becomes:
```suggestion
  const androidSdkPath = config.androidSdkPath;
```

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,70 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {Log} from '@bubblewrap/core';
+import {join} from 'path';
+import {existsSync, promises as fsPromises} from 'fs';
+import {loadOrCreateConfig} from '../config';
+
+async function jdkDoctor(log: Log): Promise<boolean> {
```
andreban, line 22: Generally, we want to check for the correctness of the path when creating or updating a config to avoid having an invalid config in the first place. We should probably make those functions more generic and move them out of `doctor`.

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,70 @@
...
+  // Checks if the jdk's version is 8 as needed.
+  if (file.indexOf('JAVA_VERSION="1.8') < 0) {
+    log.error('Wrong jdk version. You should download "OpenJDK 8(LTS)" at the link below: \n' +
+    'https://adoptopenjdk.net/releases.html?variant=openjdk8&jvmVariant=hotspot');
+    return false;
+  }
+  return true;
+}
+
+async function androidSdkDoctor(log: Log): Promise<boolean> {
```
andreban, line 47: Same as the previous comment, but for android SDK

PR #230, opened 2020-07-08
Refactors user prompts

andreban:
- Creates a Prompt class to wrap Inquirer.
- Moves prompt related strings to a different file, which should
  make internationalization easier.
- Refactors inputHelpers to use Results.

### DIFF
--- a/packages/cli/src/lib/Prompt.ts
+++ b/packages/cli/src/lib/Prompt.ts
@@ -17,39 +17,57 @@
 import {Result} from '@bubblewrap/core';
 import * as inquirer from 'inquirer';
 
+/**
+ * A function that takes a `string`, validates and tries to convert to the type `T`, and returns a
+ * {@link Result}. If the conversion is successful, the result is `Ok` and unwrapping returns `T`.
+ * Otherwise, the result is `Error` and `unwrapError()` returns the underlying error.
+ * @param {string} input the value to be validated and converted.
+ * @returns {Result<T, Error>} an `Ok` {@link Result} that unwraps `T` if the validation and
+ * conversion are successful or an `Error` if it fails.
+ */
 export type ValidateFunction<T> = (input: string) => Result<T, Error>;
 
+/**
+ * A an interface that promps for different types of user input.
+ */
 export interface Prompt {
   /**
-   * Prompts for free text input
-   * @param message a short description of the input
-   * @param defaultValue a default value or null
-   * @param convertFunction a function that converts from the input string to the return value.
-   * @param validateFunction an option function to validate the input.
+   * Prompts for text input.
+   * @param {string} message a short description of the input.
+   * @param {string | null} defaultValue a default value or null.
+   * @param {ValidateFunction<T>} validateFunction a function to validate the input.
+   * @returns {Promise<T>} a {@link Promise} that resolves to the validated user input, converted
+   * to `T` by the `validateFunction`.
    */
   promptInput<T>(
     message: string,
     defaultValue: string | null,
-    convertFunction: ValidateFunction<T>,
+    validateFunction: ValidateFunction<T>,
   ): Promise<T>;
 
   /**
-   * Prompts for free text input
-   * @param message a short description of the input
-   * @param choices a list of choices to be displayed to the user
-   * @param defaultValue a default value or null
-   * @param convertFunction a function that converts from the input string to the return value.
-   * @param validateFunction an option function to validate the input.
+   * Displays a list of options to the user and prompts the user to choose one of them.
+   * @param {string} message a short description of the input.
+   * @param {string[]} choices a list of choices to be displayed to the user.
+   * @param {string | null} defaultValue a default value or null.
+   * @param {ValidateFunction<T>} validateFunction a function to validate the input.
+   * @returns {Promise<T>} a {@link Promise} that resolves to the validated user input, converted
+   * to `T` by the `validateFunction`.
    */
   promptChoice<T>(
     message: string,
     choices: string[],
     defaultValue: string | null,
-    convertFunction: ValidateFunction<T>,
+    validateFunction: ValidateFunction<T>,
   ): Promise<T>;
 
   /**
-   * Prompts for a password
+   * Prompts the user for a password. The text typed by the user is hidden and replaced by the `*`
+   * character.
+   * @param {string} message a short description of the input.
+   * @param {ValidateFunction<string>} validateFunction a function to validate the input.
+   * @returns {Promise<string>} a {@link Promise} that resolves to the user input validated by
+   * `validateFunction`.
    */
   promptPassword(
     message: string,
@@ -58,6 +76,10 @@ export interface Prompt {
 
   /**
    * Prompts a Yes/No dialog. Returns `true` for yes and `false` for no.
+   * @param {string} message a short description of the input.
+   * @param {boolean} defaultValue a default value.
+   * @returns {Promise<boolean>} a {@link Promise} that resolves to a {@link boolean} value. The
+   * value will the `true` if the user answers `Yes` and `false` for `No`.
    */
   promptConfirm(message: string, defaultValue: boolean): Promise<boolean>;
 }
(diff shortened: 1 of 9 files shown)


packages/cli/src/lib/Prompt.ts
```diff
@@ -0,0 +1,142 @@
...
+ */
+
+import {Result} from '@bubblewrap/core';
+import * as inquirer from 'inquirer';
+
+export type ValidateFunction<T> = (input: string) => Result<T, Error>;
+
+export interface Prompt {
+  /**
+   * Prompts for free text input
```
PEConn, line 24: nit: Maybe I haven't noticed this before, but shouldn't our comments all be complete sentences (with a full stop)?

packages/cli/src/lib/Prompt.ts
```diff
@@ -0,0 +1,142 @@
...
+   * @param message a short description of the input
+   * @param defaultValue a default value or null
+   * @param convertFunction a function that converts from the input string to the return value.
+   * @param validateFunction an option function to validate the input.
```
PEConn, line 28: This is a bit out of date, you only take one now.

packages/cli/src/lib/Prompt.ts
```diff
@@ -0,0 +1,142 @@
...
+   */
+  promptChoice<T>(
+    message: string,
+    choices: string[],
+    defaultValue: string | null,
+    convertFunction: ValidateFunction<T>,
+  ): Promise<T>;
+
+  /**
+   * Prompts for a password
```
PEConn, line 52: Could you add a comment about why this is different (eg, that the characters you type don't show up).

packages/cli/src/lib/Prompt.ts
```diff
@@ -0,0 +1,142 @@
...
+   * @param validateFunction an option function to validate the input.
+   */
+  promptInput<T>(
+    message: string,
+    defaultValue: string | null,
+    convertFunction: ValidateFunction<T>,
+  ): Promise<T>;
+
+  /**
+   * Prompts for free text input
```
PEConn, line 37: nit: It's not "free text input"

packages/cli/src/lib/Prompt.ts
```diff
@@ -0,0 +1,142 @@
...
+   * @param message a short description of the input
+   * @param choices a list of choices to be displayed to the user
+   * @param defaultValue a default value or null
+   * @param convertFunction a function that converts from the input string to the return value.
+   * @param validateFunction an option function to validate the input.
```
PEConn, line 42: Same as above.

packages/cli/src/lib/Prompt.ts
```diff
@@ -0,0 +1,142 @@
...
+}
+
+export class InquirerPrompt implements Prompt {
+  async promptInput<T>(
+      message: string,
+      defaultValue: string | null,
+      convertFunction: ValidateFunction<T>,
+  ): Promise<T> {
+    const validate = buildInquirerValidate(convertFunction);
+
```
PEConn, line 90: really silly nit: you've got newlines in this method but not in promptChoice or the others, it would be nice to be consistent.

packages/cli/src/lib/strings.ts
```diff
@@ -0,0 +1,93 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+type Messages = {
+  promptHostMessage: string;
```
PEConn, line 18: Maybe we could split these up - eg have a different object for errors and prompts?

Or maybe not - I think it's fine as it is.

packages/cli/src/lib/strings.ts
```diff
@@ -0,0 +1,93 @@
...
+  promptNewAppVersionName: string;
+  errorMaxLength: (maxLength: number, actualLength: number) => string;
+  errorMinLength: (minLength: number, actualLength: number) => string;
+  errorRequireHttps: string;
+  errorInvalidUrl: (url: string) => string;
+  errorInvalidColor: (color: string) => string;
+  errorInvalidDisplayMode: (displayMode: string) => string;
+}
+
+export const en: Messages = {
```
PEConn, line 48: If you make this `en_US` I could create an `en_GB` where we actually spell colour correctly.

packages/cli/src/lib/strings.ts
andreban: Yeah, I'm not sure it's worth splitting. I'm just using this a convention for where the messages are used. My guess is that we may have the same string used in different ways.

packages/cli/src/lib/strings.ts
andreban: done. We don't have internationali**s**ation yet, but this is a step in that direction.

packages/cli/src/lib/Prompt.ts
andreban: No, that's my bad - I keep forgetting about the full stops (I'll see if there's a lint rule for that). Fixed.

packages/cli/src/lib/Prompt.ts
andreban: Removed. Improved the descriptions overall.

packages/cli/src/lib/Prompt.ts
andreban: done.

packages/cli/src/lib/Prompt.ts
andreban: done.

packages/cli/src/lib/Prompt.ts
andreban: Added more details to all methods.

packages/cli/src/lib/Prompt.ts
andreban: Removed the new line. Thanks

packages/cli/src/lib/inputHelpers.ts
```diff
@@ -15,38 +15,164 @@
...
+import {ValidateFunction} from './Prompt';
+import {enUS as messages} from './strings';
+import {domainToASCII} from 'url';
 
-const MIN_KEY_PASSWORD_LENGTH = 6;
+/**
+ * A {@link ValidateFunction} that receives a {@link string} as input and resolves to a
+ * {@link Color} when successful.
+ * @param {string} color a string to be converted to a {@link Color}.
+ * @returns {Result<Color, Error} a results that resolves to a {@link Color} on success or
```
PEConn, line 29: nit: `Error` -> `Error>`

packages/cli/src/lib/inputHelpers.ts
```diff
@@ -15,38 +15,164 @@
...
+  try {
+    return Result.ok(new Color(color));
+  } catch (_) {
+    return Result.error(new Error(messages.errorInvalidColor(color)));
+  }
+};
+
+/**
+ * A {@link ValidateFunction} that receives a {@link string} as input and resolves to a
+ * {@link URL} when successful. If the string is empty, the validation fails and the the
```
PEConn, line 42: nit: the the -> the

packages/cli/src/lib/inputHelpers.ts
PEConn: (same a couple of times below as well)

packages/cli/src/lib/inputHelpers.ts
```diff
@@ -15,38 +15,164 @@
...
+
+  return validateUrl(url);
 }
 
-export async function validateColor(color: string): Promise<boolean> {
+/**
+ * Creates a {@link ValidateFunction<string>} that checks the input {@link string} against the
+ * constraints provided as parameters.
+ * @param {number?} minLength optional minimum length for the string.
+ * @param {number?} maxLength optional maximum length for the string.
```
PEConn, line 82: nit: duplicated line

andreban: those are actually different. One for `minLength` and the other for `maxLength`. Tried making the sentence shorter and highlighting where it's different.

packages/cli/src/lib/inputHelpers.ts
andreban: done

packages/cli/src/lib/inputHelpers.ts
andreban: Yea, had this across the whole code. Fixed.

PR #232, opened 2020-07-09
Improve prompts for the `init` command

andreban:
- Splits prompts into sections
- Adds more information to each section

See how it looks like below:
init

### DIFF
--- a/packages/cli/src/lib/Prompt.ts
+++ b/packages/cli/src/lib/Prompt.ts
@@ -32,6 +32,12 @@ export type ValidateFunction<T> = (input: string) => Result<T, Error>;
  */
 export interface Prompt {
   /**
+   * Prints a message to the user.
+   * @param message the message to be printed.
+   */
+  printMessage(message: string): Promise<void>;
+
+  /**
    * Prompts for text input.
    * @param {string} message a short description of the input.
    * @param {string | null} defaultValue a default value or null.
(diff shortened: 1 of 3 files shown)


packages/cli/src/lib/Prompt.ts
```diff
@@ -31,6 +31,12 @@ export type ValidateFunction<T> = (input: string) => Result<T, Error>;
  * A an interface that promps for different types of user input.
  */
 export interface Prompt {
+  /**
+   * Prints a message to the user.
+   * @param message the message to be printed.
+   */
+  printMessage(message: string): Promise<void>;
```
PEConn, line 38: I find it strange that you don't mark the interface method `async` (I know this is a TypeScript thing, not a mistake in your code). I guess the `Promise` does somewhat imply it, but the idea of having some subclass having an `async` implementation and some not seems weird to me.

andreban: +1. I often forget and mark interfaces as `async` and get the compiler complaining about it.

andreban:
> In the output gif you included, I still find it ugly that we print out all of the files that we generated - it just seems noisy. I guess if we want to keep them, we could make the paths relative to the project root, so they are less likely to get line wrapped.

I want to replace this with a progress bar, but will need some refactoring around the code. The idea is to get `TwaGenerator` to accept a callback function that reports the progress instead of it printing messages to the output. Focuses this PR on changes to `init` though.

andreban:
Created #236 to track

PR #238, opened 2020-07-14
Uses a progress bar when generating Projects

### LINKED ISSUE #236: Improve project generation output
Currently, `TwaGenerator` outputs the progress itself to Log. Besides printing information that may not be relevant to developer (eg: we care about the progress itself, not necessarily which files are being generated), the UI becomes cumbersome.

We propose;
1.  refactoring `TwaGenerator` so it can receive a `progress` function that is invoked throughout the generation of the project and allows the caller to format the progress.
2. modify `init` and `update` to output a progress bar that indicates the progress of `TwaGenerator`

andreban:
progress-bar
Closes #236

### DIFF
--- a/packages/core/src/lib/TwaGenerator.ts
+++ b/packages/core/src/lib/TwaGenerator.ts
@@ -103,6 +103,10 @@ interface Icon {
   data: Buffer;
 }
 
+export type twaGeneratorProgress = (progress: number, total: number) => void;
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const noOpProgress: twaGeneratorProgress = () => {};
+
 /**
  * Generates TWA Projects from a TWA Manifest
  */
@@ -118,7 +122,6 @@ export class TwaGenerator {
       sourceDir: string, targetDir: string, filename: string): Promise<void> {
     const sourceFile = path.join(sourceDir, filename);
     const destFile = path.join(targetDir, filename);
-    this.log.info('\t', destFile);
     await fsMkDir(path.dirname(destFile), {recursive: true});
     await fsCopyFile(sourceFile, destFile);
   }
@@ -135,7 +138,6 @@ export class TwaGenerator {
       sourceDir: string, targetDir: string, filename: string, args: object): Promise<void> {
     const sourceFile = path.join(sourceDir, filename);
     const destFile = path.join(targetDir, filename);
-    this.log.info('\t', destFile);
     await fsMkDir(path.dirname(destFile), {recursive: true});
     const templateFile = await fsReadFile(sourceFile, 'utf-8');
     const output = template(templateFile)(args);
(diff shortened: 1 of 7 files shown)


packages/core/src/lib/TwaGenerator.ts
```diff
@@ -224,54 +225,67 @@ export class TwaGenerator {
    * @param {String} targetDirectory the directory where the project will be created
    * @param {Object} twaManifest configurations values for the project.
    */
-  async createTwaProject(targetDirectory: string, twaManifest: TwaManifest): Promise<void> {
+  async createTwaProject(targetDirectory: string, twaManifest: TwaManifest,
+      reportProgress: twaGeneratorProgress = noOpProgress): Promise<void> {
+    let currentProgress = 0;
+    const totalProgress = 8;
```
PEConn, line 231: This worries me a bit. I can't see a way around it though unless we break each of these chunks of code into a "step" and then have a list of steps which we iterate through and perform. I'm not sure that's worth it though.

Could we at least add a test to check we finish with `currentProgress == totalProgress`?

packages/core/src/lib/TwaGenerator.ts
```diff
@@ -224,54 +225,67 @@ export class TwaGenerator {
    * @param {String} targetDirectory the directory where the project will be created
    * @param {Object} twaManifest configurations values for the project.
    */
-  async createTwaProject(targetDirectory: string, twaManifest: TwaManifest): Promise<void> {
+  async createTwaProject(targetDirectory: string, twaManifest: TwaManifest,
+      reportProgress: twaGeneratorProgress = noOpProgress): Promise<void> {
```
PEConn, line 229: So we use the default value here when calling `createTwaProject` from `init.ts` - why don't we want to show the progress bar in that case?

andreban: `init` is invoking the method from `shared.ts`, so it will show the progress bar. `TwaGenerator` is part of `core` and applications using it probably don't want the progress bar.

andreban:
Added a video.

packages/core/src/lib/TwaGenerator.ts
andreban: Added a wrapper class for progress (may be overkill?). It will ensure a) trying to update progress after reaching total throws an error and b) the last progress call should be `done()` and if current and total don't match, it throws an Error.

I haven't made the class public, but if we need something similar elsewhere, it may be worth moving it somewhere else and making it public.

andreban: PS: I didn't add tests, as we need a larger refactor of `TwaGenerator` to make it testable.

PEConn:
Is there any (easy) way to get rid of the "eta 0s" when the progress bar is completed? Otherwise (or if it's too much effort), LGTM!

andreban:
> Is there any (easy) way to get rid of the "eta 0s" when the progress bar is completed? Otherwise (or if it's too much effort), LGTM!

Removed the ETA altogether (making it conditional would be more trouble that it would be worth it).

PEConn:
LGTM!

Issue #119, opened 2020-03-06
Allow `build ` to optionally output an App Bundle


andreban:
The Play Store warns developers to use App Bundles instead of APKs. 

The APK generated by llama-pack is already small and App Bundles are unlikely to generate gains. APKs are also easier to debug locally.

But we could provide developers with the choice of outputting a an App Bundle, by wrapping the proper gradle commands and providing a flab in the `build` command.

```shell
.\gradlew.bat bundleRelease
jarsigner -verbose -sigalg SHA256withRSA -digestalg SHA-256 -keystore android.keystore .\app\build\outputs\bundle\release\app-release.aab android
```

More Info:

- https://developer.android.com/studio/build/building-cmdline#build_bundle 
- https://docs.oracle.com/javase/8/docs/technotes/tools/windows/jarsigner.html 
- https://stackoverflow.com/questions/50560045/sign-android-app-bundle-from-command-line

andreban:
#264 adds a `--generateAppBundle` flag to `build`

PR #241, opened 2020-07-16
GradleWrapper can call bundleRelease

andreban:
- Introduces a GradleWrapper method to call gradle bundleRelease
  to produce app bundles as output.

This moves forward work for #119


PEConn:
This looks good, how are we going to expose it to the user?

andreban:
We'll need to add a flat to `build` that outputs bundles. We also need to sign the bundle with `jarsigner` instead of `apksigner`. I'll create a PR on that bit soon.

PR #248, opened 2020-07-18
Adds a wrapper for jarsigner

andreban:
- `jarsigner` is used to sign App Bundles instead of apksigner.
This PR introduces a wrapper for `jarsigner` as part #119.


packages/core/src/lib/jdk/JarSigner.ts
```diff
@@ -0,0 +1,55 @@
...
+ *  limitations under the License.
+ */
+
+import {JdkHelper} from './JdkHelper';
+import {executeFile} from '../util';
+import {SigningKeyInfo} from '../../lib/TwaManifest';
+
+const JARSIGNER_CMD = 'jarsigner';
+const SIGALG = 'SHA256withRSA';
+const DIGESTALG = 'SHA-256';
```
PEConn, line 23: I know you've just copied the names from the flags `jarsigner` has, but do you mind making them more descriptive - `SIGNATURE_ALGORITHM` and `DIGEST_ALGORITHM`?

andreban: done!

PR #264, opened 2020-07-23
Adds App Bundle generation to the CLI

### LINKED ISSUE #119: Allow `build ` to optionally output an App Bundle

andreban:
- Refactors `build.ts` to have a `Build` class
- Moves APK generation to a method
- Adds method to generate App Bundles
- Adds a `build` flag to generate App Bundles
- Fixes stacktraces for the failed validation showing in the wrong place.

Closes #119

### DIFF
--- a/packages/cli/src/lib/cmds/build.ts
+++ b/packages/cli/src/lib/cmds/build.ts
@@ -130,11 +130,11 @@ class Build {
     const outputFile = './app-release-signed.apk';
     await this.androidSdkTools.apksigner(
         signingKey.path,
-        passwords.keystorePassword, // keystore password
-        signingKey.alias, // alias
-        passwords.keyPassword, // key password
+        passwords.keystorePassword,
+        signingKey.alias,
+        passwords.keyPassword,
         './app-release-unsigned-aligned.apk', // input file path
-        outputFile, // output file path
+        outputFile,
     );
     this.log.info(`Generated Android APK at "${outputFile}"`);
   }
@@ -154,7 +154,7 @@ class Build {
       await this.androidSdkTools.installBuildTools();
     }
 
-    let validationPromise;
+    let validationPromise = null;
     if (!this.args.skipPwaValidation) {
       validationPromise = this.runValidation();
     }
@@ -172,7 +172,7 @@ class Build {
 
     await this.generateAssetLinks(twaManifest, passwords);
 
-    if (validationPromise) {
+    if (validationPromise !== null) {
       const result = await validationPromise;
       if (result.isOk()) {
         const pwaValidationResult = result.unwrap();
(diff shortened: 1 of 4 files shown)


packages/cli/src/lib/cmds/build.ts
```diff
@@ -30,137 +30,170 @@ interface SigningKeyPasswords {
...
+  async buildApk(signingKey: SigningKeyInfo, passwords: SigningKeyPasswords): Promise<void> {
+    await this.gradleWrapper.assembleRelease();
+    await this.androidSdkTools.zipalign(
+        './app/build/outputs/apk/release/app-release-unsigned.apk', // input file
+        './app-release-unsigned-aligned.apk', // output file
+    );
+    const outputFile = './app-release-signed.apk';
+    await this.androidSdkTools.apksigner(
+        signingKey.path,
+        passwords.keystorePassword, // keystore password
```
PEConn, line 133: The these comments needed? Your variables are named well enough to be self explanatory.

packages/cli/src/lib/cmds/build.ts
```diff
@@ -30,137 +30,170 @@ interface SigningKeyPasswords {
...
-      printValidationResult(pwaValidationResult, log);
-      if (pwaValidationResult.status === 'FAIL') {
-        log.warn('PWA Quality Criteria check failed.');
+    if (this.args.generateAppBundle) {
+      await this.buildAppBundle(twaManifest.signingKey, passwords);
+    }
+
+    await this.generateAssetLinks(twaManifest, passwords);
+
+    if (validationPromise) {
```
PEConn, line 175: Could you make this line a bit clearer - are we checking that the validationPromise is not null, or are we checking that it is not false?

packages/cli/src/lib/cmds/build.ts
andreban: done.

packages/cli/src/lib/cmds/build.ts
andreban: done - Changed to declare `let validationPromise = null` earlier so we can check `validationPromise !== null` here.

PR #270, opened 2020-08-01
Extract ShortcutInfo module

NotWoods:
Pulls `ShortcutInfo` into its own module and delegates some logic there, simplifying TwaManifest.

### DIFF
--- a/packages/core/src/lib/ShortcutInfo.ts
+++ b/packages/core/src/lib/ShortcutInfo.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Google Inc. All Rights Reserved.
+ * Copyright 2020 Google Inc. All Rights Reserved.
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
@@ -20,7 +20,7 @@ import {findSuitableIcon} from './util';
 import {WebManifestShortcutJson} from './types/WebManifest';
 
 // As described on https://developer.chrome.com/apps/manifest/name#short_name
-export const SHORT_NAME_MAX_SIZE = 12;
+const SHORT_NAME_MAX_SIZE = 12;
 
 // The minimum size needed for the shortcut icon
 const MIN_SHORTCUT_ICON_SIZE = 96;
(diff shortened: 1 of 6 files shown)


packages/core/src/lib/ShortcutInfo.ts
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2019 Google Inc. All Rights Reserved.
```
andreban, line 2:
```suggestion
 * Copyright 2020 Google Inc. All Rights Reserved.
```

packages/core/src/lib/TwaManifest.ts
```diff
@@ -22,16 +22,11 @@ import {findSuitableIcon, generatePackageId, validateNotEmpty} from './util';
 import Color = require('color');
 import Log from './Log';
 import {WebManifestIcon, WebManifestJson} from './types/WebManifest';
+import {ShortcutInfo, SHORT_NAME_MAX_SIZE} from './ShortcutInfo';
```
andreban, line 25: `SHORT_NAME_MAX_SIZE` is also use used for the launcher name, which is unrelated to `ShortcutInfo`, so it's a bit strange to import it here.

I don't think there's an established relationship between the maximum size of the Shortcut short name and the app short name, so we might want to use different variables. Alternatively, we'd define the variable somewhere else and import in both places.

packages/core/src/spec/lib/ShortcutInfoSpec.ts
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright 2019 Google Inc. All Rights Reserved.
```
andreban, line 2:
```suggestion
 * Copyright 2020 Google Inc. All Rights Reserved.
```

packages/core/src/spec/lib/ShortcutInfoSpec.ts
```diff
@@ -0,0 +1,79 @@
...
+        'url': '/invalid',
+      };
+      const manifestUrl = new URL('https://pwa-directory.com/manifest.json');
+      expect(() => ShortcutInfo.fromShortcutJson(manifestUrl, shortcut))
+          .toThrowError('missing metadata');
+    });
+  });
+
+  describe('#constructor', () => {
+    it('Builds a TwaManifest correctly', () => {
```
andreban, line 71:
```suggestion
    it('Builds a ShortcutIcon correctly', () => {
```

PR #275, opened 2020-08-10
Mock

chenlevy24:
added Mock Prompt and Log classes


package.json
```diff
@@ -38,6 +38,10 @@
   },
   "dependencies": {
     "@bubblewrap/core": "^1.4.0",
+    "@types/cli-progress": "^3.7.0",
+    "@types/mime-types": "^2.1.0",
+    "cli-progress": "^3.8.2",
+    "jimp": "^0.14.0",
```
PEConn, line 44: Why are we adding jimp in this CL?

packages/cli/src/lib/Cli.ts
```diff
@@ -57,6 +59,10 @@ export class Cli {
     switch (command) {
       case 'help':
         return await help(parsedArgs);
+      case 'doctor':
+        return await doctor();
+      case 'help':
+        return await updateConfig(parsedArgs);
```
PEConn, line 65: Wait, what is included in this PR? There seem to be a lot of changes?

chenlevy24: It should be 'updateConfig' instead of 'help'.
I added this in my last PR (added doctor and updateConfig commands), but for some reason it isn't in the code so I added it again

package.json
chenlevy24: I'm not sure, I didn't touch in my last work

andreban: Can you double-check if those dependencies are necessary here? This is the root `package.json`. It seems to me those dependencies should be in `/packages/core/package.json` or `/packages/cli/package.json`.

packages/cli/src/lib/cmds/build.ts
```diff
@@ -191,7 +191,7 @@ class Build {
 }
 
 export async function build(config: Config, args: ParsedArgs,
-    log = new Log('build'), prompt: Prompt = new InquirerPrompt()): Promise<boolean> {
+    log = new ConsoleLog('build'), prompt: Prompt = new InquirerPrompt()): Promise<boolean> {
```
andreban, line 194: Same as above:
```suggestion
    log: Log = new ConsoleLog('build'), prompt: Prompt = new InquirerPrompt()): Promise<boolean> {
```

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,70 @@
...
+import {join} from 'path';
+import {existsSync, promises as fsPromises} from 'fs';
+import {loadOrCreateConfig} from '../config';
+
+async function jdkDoctor(log: ConsoleLog): Promise<boolean> {
+  const config = loadOrCreateConfig();
+  const jdkPath = (await config).jdkPath;
+  // Checks if the given path is a real path.
+  if (!existsSync(jdkPath)) {
+    log.error('The jdkPath doesn\'t exist, please run the following command to update it:\n' +
```
andreban, line 27: We're moving all strings to `strings.ts`. The two reasons are cleaner code and that it will be easier to add internationalisation later.

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,70 @@
...
+    'bubblewrap updateConfig --jdkPath path/here. after thet run bubblewrap doctor again');
```
andreban, line 28: Is `path/here` a placeholder for the path? It's usually better to use something between brackets when its a placeholder.
```suggestion
    'bubblewrap updateConfig --jdkPath <path-to-jdk>. After that, run bubblewrap doctor again.');
```

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,70 @@
...
+  const file = await fsPromises.readFile(join(jdkPath, 'release'), 'utf-8');
+  if (!file) {
+    log.error('The jdkPath isn\'t correct, please run the following command to update it:' +
+    '\nbubblewrap updateConfig --jdkPath path/here, such that the folder "here" contains' +
+    ' the file "release". after thet run bubblewrap doctor again');
+    return false;
+  };
+  // Checks if the jdk's version is 8 as needed.
+  if (file.indexOf('JAVA_VERSION="1.8') < 0) {
+    log.error('Wrong jdk version. You should download "OpenJDK 8(LTS)" at the link below: \n' +
```
andreban, line 40:
```suggestion
    log.error('Unsupported jdk version. Please, download "OpenJDK 8(LTS)" at the link below: \n' +
```

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,70 @@
...
+  if (!existsSync(androidSdkPath)) {
+    log.error('The androidSdkPath doesn\'t exist, please run the following command to update it:' +
+    '\nbubblewrap updateConfig --androidSdkPath path/here. then run bubblewrap doctor again');
+    return false;
+  }
+  // Checks if the path given is indeed
+  if (!existsSync(join(androidSdkPath, 'build-tools'))) {
+    log.error('The androidSdkPath isn\'t correct, please run the following command to update it:' +
+    '\nbubblewrap updateConfig --jdkPath path/here, such that the folder "here" contains the ' +
+    'folder "build-tools". then run bubblewrap doctor again');
```
andreban, line 60:
```suggestion
    'folder "build-tools". Then, run bubblewrap doctor again.');
```

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,70 @@
...
+  return true;
+}
+
+async function androidSdkDoctor(log: ConsoleLog): Promise<boolean> {
+  const config = loadOrCreateConfig();
+  const androidSdkPath = (await config).androidSdkPath;
+  // Checks if the given path is a real path.
+  if (!existsSync(androidSdkPath)) {
+    log.error('The androidSdkPath doesn\'t exist, please run the following command to update it:' +
+    '\nbubblewrap updateConfig --androidSdkPath path/here. then run bubblewrap doctor again');
```
andreban, line 53:
```suggestion
    '\nbubblewrap updateConfig --androidSdkPath path/here. Then, run bubblewrap doctor again.');
```

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,70 @@
...
+  if (!existsSync(join(androidSdkPath, 'build-tools'))) {
+    log.error('The androidSdkPath isn\'t correct, please run the following command to update it:' +
+    '\nbubblewrap updateConfig --jdkPath path/here, such that the folder "here" contains the ' +
+    'folder "build-tools". then run bubblewrap doctor again');
+    return false;
+  };
+  return true;
+}
+
+export async function doctor(log = new ConsoleLog('doctor')): Promise<boolean> {
```
andreban, line 66:
```suggestion
export async function doctor(log: Log = new ConsoleLog('doctor')): Promise<boolean> {
```

packages/cli/src/lib/cmds/help.ts
```diff
@@ -30,6 +30,9 @@ const HELP_MESSAGES = new Map<string, string>(
         'validate ............ validates if an URL matches the PWA Quality Criteria for Trusted' +
             ' Web Activity',
         'install ............. installs the output application to a connected device',
+        'doctor .../.......... checks that the jdk and the androidSdk are in place and at the' +
+          ' right version',
```
andreban, line 34:
```suggestion
          ' correct version',
```

packages/cli/src/lib/cmds/help.ts
```diff
@@ -84,10 +86,27 @@ const HELP_MESSAGES = new Map<string, string>(
...
+        'Usage:',
+        '',
+        '',
+        'bubblewrap doctor',
+      ].join('\n')],
+      ['updateConfig', [
+        'Usage:',
+        '',
+        '',
+        'bubblewrap updateConfig --<jdk or androidSdk> new/path/to/folder',
```
andreban, line 99: Can I invoke the command with both --jdk and --androidSdk? Maybe something like this would work better:
```suggestion
        'bubblewrap updateConfig [--jdk <path-to-jdk>] [--androidSdk <path-to-android-sdk>]',
```

Also, we probably want to add a note that at least one option is required.

packages/cli/src/lib/cmds/init.ts
```diff
@@ -36,10 +36,10 @@ async function confirmTwaConfig(twaManifest: TwaManifest, prompt: Prompt): Promi
   prompt.printMessage(messages.messageWebAppDetailsDesc);
 
   twaManifest.host = await prompt.promptInput(
-      messages.promptHostMessage, twaManifest.host, validateHost);
+      messages.promptHostMessage, twaManifest.host, await validateHost);
 
   twaManifest.startUrl = await prompt.promptInput(
-      messages.promptStartUrl, twaManifest.startUrl, createValidateString(1));
+      messages.promptStartUrl, twaManifest.startUrl, await createValidateString(1));
```
andreban, line 42: `createValidateString` is not `async`. `await` is not needed here.
```suggestion
      messages.promptStartUrl, twaManifest.startUrl, createValidateString(1));
```

packages/cli/src/lib/cmds/init.ts
```diff
@@ -36,10 +36,10 @@ async function confirmTwaConfig(twaManifest: TwaManifest, prompt: Prompt): Promi
   prompt.printMessage(messages.messageWebAppDetailsDesc);
 
   twaManifest.host = await prompt.promptInput(
-      messages.promptHostMessage, twaManifest.host, validateHost);
+      messages.promptHostMessage, twaManifest.host, await validateHost);
```
andreban, line 39: This `await` is not needed as the function is not being invoked, just passed as parameter. Please, check others throughout the code.

```suggestion
      messages.promptHostMessage, twaManifest.host, validateHost);
```

packages/cli/src/lib/cmds/install.ts
```diff
@@ -23,7 +23,7 @@ const DEFAULT_APK_FILE = './app-release-signed.apk';
 const PARAMETERS_TO_IGNORE = ['--verbose', '-r'];
 
 export async function install(
-    args: ParsedArgs, config: Config, log = new Log('install')): Promise<boolean> {
+    args: ParsedArgs, config: Config, log = new ConsoleLog('install')): Promise<boolean> {
```
andreban, line 26:
```suggestion
    args: ParsedArgs, config: Config, log: Log = new ConsoleLog('install')): Promise<boolean> {
```

packages/cli/src/lib/cmds/version.ts
```diff
@@ -16,9 +16,9 @@
 
 import * as fs from 'fs';
 import * as path from 'path';
-import {Log} from '@bubblewrap/core';
+import {ConsoleLog} from '@bubblewrap/core';
 
-export async function version(log = new Log('version')): Promise<boolean> {
+export async function version(log = new ConsoleLog('version')): Promise<boolean> {
```
andreban, line 21:
```suggestion
export async function version(log: Log = new ConsoleLog('version')): Promise<boolean> {
```

packages/cli/src/lib/config.ts
```diff
@@ -29,22 +29,34 @@ const LEGACY_CONFIG_FOLDER = join(homedir(), '.llama-pack');
...
+    return Result.error(Error('Invalid path. Run \'bubblewrap doctor for more information\''));
+  } else {
+    const file = await fsPromises.readFile(join(path, 'release'), 'utf-8');
+    if (!file || (file.indexOf('JAVA_VERSION="1.8') < 0)) {
+      return Result.error(Error('Invalid path. Run \'bubblewrap doctor for more information\''));
+    }
+  }
+  return Result.ok(path);
+}
+
```
andreban, line 50: nit: extra blank line.

packages/cli/src/lib/config.ts
```diff
@@ -29,22 +29,34 @@ const LEGACY_CONFIG_FOLDER = join(homedir(), '.llama-pack');
...
+async function createConfig(prompt: Prompt = new InquirerPrompt): Promise<Config> {
+  // const jdkPath = prompt.promptInput()
+  const jdkPath = prompt.promptInput('Path to the JDK:', null, validateJdkInput);
+  const androidSdkPath = prompt.promptInput('Path to the SDK:', null, validateSdkInput);
+  return new Config(await jdkPath, await androidSdkPath);
 }
 
-async function renameConfigIfNeeded(log = new Log('config')): Promise<void> {
+async function renameConfigIfNeeded(log = new ConsoleLog('config')): Promise<void> {
```
andreban, line 59:
```suggestion
async function renameConfigIfNeeded(log: Log = new ConsoleLog('config')): Promise<void> {
```

packages/cli/src/lib/config.ts
```diff
@@ -65,12 +77,15 @@ async function renameConfigIfNeeded(log = new Log('config')): Promise<void> {
 }
 
 
-export async function loadOrCreateConfig(path = DEFAULT_CONFIG_FILE_PATH): Promise<Config> {
+export async function loadOrCreateConfig(prompt: Prompt = new InquirerPrompt,
+    path = DEFAULT_CONFIG_FILE_PATH): Promise<Config> {
+  console.log(path);
```
andreban, line 82: I'm assuming those two `console.log` are leftovers from debugging?

packages/cli/src/lib/pwaValidationHelper.ts
```diff
@@ -11,7 +11,8 @@ function getColor(score: ScoreResult): string {
   }
 }
 
-export function printValidationResult(validationResult: PwaValidationResult, log: Log): void {
+export function printValidationResult(validationResult: PwaValidationResult, log: ConsoleLog):
```
andreban, line 14:
```suggestion
export function printValidationResult(validationResult: PwaValidationResult, log: Log):
```

packages/cli/src/lib/pwaValidationHelper.ts
```diff
@@ -11,7 +11,8 @@ function getColor(score: ScoreResult): string {
...
+ void {
```
andreban, line 15: There's not need to declare `void` return types. TypeScript will infer it.

packages/cli/src/spec/inputHelpersSpec.ts
```diff
@@ -18,169 +18,175 @@ import * as inputHelpers from '../lib/inputHelpers';
...
+      expect((await inputHelpers.validateImageUrl('')).isError()).toBeTrue();
+      expect((await inputHelpers.validateImageUrl('ftp://example.com')).isError()).toBeTrue();
     });
 
-    it('returns Error for non-image mime-types', () => {
-      expect(
-          inputHelpers.validateImageUrl('https://www.example.com/html.svg').isError()).toBeTrue();
+    it('returns Error for non-image mime-types', async () => {
+      expect((await
+      inputHelpers.validateImageUrl('https://www.example.com/html.svg')).isError()).toBeTrue();
```
andreban, line 134: nit: indentation

packages/cli/src/spec/inputHelpersSpec.ts
```diff
@@ -18,169 +18,175 @@ import * as inputHelpers from '../lib/inputHelpers';
...
     });
 
-    it('returns Error for SVG images', () => {
-      expect(
-          inputHelpers.validateImageUrl('https://www.example.com/test.svg').isError()).toBeTrue();
+    it('returns Error for SVG images', async () => {
+      expect((await
+      inputHelpers.validateImageUrl('https://www.example.com/test.svg')).isError()).toBeTrue();
```
andreban, line 139: nit: indentation

packages/core/src/lib/jdk/KeyTool.ts
```diff
@@ -44,7 +44,7 @@ export class KeyTool {
   private jdkHelper: JdkHelper;
   private log: Log;
 
-  constructor(jdkHelper: JdkHelper, log = new Log('keytool')) {
+  constructor(jdkHelper: JdkHelper, log = new ConsoleLog('keytool')) {
```
andreban, line 47:
```suggestion
  constructor(jdkHelper: JdkHelper, log: Log = new ConsoleLog('keytool')) {
```

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
```diff
@@ -40,7 +40,7 @@ export class AndroidSdkTools {
    * @param {jdkHelper} jdkHelper the JDK information to be used by the Android SDK
    */
   constructor(process: NodeJS.Process, config: Config, jdkHelper: JdkHelper,
-       readonly log = new Log('AndroidSdkTools')) {
+       readonly log = new ConsoleLog('AndroidSdkTools')) {
```
andreban, line 43:
```suggestion
       readonly log: Log = new ConsoleLog('AndroidSdkTools')) {
```

packages/core/src/lib/TwaGenerator.ts
```diff
@@ -145,9 +145,9 @@ class Progress {
  * Generates TWA Projects from a TWA Manifest
  */
 export class TwaGenerator {
-  private log: Log;
+  private log: ConsoleLog;
 
-  constructor(log = new Log('twa-generator')) {
+  constructor(log = new ConsoleLog('twa-generator')) {
```
andreban, line 150:
```suggestion
  constructor(log: Log = new ConsoleLog('twa-generator')) {
```

packages/core/src/lib/MockLog.ts
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+import {Log} from '..';
+
+/**
+ * An utility class to print nice Log messages.
+ */
+export class MockLog implements Log {
+    /**
+     * Creates a new Log instance
```
andreban, line 24:
```suggestion
     * Creates a new MockLog instance
```

andreban: The comments throughout this calls don't reflect what it does. Can you check?

packages/cli/src/spec/inputHelpersSpec.ts
```diff
@@ -18,169 +18,175 @@ import * as inputHelpers from '../lib/inputHelpers';
...
```
andreban, line 162: nit: indentation

packages/cli/src/spec/inputHelpersSpec.ts
```diff
@@ -18,169 +18,175 @@ import * as inputHelpers from '../lib/inputHelpers';
...
+    it('returns Error for invalid urls', async () => {
+      expect((await inputHelpers.validateImageUrl('ftp://example.com')).isError()).toBeTrue();
     });
 
-    it('returns Error for non-image mime-types', () => {
-      expect(
-          inputHelpers.validateImageUrl('https://www.example.com/html.svg').isError()).toBeTrue();
+    it('returns Error for non-image mime-types', async () => {
+      expect((await
+      inputHelpers.validateImageUrl('https://www.example.com/html.svg')).isError()).toBeTrue();
```
andreban, line 157: nit: indentation

packages/cli/src/spec/inputHelpersSpec.ts
andreban, line 139: nit: indentation

packages/cli/src/spec/inputHelpersSpec.ts
andreban, line 134: nit: indentation

packages/cli/src/spec/inputHelpersSpec.ts
```diff
@@ -18,169 +18,175 @@ import * as inputHelpers from '../lib/inputHelpers';
...
-      expect(inputHelpers.validatePackageId('com.pwa_directory.appspot.com').isOk()).toBeTrue();
-      expect(inputHelpers.validatePackageId('com.pwa1directory.appspot.com').isOk()).toBeTrue();
-    });
-
-    it('returns Error for invalid package ids', () => {
-      expect(inputHelpers.validatePackageId('com.pwa@directory.appspot.com').isError()).toBeTrue();
-      expect(inputHelpers.validatePackageId('com..example').isError()).toBeTrue();
-      expect(inputHelpers.validatePackageId('').isError()).toBeTrue();
-      expect(inputHelpers.validatePackageId('com.1char.twa').isError()).toBeTrue();
+    it('returns Ok for valid package ids', async () => {
```
andreban, line 65: nit: indentation on the lines bellow

packages/cli/src/lib/config.ts
andreban, line 82: Are those console.log messages for debugging?

packages/cli/src/lib/Cli.ts
andreban: Does this PR replace #220? Should we close the other one?

packages/core/src/lib/Log.ts
```diff
@@ -14,10 +14,45 @@
  *  limitations under the License.
  */
 
+/**
+ * An interface for loggers.
+ */
+export interface Log {
+
+  verbose: boolean;
```
PEConn, line 22: Why does this need to be on the interface? (also, nit: empty line above).

packages/core/src/lib/Log.ts
```diff
@@ -14,10 +14,45 @@
...
+
+  /**
+   * Prints a debug message to the Log. message is ignored if the Log is not set to verbose.
+   * @param message the message the be printed.
+   * @param args extra arguments for the console.
+   */
+  debug(message: string, ...args: string[]): void;
+
+  /**
+   * Prints an info message to the Log. message is ignored if the Log is not set to verbose.
```
PEConn, line 32: nit: capital letters at the start of sentences please (other comments as well)

packages/core/src/lib/Log.ts
```diff
@@ -14,10 +14,45 @@
...
+   * @param message the message the be printed.
+   * @param args extra arguments for the console.
+   */
+  info(message: string, ...args: string[]): void;
+
+  /**
+   * Prints an warning message to the Log. message is ignored if the Log is not set to verbose.
```
PEConn, line 39: Surely you only want "message is ignored if the Log is not set to verbose." on `debug` right? Not the other methods as well.

packages/core/src/lib/MockLog.ts
```diff
@@ -0,0 +1,85 @@
...
+     * @param tag the tag used when logging. Printed at the beggining of a log message.
+     * @param verbose if the Log is verbose. Debug messages are only printed on verbose logs.
+     */
+    public verbose: boolean;
+    private tag: string;
+    private receivedData: Array<string>;
```
PEConn, line 30: I'd probably just call this `contents` (although `receivedData` is good if you'd prefer to keep that).

packages/core/src/lib/MockLog.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    getReceivedData(): Array<string> {
+      return this.receivedData;
+    }
+
+    /**
+     * Prints a debug message to the Log. message is ignored if the Log is not set to verbose.
+     * @param message the message the be printed.
+     * @param args extra arguments for the console.
+     */
```
PEConn, line 46: @andreban do we need full docs for methods that override stuff for an interface?

andreban: I'm happy with just keeping those on the interface.

packages/cli/src/lib/Cli.ts
chenlevy24: It includes it, but #220 was after I finished writing Doctor and Update and this PR added the Mock Log and Prompt

packages/cli/src/lib/config.ts
chenlevy24: After the return?

packages/cli/src/spec/inputHelpersSpec.ts
chenlevy24: That's what I did at the beginning but lint didn't like it for some reason

packages/core/src/lib/Log.ts
chenlevy24: That's what we talked about didn't we?
We make Log an interface and create 2 classes that will implement it: ConsoleLog(which will be exactly as log was), and MockLog

chenlevy24: by line above do you mean to remove the empty line on 21 or add a new one before 20?

packages/core/src/lib/Log.ts
chenlevy24: I didn't right it, it is the same comment as Log had before.. should I change it?

packages/core/src/lib/MockLog.ts
chenlevy24: So should I remove all of them from ConsoleLog and MockLog?

packages/cli/src/spec/inputHelpersSpec.ts
andreban: Try 4 spaces instead of 2. Also, make sure the line is not hitting the 100 char limit.

packages/cli/src/lib/pwaValidationHelper.ts
chenlevy24: it throws an error when I remove it.

andreban: Yeah, I checked other places. It seems we have a rule around that.

packages/cli/src/lib/cmds/help.ts
```diff
@@ -30,6 +30,9 @@ const HELP_MESSAGES = new Map<string, string>(
         'validate ............ validates if an URL matches the PWA Quality Criteria for Trusted' +
             ' Web Activity',
         'install ............. installs the output application to a connected device',
+        'doctor .../.......... checks that the jdk and the androidSdk are in place and at the' +
```
PEConn, line 33: What's with the random `/`?

packages/cli/src/lib/cmds/help.ts
```diff
@@ -51,7 +54,6 @@ const HELP_MESSAGES = new Map<string, string>(
         '',
         'Options:',
         '--skipPwaValidation ....... skips validating the wrapped PWA against the Quality Criteria',
-        '--generateAppBundle ....... outputs an Android App Bundle additionally to the APK',
```
PEConn, line 54: Did you mean to remove this?

packages/cli/src/lib/cmds/help.ts
```diff
@@ -84,10 +86,29 @@ const HELP_MESSAGES = new Map<string, string>(
...
+        'Usage:',
+        '',
+        '',
+        'bubblewrap doctor',
+      ].join('\n')],
+      ['updateConfig', [
+        'Usage:',
+        '',
+        '',
+        // eslint-disable-next-line no-implicit-coercion
```
PEConn, line 99: What's this about?

packages/cli/src/spec/mock/MockPrompt.ts
```diff
@@ -0,0 +1,58 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {Prompt} from '../../lib/Prompt';
+
+/**
+ * A an interface that promps for different types of user input.
+ */
+
```
PEConn, line 22: nit: random newline.

packages/cli/src/spec/mock/MockPrompt.ts
```diff
@@ -0,0 +1,58 @@
...
+export class MockPrompt implements Prompt {
```
PEConn, line 23: What is this class meant to be doing? It's being used for testing right?

In that case, we probably don't want to actually log messages to the console on line 25 (otherwise our test output will be all noisy).

packages/cli/src/spec/mock/MockPrompt.ts
```diff
@@ -0,0 +1,58 @@
...
+  async printMessage(message: string): Promise<void> {
+    console.log(message);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  async promptInput<T>(message: string): Promise<any> {
```
PEConn, line 29: I think this may want to return `Promise<T>`, the same way that `Prompt` does.

packages/cli/src/spec/mock/MockPrompt.ts
```diff
@@ -0,0 +1,58 @@
...
+    const result = Promise.resolve({
+      name: 'question',
+      result: message,
```
PEConn, line 32: I'm not quite sure what you're returning here - why two strings?

packages/cli/src/lib/cmds/updateConfig.ts
```diff
@@ -0,0 +1,68 @@
...
+
+async function updateAndroidSdkPath(path: string, log: Log): Promise<boolean> {
+  if (!existsSync(path)) {
+    log.error('Please enter a valid path.');
+    return false;
+  }
+  const config = loadOrCreateConfig();
+  const jdkPath = (await config).jdkPath;
+  const newConfig = new Config(jdkPath, path);
+  fsPromises.unlink(CONFIG_FILE_PATH);
```
PEConn, line 34: What does this do?

packages/cli/src/lib/cmds/updateConfig.ts
```diff
@@ -0,0 +1,68 @@
...
+  const newConfig = new Config(path, androidSdkPath);
+  fsPromises.unlink(CONFIG_FILE_PATH);
+  newConfig.saveConfig(CONFIG_FILE_PATH);
+  return true;
+}
+
+async function invoke(args: ParsedArgs, log: Log): Promise<boolean> {
+  if (args.jdkPath) {
+    await updateJdkPath(args.JdkPath, log);
+  } else if (args.androidSdkPath) {
```
PEConn, line 55: In the help text, you mention you can use one of JDK path or Android SDK path, or both of them. With this logic if a user enters both of them, it will only use the first one.

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,67 @@
...
+  if (!existsSync(join(androidSdkPath, 'build-tools'))) {
+    log.error(messages.androidSdkPathIsNotCorrect);
+    return false;
+  };
+  return true;
+}
+
+export async function doctor(log: Log = new ConsoleLog('doctor')): Promise<boolean> {
+  const jdkResult = jdkDoctor(log);
+  const androidSdkResult = androidSdkDoctor(log);
```
PEConn, line 62: I think you need an `await` on these two lines.

packages/cli/src/spec/mock/MockPrompt.ts
chenlevy24: I put there any on purpose, you can't return T type without using the validation function, and I don't need to use the validation function for this method purpose

packages/cli/src/lib/cmds/updateConfig.ts
chenlevy24: It creates a new config file, deletes the old one, and then saves the new one in it's designated place

packages/cli/src/lib/cmds/updateConfig.ts
chenlevy24: you are right.. fixed it

packages/cli/src/lib/cmds/help.ts
chenlevy24: without that lint throws 'no-implicit-coercion' error and tells me to put the Number(<next-line>) for some reason

PEConn: That's weird - I wonder why it isn't needed for the other commands.

packages/cli/src/spec/mock/MockPrompt.ts
PEConn: Ah OK. So I do think it would be good for this function to have the same signature as the function it is overriding on `Prompt`. There must be some way to create a `T`.

packages/cli/src/lib/cmds/updateConfig.ts
PEConn: I don't think we need `unlink` - if you just save the new one it will overwrite the old one.

packages/cli/src/spec/mock/MockPrompt.ts
chenlevy24: I tried to think and read about it and didn't find anything

packages/cli/src/lib/cmds/updateConfig.ts
chenlevy24: this is the error: 
error  use `Number('(You can insert one or both of them)')` instead  no-implicit-coercion
might it be because of the whole line is in ()?

packages/cli/src/lib/cmds/help.ts
andreban: There's a dangling `+` sign at the of the line below. Can that be the cause of the error? Generally, we want to avoid eslint-disable as much as possible.

packages/cli/src/spec/mock/MockPrompt.ts
```diff
@@ -0,0 +1,58 @@
...
+    });
+    return (await result).result;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  async promptChoice<T>(message: string): Promise<any> {
+    const result = Promise.resolve({
+      name: 'question',
```
andreban: Declaring the function as `async` will automatically wrap the response in a Promise. This function can be simplified as the following:

```typescript
async promptChoice<T>(message: string): Promise<any> {
  return {
    name: 'question',
    result: message,
  };
}
```

packages/core/src/lib/MockLog.ts
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * An utility class for testing used for get the logging messages without printing to the user.
+ */
+export class MockLog implements Log {
+    /**
+     * Creates a new MockLog instance
+     * @param tag the tag used when logging. Printed at the beggining of a log message.
+     * @param verbose if the Log is verbose. Debug messages are only printed on verbose logs.
+     */
+    public verbose: boolean;
```
andreban, line 28: It doesn't seem like `verbose` and `tag` are used in the implementation. Should they exist here?

packages/core/src/lib/MockLog.ts
```diff
@@ -0,0 +1,80 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {Log} from '..';
+
+/**
+ * An utility class for testing used for get the logging messages without printing to the user.
+ */
+export class MockLog implements Log {
```
andreban, line 22: `MockLog` is used only for testing. Should it be in the `spec` folder instead?

packages/cli/src/lib/config.ts
```diff
@@ -29,22 +29,34 @@ const LEGACY_CONFIG_FOLDER = join(homedir(), '.llama-pack');
...
+    }
+  }
+  return Result.ok(path);
 }
 
-async function renameConfigIfNeeded(log = new Log('config')): Promise<void> {
+
+async function createConfig(prompt: Prompt = new InquirerPrompt): Promise<Config> {
+  // const jdkPath = prompt.promptInput()
+  const jdkPath = prompt.promptInput('Path to the JDK:', null, validateJdkInput);
```
andreban, line 54: We should add `await` when calling the function instead of when using the value.

packages/cli/src/spec/mock/MockPrompt.ts
andreban: I don't fully understand the implementation - MockPrompt is only being used right now in `createConfig`, right? That function always expects it to return a `string`. It seems the implementation could be simplified to the following:

```typescript
  async promptInput<T>(message: string): Promise<T> {
    return message as unknown as T;
  } 
```
Casting to `unknown` and then to `T` allows us to get rid of the eslint warning. An important thing is that, at this point, `MockLock` only works when prompting strings. If the expected object is different (eg: `Color` or `URL`) it won't work.

We should either make this more generic so it can be used anywhere or move the class to `ConfigSpec.ts`, where it is used. This PR is already quite large, so the last might be better for now.

packages/cli/src/lib/cmds/help.ts
chenlevy24: I'll check what can I do

packages/cli/src/spec/mock/MockPrompt.ts
chenlevy24: What does `as unknown` does?

packages/core/src/lib/MockLog.ts
chenlevy24: I thought about removing them, but when I read about promps I saw that every prompt class should have them.
I wasn't sure what the best practice is but what I thought here is: if every prompt needs to have it, the best practice is to put it in the interface so if we want to build another prompt class for some reason we will have a proper interface for it

chenlevy24: and if we will remove them, some functions that will use `prompt` could give it `tag` or `verbose`, and if we will not have it in our implementation we couldn't give it `MockPrompt` instead

Issue #38, opened 2019-11-27
Add integration testing


PEConn:
Create a test that uses the dependencies and generates APKs.

andreban:
Adding the comment from #70:

- Add a generated project to the test folder
- Create an end-to-end tests that generates a project to folder
- If files on both folders match, it passes the test

PR #278, opened 2020-08-12
Supports enabling AppsFlyer Analytics

andreban:
- Allows developers to easily enable AppsFlyer. See more at:
https://support.appsflyer.com/hc/en-us/articles/360002330178-Using-AppsFlyer-with-TWA-and-PWA#inapp-events-with-twa-getting-the-appsflyer-id


packages/core/src/lib/TwaManifest.ts
```diff
@@ -105,6 +105,7 @@ export class TwaManifest {
   generatorApp: string;
   webManifestUrl?: URL;
   fallbackType: FallbackType;
+  appsFlyer: boolean;
```
rajpalsaurabh: Can we rather keep it open ended to include any analytics provider instead of just providing an option of Appsflyer? We can have a parameter like "addExistingAnalyticsProvider" and then next parameter can be a set of supported providers which could just appsflyer to begin with.

rajpalsaurabh: Realized this has been taken care of in the future commits of including an array of plugins.

packages/core/template_project/app/src/main/AndroidManifest.xml
```diff
@@ -47,7 +47,7 @@
             android:name="twa_generator"
             android:value="@string/generatorApp" />
```
rajpalsaurabh, line 49: There has to be an additional application level receiver to be added for capturing app installs correctly with Appsflyer. 

<receiver android:name="com.appsflyer.SingleInstallBroadcastReceiver" android:exported="true">
            <intent-filter>
                <action android:name="com.android.vending.INSTALL_REFERRER" />
            </intent-filter>
        </receiver>

rajpalsaurabh: Need to add some additional permissions as well - ACCESS_NETWORK_STATE; ACCESS_WIFI_STATE; READ_PHONE_STATE

rajpalsaurabh: there has to be a new Application class and includes callback code for appsflyer. This new applciation class then has to be referenced something like android:name=".MyApp" inside the application tag within the manifest file.

packages/core/src/lib/plugins/Plugin.ts
```diff
@@ -0,0 +1,27 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+export type Plugin = {
+  name: string;
+  build?: {
```
PEConn, line 19: I think we can make (almost) everything here non-nullable. `repositories`, `dependencies` and `imports` can just default to empty lists, and that in turn means that `build` can also be non-nullable as well (just defaults to the empty  `repositories` and `dependencies`). Then we can remove half of the null checks from `PluginManager`.

We may not be able to do the same with `launchUrl` though.

packages/core/src/lib/plugins/AppsFlyerPlugin.ts
```diff
@@ -0,0 +1,34 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {Plugin} from './Plugin';
+
+export const appsFlyerPlugin: Plugin = {
```
PEConn, line 19: I see how the `Plugin` class works well for AppsFlyer - do you know whether it's likely to work more generally as well? (If we wanted to add a new plugin, would it fit nicely into the existing class?)

packages/core/src/lib/TwaManifest.ts
```diff
@@ -304,6 +306,8 @@ export interface TwaManifestJson {
   generatorApp?: string;
   webManifestUrl?: string;
   fallbackType?: FallbackType;
+  appsFlyer?: boolean;
+  plugins?: string[];
```
PEConn, line 310: Why both of these? Surely we just need one?

packages/core/template_project/app/src/main/java/LauncherActivity.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package <%= packageId %>;
+
+import android.net.Uri;
+
+<% for(const imp of pluginManager.launcherActivity.imports) { %>
+    import <%= imp %>;
+<% } %>
+
+public class LauncherActivity
```
PEConn, line 24: We're going to want some tests that check whether the generated project actually builds.

packages/core/template_project/app/src/main/AndroidManifest.xml
andreban: Added all of them. PTAL

packages/core/src/lib/plugins/Plugin.ts
andreban: Yeah, moved this to be an interface. Some plugins will require a config, and it's easier like that.

packages/core/src/lib/plugins/AppsFlyerPlugin.ts
andreban: Yeah, we'll need to expand this as we implement more features.

packages/core/template_project/app/src/main/java/LauncherActivity.java
andreban: Yeah - ideally we could add something like this to CI

packages/core/src/lib/TwaManifest.ts
andreban: Indeed. Removed `appsFlyer`.

packages/core/src/lib/plugins/AppsFlyerPlugin.ts
andreban: We'll need to add templates for the DelegationService too, and expand the Plugin (now renamed to Features) interface as we add new features.

packages/core/template_project/app/src/main/java/LauncherActivity.java
andreban: Added tests for `AppsFlyerFeature` and `FeatureManager`. I'll start working on #38 on follow-up PR - those moving parts will make more critical to ensure the Android project build.

packages/core/src/lib/features/Feature.ts
```diff
@@ -0,0 +1,108 @@
...
+     * by each plugin, and the Uri should be extended by calling `Uri.buildUpon`.
+     * Example:
+     * ```
+     * uri = uri
+     *     .buildUpon()
+     *     .appendQueryParameter("my_extra_parameter", "value")
+     *     .build();
+     * ```
+     */
+    launchUrl?: string;
```
PEConn, line 106: Where do we use this?

packages/core/template_project/app/src/main/java/LauncherActivity.java
```diff
@@ -0,0 +1,38 @@
...
+        extends com.google.androidbrowserhelper.trusted.LauncherActivity {
+
+    @Override
+    protected Uri getLaunchingUrl() {
+        // Get the original launch Url.
+        Uri uri = super.getLaunchingUrl();
+
+        <% for(const code of launcherActivity.launchUrl) { %>
```
PEConn, line 32: OK, so launchUrl isn't necessarily "override the launchUrl", it could be "modify the launchUrl".

packages/core/src/spec/lib/features/FeatureManagerSpec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+describe('FeatureManager', () => {
+  describe('#constructor', () => {
+    it('Creates from empty features', () => {
+      const manifest = {
+        features: {},
+        fallbackType: 'customtabs',
+      } as TwaManifest;
+      const features = new FeatureManager(manifest);
+      expect(features.androidManifest.components.length).toBe(0);
```
PEConn, line 29: Do you think these would read better as:

```
expect(features.androidManifest.components).toBe([]);
```

PEConn (approved):
Looking good, just a question and a style comment.

packages/core/template_project/app/src/main/java/LauncherActivity.java
andreban: That's right.

packages/core/src/spec/lib/features/FeatureManagerSpec.ts
andreban: Done. I think this really shines in this example:

```typescript
      expect(appsFlyerFeature.applicationClass.variables)
          .toEqual(['private static final String AF_DEV_KEY = "12345";']);
```

PR #279, opened 2020-08-13
Add site settings shortcut generation

padr31:
- Add a variable enableSiteSettingsShortcut into TwaManifest.ts that defaults to true
- Add ManageDataLauncherActivity into the template manifest
- Add enableSiteSettingsShortcut into tests

Behaviour tested by running npm run test and also by building an .apk and launching it in an android emulator.

### DIFF
--- a/packages/core/src/lib/TwaManifest.ts
+++ b/packages/core/src/lib/TwaManifest.ts
@@ -135,7 +135,7 @@ export class TwaManifest {
     this.webManifestUrl = data.webManifestUrl ? new URL(data.webManifestUrl) : undefined;
     this.fallbackType = data.fallbackType || 'customtabs';
     this.enableSiteSettingsShortcut = data.enableSiteSettingsShortcut != undefined ?
-    data.enableSiteSettingsShortcut : true;
+      data.enableSiteSettingsShortcut : true;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/core/src/lib/TwaManifest.ts
```diff
@@ -132,6 +134,8 @@ export class TwaManifest {
     this.generatorApp = data.generatorApp || DEFAULT_GENERATOR_APP_NAME;
     this.webManifestUrl = data.webManifestUrl ? new URL(data.webManifestUrl) : undefined;
     this.fallbackType = data.fallbackType || 'customtabs';
+    this.enableSiteSettingsShortcut = data.enableSiteSettingsShortcut != undefined ?
+    data.enableSiteSettingsShortcut : true;
```
andreban, line 138: nit: indentation

packages/core/src/lib/TwaManifest.ts
```diff
@@ -70,6 +70,7 @@ export type FallbackType = 'customtabs' | 'webview';
  * navigationColor: '<%= themeColor %>', // The color used for the navigation bar.
  * backgroundColor: '<%= backgroundColor %>', // The color used for the splash screen background.
  * enableNotifications: false, // Set to true to enable notification delegation.
+ * enableSiteSettingsShortcut: false, // Set to true to enable a dynamic shortcut into site settings.
```
PEConn, line 73: Maybe this should be `: true` instead of `: false` because it defaults to true? (I'm not quite sure what the structure of the comment here is though.)

andreban:
PS: This will depend on the next android-browser-helper version. So, let's hold merging until that's published. We should update the PR to make the template project use the correct version of that too.

PR #284, opened 2020-08-17
Console log

chenlevy24:
Changed `Log` into an interface and added `ConsoleLog`


packages/cli/src/index.ts
```diff
@@ -15,11 +15,11 @@
  */
 
 import {Cli} from './lib/Cli';
-import {Log} from '@bubblewrap/core';
+import {Log, ConsoleLog} from '@bubblewrap/core';
 
 module.exports = async (): Promise<void> => {
   const cli = new Cli();
-  const log = new Log('cli');
+  const log: Log = new ConsoleLog('cli');
```
PEConn, line 22: So while we do want to change the types of parameters to `Log`, we don't really care about explicitly changing the type of local variables to `Log`.

packages/core/package.json
```diff
@@ -47,5 +46,8 @@
     "node-fetch": "^2.6.0",
     "tar": "^6.0.2",
     "valid-url": "^1.0.9"
+  },
+  "devDependencies": {
+    "@types/mime-types": "^2.1.0"
```
PEConn, line 51: I think @andreban made a change that moved all the dev dependencies somewhere - https://github.com/GoogleChromeLabs/bubblewrap/commit/ad139d138146a3bb1a22b310db38efa1bfe0d288 .

I'm not sure if this goes against that, but lets see what he says.

packages/core/src/lib/Log.ts
```diff
@@ -14,10 +14,45 @@
...
+
+  /**
+   * Prints a debug message to the Log. message is ignored if the Log is not set to verbose.
+   * @param message the message the be printed.
+   * @param args extra arguments for the console.
+   */
+  debug(message: string, ...args: string[]): void;
+
+  /**
+   * Prints an info message to the Log. message is ignored if the Log is not set to verbose.
```
PEConn, line 32: These comments seem to be wrong (the bit about verbose only applies to `debug` I think).

Also, capital letters at the start of sentences please.

chenlevy24: I didn't write it, and I think it is right, if `verbose == false` it will not print anything

packages/core/package.json
andreban: Can you revert this change, please. `mime-types` is needed as a runtime dependency. This is also downgrading the minimum version from `2.1.27` to `2.1.0`. Generally, `devDependencies` should be empty in `core`, `cli` and `validator` and listed on the root `package.json`.

packages/core/src/lib/Log.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Google Inc. All Rights Reserved.
+ * Copyright 2020 Google Inc. All Rights Reserved.
```
andreban, line 2: Log has existed for a while. We don't need to change to 2020.

packages/core/package.json
```diff
@@ -32,20 +32,20 @@
...
     "@types/node-fetch": "^2.5.7",
     "@types/tar": "^4.0.3",
     "@types/valid-url": "^1.0.3",
+    "@types/sharp": "^0.25.0",
     "color": "^3.1.2",
     "extract-zip": "^1.7.0",
-    "inquirer": "^7.3.2",
-    "jimp": "^0.14.0",
+    "inquirer": "^7.0.4",
+    "sharp": "^0.23.2",
```
PEConn, line 45: Did you mean to change this line?

PR #285, opened 2020-08-18
Added MockLog class.

chenlevy24:


packages/core/src/spec/mock/MockLog.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Saves the debug message to the Logger.
+     * @param message The message the be saved.
+     */
+    error(message: string): void {
+      this.receivedData.push(message);
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
```
PEConn, line 68: So according to https://eslint.org/docs/rules/no-empty-function, we can also get around this lint rule by adding a comment inside the method, for example:

```
setVerbose(): void {
    // Not implemented for testing.
}
```

andreban: I didn't know this was a thing. Sounds like a better approach indeed.

chenlevy24: Cool

Issue #51, opened 2019-12-14
Dark (night) mode with TWA/PWA


juho-hanhimaki:
Is there a way to achieve light/dark (day/night) theming with TWA?

My TWA/PWA uses the prefers-color-scheme media query to adjust the CSS theming as requested by the OS/user. However the llama-pack and TWA samples I have seen accept only single color/colorPrimary and color/navigationColor which makes status bar and navigation bar ignore any OS light/dark theme.

andreban:
I don't see a reason why [`setColorSchemeParams`][1] and [`setColorScheme`][2] wouldn't work as expected on TWAs. We'd need to add support for it in [`android-browser-helper`][3] before adding this to `llama-pack`.

On the llama-pack side, I don't think there are fields dark-theme that can be extracted from the Web Manifest, but we can either generate the values from the light themes or ask users to provide colors for the dar theme (defaulting to using the same as the light theme).


[1]: https://developer.android.com/reference/androidx/browser/customtabs/CustomTabsIntent.Builder.html#setColorSchemeParams(int,%20androidx.browser.customtabs.CustomTabColorSchemeParams)

[2]: https://developer.android.com/reference/androidx/browser/customtabs/CustomTabsIntent.Builder.html#setColorScheme(int)

[3]: https://github.com/GoogleChrome/android-browser-helper

andreban:
Created an issue for android-browser-helper: https://github.com/GoogleChrome/android-browser-helper/issues/50

andreban:
Changed from `bug` to `enhancement`, as this is not currently supported in the Web Manifest specification.

Stevie-Ray:
Hi @andreban in [1.2.0](https://developer.android.com/jetpack/androidx/releases/browser#1.2.0)  `CustomTabColorSchemeParams` is supported how can i enable this for my TWA/PWA?

andreban:
Closed via #287

Benfordy:
Hi All,
In Samsung devices the Samsung WebView is used with an auto-dark mode that invert everything is a strange way, it is not using the regulars angular material light/dark theme in the same way that chrome does it. What can be done about that?

PR #287, opened 2020-08-18
Add divider color functionality

### LINKED ISSUE #66: Improve consistency between WebAPK (PWA) and TWA Splash Screens
The generated splash screen is not very faithful to the original PWA splash screen.

Also, I see that the splash use the normal icon. I think it would be nice to use the maskable icon.

| PWA splash | Llama Pack splash |
|-------------|--------------------|
| Screenshot_20200115-114932 | Screenshot_20200115-115016 |

padr31:
Add support for changing navbar divider color in TWAs generated via bubblewrap

Users required the ability to change the color of the navigation bar
divider custom tabs, which in some cases looked inconsistent. After adding
this feature to chromium and androidX, this PR adds it into bubblewrap
via a navigationColor variable that can be set in the web-manifest.json

Test: Tested the integration of bubblewrap and android-browser-helper 
by building an example app and changing the navbar color. 
Also modified tests to account for new navbar color variables.

Closes #66

### DIFF
--- a/packages/core/src/lib/TwaManifest.ts
+++ b/packages/core/src/lib/TwaManifest.ts
@@ -72,7 +72,7 @@ export type FallbackType = 'customtabs' | 'webview';
  * navigationColorDark: '<%= navigationColorDark %>', // The color used for the dark navbar.
  * navigationDividerColor: '<%= navigationDividerColor %>', // The color used for the
  * navbar divider.
- * navigationDividerColorDark: '<%= navigationDividerColorDark %>', // The color used for the dark 
+ * navigationDividerColorDark: '<%= navigationDividerColorDark %>', // The color used for the dark
  * navbar divider.
  * backgroundColor: '<%= backgroundColor %>', // The color used for the splash screen background.
  * enableNotifications: false, // Set to true to enable notification delegation.
--- a/packages/core/src/lib/TwaManifest.ts
+++ b/packages/core/src/lib/TwaManifest.ts
@@ -129,12 +129,12 @@ export class TwaManifest {
     this.display = asDisplayMode(data.display!) || DEFAULT_DISPLAY_MODE;
     this.themeColor = new Color(data.themeColor);
     this.navigationColor = new Color(data.navigationColor);
-    this.navigationColorDark = data.navigationColorDark == undefined ?
-      new Color(DEFAULT_NAVIGATION_COLOR) : new Color(data.navigationColorDark);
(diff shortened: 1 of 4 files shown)


andreban:
This seems to be solving #51 as well. Is that right?

padr31:
> This seems to be solving #51 as well. Is that right?

Yes, I just noticed that android-browser helper is expecting to get a dark-mode color from bubblewrap, but it was not getting it, so I added that.

packages/core/src/lib/TwaManifest.ts
```diff
@@ -120,6 +129,12 @@ export class TwaManifest {
     this.display = asDisplayMode(data.display!) || DEFAULT_DISPLAY_MODE;
     this.themeColor = new Color(data.themeColor);
     this.navigationColor = new Color(data.navigationColor);
+    this.navigationColorDark = new Color(data.navigationColorDark == undefined ?
+      DEFAULT_NAVIGATION_COLOR : data.navigationColorDark);
+    this.navigationDividerColor = new Color(data.navigationDividerColor == undefined ?
+      DEFAULT_NAVIGATION_DIVIDER_COLOR : data.navigationDividerColor);
+    this.navigationDividerColorDark = new Color(data.navigationDividerColorDark == undefined ?
+      DEFAULT_NAVIGATION_COLOR : data.navigationDividerColorDark);
```
PEConn, line 137: This can be `new Color(data.navigationDividerColorDark ?? DEFAULT_NAVIGATION_COLOR)`.

PEConn (approved):
Approved with one style comment.

JudahGabriel:
@andreban how do we control dark mode splash screen color? Inside [TwaManifest](https://github.com/GoogleChromeLabs/bubblewrap/blob/master/packages/core/src/lib/TwaManifest.ts), the only dark variables I see are:

- [navigationColorDark](https://github.com/GoogleChromeLabs/bubblewrap/blob/master/packages/core/src/lib/TwaManifest.ts#L77)
- [navigationDividerColorDark](https://github.com/GoogleChromeLabs/bubblewrap/blob/master/packages/core/src/lib/TwaManifest.ts#L80)

Those affect the nav bar...do they also affect the splash screen?

JudahGabriel:
Testing this on Android, those values only affect nav bar. Thus, unless I'm missing something, there isn't currently a way to change the splash screen color for dark mode. Let me know if I'm mistaken.

andreban:
🤦  Yes, that's my mistake, that's dark mode for the navigation bar. Thanks for noticing and checking.

PR #290, opened 2020-08-18
Added `npm run clean`

andreban:
- Running `npm run clean` cleans all `dist` folders.
- Renamed previous `clean` to `clean-modules`. That removes
  `node_modules`.

### DIFF
--- a/packages/validator/package-lock.json
+++ b/packages/validator/package-lock.json
@@ -4,11 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@types/jasmine": {
-      "version": "3.5.11",
-      "resolved": "https://registry.npmjs.org/@types/jasmine/-/jasmine-3.5.11.tgz",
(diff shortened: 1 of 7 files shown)


PEConn (approved):

packages/validator/package-lock.json
```diff
@@ -4,11 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@types/jasmine": {
-      "version": "3.5.11",
-      "resolved": "https://registry.npmjs.org/@types/jasmine/-/jasmine-3.5.11.tgz",
-      "integrity": "sha512-fg1rOd/DehQTIJTifGqGVY6q92lDgnLfs7C6t1ccSwQrMyoTGSoH6wWzhJDZb6ezhsdwAX4EIBLe8w5fXWmEng=="
-    },
```
chenlevy24, line 11: Isn't it a problem that jasmine's dependencies were removed?

andreban: It's not a problem - this is due to #283 where we moved all `devDependencies` to the root package, since they are shared. So, `package-lock.json` shouldn't contain `jasmine` anymore.

chenlevy24 (approved):

PR #292, opened 2020-08-19
Changed Prompt's ValidateFunctions to return a promise

chenlevy24:


PEConn (approved):
Oh, this was actually a lot more straightforward than I thought it would be!

andreban (approved):
LGTM too! Thank you!

PR #293, opened 2020-08-19
Added updateConfig function

chenlevy24:


packages/cli/src/lib/cmds/updateConfig.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+import {Config, Log, ConsoleLog} from '@bubblewrap/core';
+import {ParsedArgs} from 'minimist';
+import {join} from 'path';
+import {homedir} from 'os';
+import {existsSync} from 'fs';
+import {loadOrCreateConfig} from '../config';
+
+const CONFIG_FILE_PATH = join(homedir(), '.bubblewrap-config/bubblewrap-config.json');
```
andreban, line 24: We should probably use the constants from `config.ts` and avoid duplication.

packages/cli/src/lib/cmds/updateConfig.ts
```diff
@@ -0,0 +1,68 @@
...
+
+async function updateAndroidSdkPath(path: string, log: Log): Promise<boolean> {
+  if (!existsSync(path)) {
+    log.error('Please enter a valid path.');
+    return false;
+  }
+  const config = loadOrCreateConfig();
+  const jdkPath = (await config).jdkPath;
```
andreban, line 32: It's more idiomatic to move the `await` call to the line above where we are calling `loadOrCreateConfig`. So:
 
```ts
const config = await loadOrCreateConfig();
const jdkPath = config.jdkPath;
```

packages/cli/src/lib/cmds/updateConfig.ts
```diff
@@ -0,0 +1,68 @@
...
+  return true;
+}
+
+async function updateJdkPath(path: string, log: Log): Promise<boolean> {
+  if (!existsSync(path)) {
+    log.error('Please enter a valid path.');
+    return false;
+  }
+  const config = await loadOrCreateConfig();
+  const androidSdkPath = (await config).androidSdkPath;
```
andreban, line 44: We have already called `await` on the line above. This line can be just:
```suggestion
  const androidSdkPath = config.androidSdkPath;
```

packages/cli/src/lib/cmds/updateConfig.ts
```diff
@@ -0,0 +1,68 @@
...
+    log.error('usage: bubblewrap updateConfig [--androidSdk <path-to-jdk>]' +
+        '[--androidSdk <path-to-android-sdk>]\nYou can insert one or both of them.');
+    return false;
+  }
+  return true;
+}
+
+export async function updateConfig(args: ParsedArgs, log: Log = new ConsoleLog('updateConfig')):
+        Promise<boolean> {
+  return await invoke(args, log);
```
andreban, line 67: All `updateConfig` is doing right now is call `invoke`. Maybe the logic from `invoke` could be added directly to `updateConfig`? Or are we planning to make `updateConfig` more complex later?

andreban (changes requested):
A few small changes.

andreban (approved):
LGTM, Thank you.

PEConn (approved):

MrVibe:
please add 1 sample line on how to use it.  Getting this error "updateConfig ERROR Usage"

PR #295, opened 2020-08-20
Added doctor command

chenlevy24:


packages/cli/src/lib/strings.ts
```diff
@@ -75,6 +75,11 @@ type Messages = {
   promptNewAppVersionName: string;
   warnPwaFailedQuality: string;
   updateConfigUsage: string;
+  jdkPathIsNotValid: string;
+  jdkPathIsNotCorrect: string;
+  jdkIsNotsupported: string;
```
PEConn, line 80: nit: capital S on supported.

packages/cli/src/lib/strings.ts
```diff
@@ -248,4 +253,17 @@ the PWA:
   warnPwaFailedQuality: red('PWA Quality Criteria check failed.'),
   updateConfigUsage: 'Usage\n\n:[--jdk <path-to-jdk>] [--androidSdk <path-to-android-sdk>]' +
       '(You can insert one or both of them)',
+  jdkPathIsNotValid: 'The jdkPath doesn\'t exist, please run the following command to update it:' +
+      '\nbubblewrap updateConfig --jdkPath <path-to-jdk>. Then run bubblewrap doctor again.',
+  jdkPathIsNotCorrect: 'The jdkPath isn\'t correct, please run the following command to update ' +
```
PEConn, line 258: I'm not quite sure why we need these to be separate - they're both essentially the same thing right?

chenlevy24: Your'e right, it's redundant.. I will change it

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,57 @@
...
+import {ConsoleLog, Log} from '@bubblewrap/core';
+import {join} from 'path';
+import {existsSync, promises as fsPromises} from 'fs';
+import {loadOrCreateConfig} from '../config';
+import {enUS as messages} from '../strings';
+
+async function jdkDoctor(log: Log): Promise<boolean> {
+  const config = await loadOrCreateConfig();
+  const jdkPath = config.jdkPath;
+  const file = await fsPromises.readFile(join(jdkPath, 'release'), 'utf-8');
```
andreban, line 26: nit: the variable name is a bit deceiving - this not a file, but the content of the `release` file.

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,57 @@
...
+    return false;
+  };
+  return true;
+}
+
+export async function doctor(log: Log = new ConsoleLog('doctor')): Promise<boolean> {
+  const jdkResult = await jdkDoctor(log);
+  const androidSdkResult = await androidSdkDoctor(log);
+  if (jdkResult && androidSdkResult) {
+    log.info('Your jdkpath and androidSdkPath are valid.');
```
andreban, line 54: Would you mind moving this string to messages too?

andreban (changes requested):
Request a couple of small changes.

We'll likely want to use the implementation for `jdkDoctor` and `androidSdkDoctor` in other places - for example, when validating input for the JDK and AndroidSDK locations, so we'll likely need to move them to a better place soon. I'm happy to wait until a PR that needs this comes along though.

PR #296, opened 2020-08-24
New doctor

chenlevy24:


packages/cli/src/lib/strings.ts
```diff
@@ -248,4 +252,13 @@ the PWA:
   warnPwaFailedQuality: red('PWA Quality Criteria check failed.'),
   updateConfigUsage: 'Usage\n\n:[--jdk <path-to-jdk>] [--androidSdk <path-to-android-sdk>]' +
       '(You can insert one or both of them)',
+  jdkPathIsNotCorrect: 'The jdkPath isn\'t correct, please run the following command to update ' +
+      'it:\nbubblewrap updateConfig --jdkPath path/here, such that the folder "here" contains' +
```
PEConn, line 256: Could you change `path/here` to `<path-to-jdk>` so it's consistent with `androidSdkPathIsNotCorrect`?

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,62 @@
...
+  // Checks if the path given is valid.
+  if (!existsSync(join(androidSdkPath, 'build-tools')) || !existsSync(androidSdkPath)) {
+    log.error(messages.androidSdkPathIsNotCorrect);
+    return false;
+  };
+  return true;
+}
+
+export async function doctor(log: Log = new ConsoleLog('doctor')): Promise<boolean> {
+  const jdkResult = await jdkDoctor(log);
```
PEConn, line 56: Maybe it would be better loading the config in this method and passing it into both of the functions instead of loading the config twice (once in each function).

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,62 @@
...
+  const config = await loadOrCreateConfig();
+  const jdkPath = config.jdkPath;
+  // Checks if the path given is valid.
+  if (!existsSync(jdkPath)) {
+    log.error(messages.jdkPathIsNotCorrect);
+    return false;
+  };
+  try {
+    const file = await fsPromises.readFile(join(jdkPath, 'release'), 'utf-8');
+    if (file.indexOf('JAVA_VERSION="1.8') < 0) { // Checks if the jdk's version is 8 as needed.np
```
PEConn, line 33: nit: a random "np" at the end of the line.

PEConn (commented):
Looking good, thanks!

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -0,0 +1,61 @@
...
+    return false;
+  };
+  try {
+    const file = await fsPromises.readFile(join(jdkPath, 'release'), 'utf-8');
+    if (file.indexOf('JAVA_VERSION="1.8') < 0) { // Checks if the jdk's version is 8 as needed
+      log.error(messages.jdkIsNotSupported);
+      return false;
+    }
+  } catch {
+    log.error(messages.jdkPathIsNotCorrect);
```
andreban, line 37: nit: It might be helpful to append the error message to the error (`e.message`).

andreban (approved):
LGTM, one note that it might be helpful to print the error message when failing to read the file.

PEConn (approved):

PR #297, opened 2020-08-24
Added MockPrompt class

chenlevy24:


packages/cli/src/spec/mock/MockPrompt.ts
```diff
@@ -0,0 +1,59 @@
...
+  async promptInput<T>(message: string): Promise<T> {
+    return message as unknown as T;
+  }
+
+  /**
+   * Sets the output to be the given message.
+   * @param message the message to be returned.
+   */
+
+  async promptChoice<T>(message: string): Promise<T> {
```
PEConn, line 40: `promptInput`, `promptChoice` and `promptPasword` have the wrong signatures.

chenlevy24: It is fine, I don't get all of the inputs because I don't need them.
I can remove inputs from the end and it is fine by js..
It sends me no errors for implementing Prompt

andreban (changes requested):
My view is that `MockPrompt` is cannot be generally used in place of a `Prompt`, since its return type will always be a string - as an example, we can't use to test `init`. 

We can either make it more general, which may be an overkill for now, or declare it in the spec where it is used, so we know it's tightly coupled to that test.

packages/cli/src/spec/mock/MockPrompt.ts
andreban: See my general note. I'm fine with this, but this Prompt implementation seems tighly coupled with specific tests and should be declared with them. Otherwise, we'd need to make it a general class and Peter's comment applies.

PR #303, opened 2020-08-25
Bug fixes on Doctor

chenlevy24:


packages/core/src/lib/jdk/JdkHelper.ts
```diff
@@ -20,6 +20,21 @@ import {Config} from '../Config';
 import * as path from 'path';
 import {executeFile} from '../util';
 
+/**
+ * Returns the Home folder of the jdk.
+ * @param {Config} config The bubblewrap general configuration
+ * @param {NodeJS.Process} process Information from the OS process
+ */
+export function getJavaHome(config: Config, process: NodeJS.Process): string {
```
andreban, line 28: We can declare this as a `static` function in `JdkHelper`.

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -19,9 +19,10 @@ import {join} from 'path';
 import {existsSync, promises as fsPromises} from 'fs';
 import {loadOrCreateConfig} from '../config';
 import {enUS as messages} from '../strings';
+import {getJavaHome} from '../../../../core/src/lib/jdk/JdkHelper';
```
andreban, line 22: You are importing classes from `core` by relative path. You need to import using `@bubblewrap/core`. This is causing the build to break.

packages/cli/src/lib/cmds/doctor.ts
```diff
@@ -19,9 +19,10 @@ import {join} from 'path';
...
 
 async function jdkDoctor(config: Config, log: Log): Promise<boolean> {
-  const jdkPath = config.jdkPath;
+  const jdkPath = getJavaHome(config, process);
```
andreban, line 25: What about moving the entire path validation to a static function in `JdkHelper` and just using it in doctor?

packages/cli/src/lib/cmds/doctor.ts
chenlevy24: And let the androidSdk check stay in `doctor`? it seems weird a little bit that only one part of the two will be done in `doctor` don't it?

packages/cli/src/lib/cmds/doctor.ts
andreban: We could move the androidSdk check to AndroidSdkTools, and remove code duplication here: https://github.com/GoogleChromeLabs/bubblewrap/blob/master/packages/core/src/lib/androidSdk/AndroidSdkTools.ts#L44-L46

chenlevy24:
That's weird, in my code I pass both `npm run lint` and `npm run test`

PR #306, opened 2020-08-27
AndroidSdkTools constructor invoker  

chenlevy24:


packages/core/src/lib/androidSdk/AndroidSdkTools.ts
```diff
@@ -32,6 +33,19 @@ export class AndroidSdkTools {
   private jdkHelper: JdkHelper;
   private pathJoin: (...paths: string[]) => string;
 
+  static async newAndroidSdkTools(process: NodeJS.Process, config: Config, jdkHelper: JdkHelper,
+      log: Log = new ConsoleLog('AndroidSdkTools')): Promise<AndroidSdkTools> {
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
+    if ((await validatePath(config)).isError()) {
+      throw new Error(`androidSdkPath does not exist: ${config.androidSdkPath}`);
+    }
```
andreban, line 41: This entire block can be replaced with the following:

```suggestion
    (await AndroidSdkTools.validatePath(config)).unwrap();
```

Also, `eslint-disable-next-line` is a last resort. Let's avoid using it.

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
```diff
@@ -194,4 +205,19 @@ export class AndroidSdkTools {
     ];
     await util.execute(installCmd, env, this.log);
   }
+
+  /**
+   * Checks if the androidSdkPath in the config file is valid.
+   * @param {Config} config the bubblewrap general configuration.
+   */
+  static async validatePath(config: Config): Promise<Result<boolean, Error>> {
```
andreban, line 213: We can pass just the path instead of the entire config and make the function more generic - this allows using this for validating input too:

```suggestion
  static async validatePath(path: string): Promise<Result<boolean, Error>> {
```

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
```diff
@@ -20,6 +20,7 @@ import util = require('../util');
 import {Config} from '../Config';
 import {JdkHelper} from '../jdk/JdkHelper';
 import {Log, ConsoleLog} from '../../lib/Log';
+import {Result} from '../..';
```
andreban, line 23: Prefer explicit imports when importing via path.

```suggestion
import {Result} from '../../lib/Result';
```

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
```diff
@@ -194,4 +205,19 @@ export class AndroidSdkTools {
...
+  static async validatePath(config: Config): Promise<Result<boolean, Error>> {
+    const androidSdkPath = config.androidSdkPath;
+    // Checks if the path given is valid.
+    if (!fs.existsSync(path.join(androidSdkPath, 'tools'))|| !fs.existsSync(androidSdkPath)) {
+      return Result.error(new Error('androidSdkPathIsNotCorrect'));
+    };
+    return Result.ok(true);
+  }
 }
+export const newAndroidSdkTools = AndroidSdkTools.newAndroidSdkTools;
```
andreban, line 222: There's not need to export those. They can be accessed via `AndroidSdkTools.validatePath()` / `AndroidSdkTools. newAndroidSdkTools()`

packages/core/src/spec/lib/androidSdk/AndroidSdkToolsSpec.ts
```diff
@@ -16,7 +16,7 @@
 
 import {Config} from '../../../lib/Config';
 import {JdkHelper} from '../../../lib/jdk/JdkHelper';
-import {AndroidSdkTools} from '../../../lib/androidSdk/AndroidSdkTools';
+import {AndroidSdkTools, newAndroidSdkTools} from '../../../lib/androidSdk/AndroidSdkTools';
```
andreban, line 19: There are unused import. That will go away one we stop exporting newAndroidSdkTools, but as a best practice `lint` will tell you that and Github highlights it too.

andreban (changes requested):
This is looking good. Requested some small changes.

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
andreban: Also, let's rename the method to `create`, since `newAndroidSdkTools` becomes redundant after tha.

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
chenlevy24: If there is an error it will throw it by itself?
And about the eslint-disable-next-line, it tells me that I use the function before I define it.. the only way that I can define it before I'm using it is to remove it from the class and write it at the start of the file, but we wanted it to be static..

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
chenlevy24: I'm calling it sdkPath instead, we already import `path` so this name is taken

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
andreban: Check the [`Result.unwrap()`](https://github.com/GoogleChromeLabs/bubblewrap/blob/master/packages/core/src/lib/Result.ts#L74-L79) implementation. If you invoke `validatePath` with `AndroidSdkTools.validatePath()`, does it complain?

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
andreban: SGTM

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
andreban: @chenlevy24 this and other items are marked as resolved, but the code is unchanged. Are we possibly missing a commit?

packages/core/src/spec/lib/androidSdk/AndroidSdkToolsSpec.ts
```diff
@@ -92,13 +91,14 @@ describe('AndroidSdkTools', () => {
...
-      it(`Sets the correct ANDROID_HOME on ${test.platform}`, () => {
+      it(`Sets the correct ANDROID_HOME on ${test.platform}`, async () => {
         spyOn(fs, 'existsSync').and.returnValue(true);
         const config = buildMockConfig(test.platform);
         const process = buildMockProcess(test.platform);
         const jdkHelper = new JdkHelper(process, config);
         const mockLog = new MockLog();
-        const androidSdkTools = new AndroidSdkTools(process, config, jdkHelper, mockLog);
+        const androidSdkTools = (await AndroidSdkTools.create(process, config,
+            jdkHelper, mockLog));
```
andreban, line 101: This parenthesis is unnecessary. We can remove it and get a line with 99 charts.
```suggestion
        const androidSdkTools = await AndroidSdkTools.create(process, config, jdkHelper, mockLog);
```

packages/core/src/spec/lib/androidSdk/AndroidSdkToolsSpec.ts
```diff
@@ -125,7 +125,8 @@ describe('AndroidSdkTools', () => {
...
```
andreban, line 129: Same thing here
```suggestion
        const androidSdkTools = await AndroidSdkTools.create(process, config, jdkHelper, mockLog);
```

packages/core/src/spec/lib/androidSdk/AndroidSdkToolsSpec.ts
```diff
@@ -144,7 +145,8 @@ describe('AndroidSdkTools', () => {
       const process = buildMockProcess(tests[0].platform);
       const jdkHelper = new JdkHelper(process, config);
       const mockLog = new MockLog();
-      const androidSdkTools = new AndroidSdkTools(process, config, jdkHelper, mockLog);
+      const androidSdkTools = (await AndroidSdkTools.create(process, config,
+          jdkHelper, mockLog));
```
andreban, line 149:
```suggestion
      const androidSdkTools = await AndroidSdkTools.create(process, config, jdkHelper, mockLog);
```

packages/core/src/spec/lib/androidSdk/AndroidSdkToolsSpec.ts
```diff
@@ -184,23 +186,25 @@ describe('AndroidSdkTools', () => {
         const process = buildMockProcess(test.platform);
         const jdkHelper = new JdkHelper(process, config);
         const mockLog = new MockLog();
-        const androidSdkTools = new AndroidSdkTools(process, config, jdkHelper, mockLog);
+        const androidSdkTools = (await AndroidSdkTools.create(process, config,
+            jdkHelper, mockLog));
```
andreban, line 190:
```suggestion
        const androidSdkTools = await AndroidSdkTools.create(process, config, jdkHelper, mockLog);
```

packages/core/src/spec/lib/androidSdk/AndroidSdkToolsSpec.ts
```diff
@@ -184,23 +186,25 @@ describe('AndroidSdkTools', () => {
...
       const fsSpy = spyOn(fs, 'existsSync');
       fsSpy.and.returnValue(true);
 
       const config = buildMockConfig(tests[0].platform);
       const process = buildMockProcess(tests[0].platform);
       const jdkHelper = new JdkHelper(process, config);
       const mockLog = new MockLog();
-      const androidSdkTools = new AndroidSdkTools(process, config, jdkHelper, mockLog);
+      const androidSdkTools = (await AndroidSdkTools.create(process, config,
+          jdkHelper, mockLog));
```
andreban, line 207:
```suggestion
      const androidSdkTools = await AndroidSdkTools.create(process, config, jdkHelper, mockLog);
```

packages/core/src/spec/lib/androidSdk/AndroidSdkToolsSpec.ts
```diff
@@ -251,7 +255,8 @@ describe('AndroidSdkTools', () => {
         const process = buildMockProcess(test.platform);
         const jdkHelper = new JdkHelper(process, config);
         const mockLog = new MockLog();
-        const androidSdkTools = new AndroidSdkTools(process, config, jdkHelper, mockLog);
+        const androidSdkTools = (await AndroidSdkTools.create(process, config,
+            jdkHelper, mockLog));
```
andreban, line 259:
```suggestion
        const androidSdkTools = await AndroidSdkTools.create(process, config, jdkHelper, mockLog);
```

andreban (approved):
Looking good, thank you. Left a few suggestions around unnecessary parenthesis around an async call. Once that one is fixed, we can merge.

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
```diff
@@ -32,6 +33,16 @@ export class AndroidSdkTools {
   private jdkHelper: JdkHelper;
   private pathJoin: (...paths: string[]) => string;
 
+  static async create(process: NodeJS.Process, config: Config, jdkHelper: JdkHelper,
+      log: Log = new ConsoleLog('AndroidSdkTools')): Promise<AndroidSdkTools> {
+    (await AndroidSdkTools.validatePath(config.androidSdkPath)).unwrap();
```
PEConn, line 38: I'm not quite clear on the intent of this line since we seem to be throwing away the result.

Are we just checking that the call to `unwrap` doesn't throw?

andreban: `unwrap()` will cause the underlying error to be thrown, if the result is an `Error`. Otherwise, we ignore the response.

chenlevy24: Yes, we replaced a try and catch with that line..
validate path will return `Result<boolean>` if the path is valid, and  `Result<Error>` if it isn't

chenlevy24:
That's weird, in my code I pass both `npm run test` and `npm run lint`

PR #311, opened 2020-09-02
JdkHelper - Added a path validation function

chenlevy24:
Added a path validaton to jdkHelper and changed getJavaHome function to be static

### DIFF
--- a/packages/core/src/lib/jdk/JdkHelper.ts
+++ b/packages/core/src/lib/jdk/JdkHelper.ts
@@ -58,7 +58,7 @@ export class JdkHelper {
    */
   async runJava(args: string[]): Promise<{stdout: string; stderr: string}> {
     const java = this.process.platform === 'win32' ? '/bin/java.exe' : '/bin/java';
-    const runJavaCmd = this.joinPath(this.getJavaHome(this.config.jdkPath, this.process),
+    const runJavaCmd = this.joinPath(this.getJavaHome(),
         java);
     return await executeFile(runJavaCmd, args, this.getEnv());
   }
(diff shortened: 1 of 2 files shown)


andreban (changes requested):
This is generally good. I'd rather if we keep the API stable and avoid breaking changes.

PEConn (commented):
LGTM % Andre's comments

packages/core/src/lib/jdk/JdkHelper.ts
```diff
@@ -56,29 +58,58 @@ export class JdkHelper {
...
-   * @returns {string} the value for the JAVA_HOME
+   * @param {Config} config The bubblewrap general configuration
+   * @param {NodeJS.Process} process Information from the OS process
    */
-  getJavaHome(): string {
-    if (this.process.platform === 'darwin') {
-      return this.joinPath(this.config.jdkPath, '/Contents/Home/');
-    } else if (this.process.platform === 'linux' || this.process.platform === 'win32') {
-      return this.joinPath(this.config.jdkPath, '/');
+  getJavaHome(jdkPath: string, process: NodeJS.Process): string {
```
andreban, line 71: The old version of the method doesn't have any parameters and uses the jdkPath / process passed as parameter to the config. We want to keep the old signature.

packages/core/src/lib/jdk/JdkHelper.ts
```diff
@@ -56,21 +58,50 @@ export class JdkHelper {
...
-    const runJavaCmd = this.joinPath(this.getJavaHome(), java);
+    const runJavaCmd = this.joinPath(this.getJavaHome(),
+        java);
     return await executeFile(runJavaCmd, args, this.getEnv());
   }
 
   /**
    * Returns information from the JAVA_HOME, based on the config and platform.
-   * @returns {string} the value for the JAVA_HOME
+   * @param {Config} config The bubblewrap general configuration
```
andreban, line 68: Can we change back the comments here?

chenlevy24: Ah, I missed it

andreban (approved):
LGTM, Thank you!

PR #315, opened 2020-09-03
Add JdkInstaller

joycetoh8:
If JDK 8 is not already installed, JdkInstaller will download
the binary and source code at given path.

### DIFF
--- a/packages/core/src/lib/jdk/JdkInstaller.ts
+++ b/packages/core/src/lib/jdk/JdkInstaller.ts
@@ -16,7 +16,7 @@
 
 'use strict';
 
-import * as path from 'path'
+import * as path from 'path';
 import util = require('../util');
 
 const JDK_VERSION = '8u265-b01';
@@ -37,11 +37,12 @@ const JDK_SRC_ZIP = `jdk${JDK_VERSION}.zip`;
 export class JdkInstaller {
   private process: NodeJS.Process;
   private downloadFile: string;
-  private unzipFunction: (srcPath: string, dstPath: string, deleteWhenDone: boolean) => Promise<void>;
+  private unzipFunction: (srcPath: string, dstPath: string, deleteWhenDone: boolean)
+    => Promise<void>;
   private joinPath: (...paths: string[]) => string;
   /**
    * Constructs a new instance of JdkInstaller
-   * 
+   *
    * @param process {NodeJS.Process} process information from the OS process
    */
   constructor(process: NodeJS.Process) {
@@ -64,22 +65,22 @@ export class JdkInstaller {
         break;
       }
       default:
-          this.downloadFile = '';
-          console.log('Platform not found. Cannot download appropriate JDK.')
+        this.downloadFile = '';
+        console.log('Platform not found. Cannot download appropriate JDK.');
     }
   }
 
   /**
    * Downloads the platform-appropriate version of JDK 8, including
    * binary and source code.
-   * 
+   *
    * @param installPath {string} path to install JDK at
    */
   async install(installPath: string): Promise<string> {
     if (this.downloadFile === '') {
-      throw new Error(`Platform not found or unsupported: ${this.process.platform}. Cannot download appropriate JDK.`);
+      throw new Error(`Platform not found or unsupported: ${this.process.platform}.`);
     }
-    let dstPath = path.resolve(installPath);
+    const dstPath = path.resolve(installPath);
     const downloadSrcUrl = DOWNLOAD_JDK_SRC_ROOT + JDK_SRC_ZIP;
     const localSrcZipPath = this.joinPath(dstPath, JDK_SRC_ZIP);
     await util.downloadFile(downloadSrcUrl, localSrcZipPath);
@@ -93,4 +94,3 @@ export class JdkInstaller {
     return this.joinPath(dstPath, JDK_DIR);
   }
 }
- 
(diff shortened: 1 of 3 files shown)


packages/core/src/lib/jdk/JdkInstaller.ts
```diff
@@ -0,0 +1,96 @@
...
+ * download the binary and source code and decompress at
+ * path given by user.
+ */
+export class JdkInstaller {
+  private process: NodeJS.Process;
+  private downloadFile: string;
+  private unzipFunction: (srcPath: string, dstPath: string, deleteWhenDone: boolean)
+    => Promise<void>;
+  private joinPath: (...paths: string[]) => string;
+  /**
```
PEConn, line 43: nit: newline before the comment.

packages/core/src/lib/jdk/JdkInstaller.ts
```diff
@@ -0,0 +1,96 @@
...
+   * Constructs a new instance of JdkInstaller
```
PEConn, line 44: nit: full stops at the end of sentences (here, line 46, line 77).

packages/core/src/lib/jdk/JdkInstaller.ts
```diff
@@ -0,0 +1,96 @@
...
+        this.downloadFile = JDK_FILE_NAME_MAC;
+        break;
+      }
+      case 'linux': {
+        this.downloadFile = JDK_FILE_NAME_LINUX64;
+        break;
+      }
+      default:
+        this.downloadFile = '';
+        console.log('Platform not found. Cannot download appropriate JDK.');
```
PEConn, line 69: Would it not be better to throw the error at this point instead of in `install`? We know that the `JdkInstaller` is going to be useless here, I don't see the point in letting it exist for a while in an invalid state.

packages/core/src/lib/jdk/JdkInstaller.ts
```diff
@@ -0,0 +1,96 @@
...
+const DOWNLOAD_JDK_BIN_ROOT = `https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk${JDK_VERSION}/`;
+const DOWNLOAD_JDK_SRC_ROOT = 'https://github.com/AdoptOpenJDK/openjdk-jdk8u/archive/';
+const JDK_BIN_VERSION = JDK_VERSION.replace('-', '');
+const JDK_FILE_NAME_MAC = `OpenJDK8U-jdk_x64_mac_hotspot_${JDK_BIN_VERSION}.tar.gz`;
+const JDK_FILE_NAME_WIN32 = `OpenJDK8U-jdk_x86-32_windows_hotspot_${JDK_BIN_VERSION}.zip`;
+const JDK_FILE_NAME_LINUX64 = `OpenJDK8U-jdk_x64_linux_hotspot_${JDK_BIN_VERSION}.tar.gz`;
+const JDK_SRC_ZIP = `jdk${JDK_VERSION}.zip`;
+
+/**
+ * Checks whether JDK 8 is installed. If not installed,
```
PEConn, line 33: This comment is wrong isn't it? I can't see where this class checks whether JDK 8 is installed.

packages/core/src/lib/jdk/JdkInstaller.ts
```diff
@@ -0,0 +1,96 @@
...
+ * download the binary and source code and decompress at
+ * path given by user.
```
PEConn, line 35: Do you mind adding a comment saying why we need to download the JDK source code? Otherwise I could see some well intentioned person coming along and removing it since it's unused.

packages/core/src/lib/jdk/JdkInstaller.ts
```diff
@@ -0,0 +1,96 @@
...
+const DOWNLOAD_JDK_SRC_ROOT = 'https://github.com/AdoptOpenJDK/openjdk-jdk8u/archive/';
+const JDK_BIN_VERSION = JDK_VERSION.replace('-', '');
+const JDK_FILE_NAME_MAC = `OpenJDK8U-jdk_x64_mac_hotspot_${JDK_BIN_VERSION}.tar.gz`;
+const JDK_FILE_NAME_WIN32 = `OpenJDK8U-jdk_x86-32_windows_hotspot_${JDK_BIN_VERSION}.zip`;
+const JDK_FILE_NAME_LINUX64 = `OpenJDK8U-jdk_x64_linux_hotspot_${JDK_BIN_VERSION}.tar.gz`;
+const JDK_SRC_ZIP = `jdk${JDK_VERSION}.zip`;
+
+/**
+ * Checks whether JDK 8 is installed. If not installed,
+ * download the binary and source code and decompress at
```
PEConn, line 34: nit: your line wrapping is a bit weird here.

andreban:
I'm wondering if we should assume the "default" flow as installing the JDK to a default place. This would make the user flow easier. Something like:

> Do you want Bubblewrap to download / install a JDK now? (Answer No to re-use an existing JDK installation). (Y/n).

If the user answers yes (default), we can just install that into `$HOME/.bubblewrap/jdk` to make it simpler. If the user answers No, we ask them for the JDK Path.

packages/core/src/lib/jdk/JdkInstaller.ts
joycetoh8: Good point! I've added a line about it

joycetoh8:
@PEConn thanks for the review! I think I've addressed your comments now.

@andreban I also considered that but ultimately went with letting the user choose the location to have more choice. But I think you're right. Making it install in a default location is simpler for the user. So I've changed it to do this instead.

packages/core/src/lib/jdk/JdkInstaller.ts
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright 2019 Google Inc. All Rights Reserved.
```
andreban, line 2: nit: wrong year for the copyright header.

```suggestion
 * Copyright 2020 Google Inc. All Rights Reserved.
```

andreban (approved):
LGTM, with a nit on the date on the header of JdkInstaller.

Thank you!

PEConn (approved):

PR #316, opened 2020-09-03
Changed the validatepath return value of jdkHelper and androidSdkTools

chenlevy24:
Changed the validatepath return value of jdkHelper and androidSdkTools from Result<boolean, Error> to Result<string, Error>


andreban (approved):
LGTM, Thank you!

PEConn (approved):

PR #317, opened 2020-09-03
ValidatePathError

chenlevy24:
Added sub class of Error: validatePathError and changed the validatePath function of jdkHelper and androidSdkTools to return Promise<Result<string, validatePathError>>


andreban (changes requested):
LGTM, with one requestest. Let's move the error to `src/lib/errors/ValidatePathError.ts`. `types` is used to express external types (eg: a Web Manifest)

packages/core/src/lib/jdk/JdkHelper.ts
```diff
@@ -90,18 +91,21 @@ export class JdkHelper {
...
       const file = await fsPromises.readFile(path.join(javaHome, 'release'), 'utf-8');
       if (file.indexOf('JAVA_VERSION="1.8') < 0) { // Checks if the jdk's version is 8 as needed
-        return Result.error(new Error('jdkIsNotSupported'));
+        return Result.error(new ValidatePathError('The given jdk isn\'t supported.'
+            , 'PathIsNotSupported'));
       }
     } catch (e) {
-      return Result.error(new Error(e.message));
+      return Result.error(new ValidatePathError('Did not find the release file of the jdk.',
+          'PathIsNotCorrect'));
```
PEConn, line 108: Could this and PathIsNotSupported be constants in `ValidatePathError`? That way when we're checking the return value of this function, we can use those constants.

chenlevy24: Do you mean to add those constants in addition to the ErrorCode type which we talked about?

PEConn: Oh, they already are a type (I'd missed `ErrorCode` and thought you were just using these as strings).

Nevermind, code looks good to me.

PEConn (approved):

andreban (approved):
LGTM, Thank you!

PR #319, opened 2020-09-03
Include path instruction for spaces

### LINKED ISSUE #233: Include Android CLI path requirements in README
Android CLI location should not contain whitespace.

an error i made when following the tutorial. [#223](https://github.com/GoogleChromeLabs/bubblewrap/issues/223#issuecomment-656181524).

Please include a note in the Readme to this effect when setting llama-pack-config.json

amanintech:
Partially fixes #233

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -65,6 +65,7 @@ be changed by editing the configuration file at `${USER_HOME}/.bubblewrap/config
  }
 
 ```
+*(Note : Make sure you don't have `spaces` in the androidSdkPath.)*
 ## Quickstart Guide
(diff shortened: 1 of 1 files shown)


andreban (changes requested):
It's probably worth adding the reason why spaces are not supported. Maybe link to this StackOverflow question: https://stackoverflow.com/questions/37052934/android-sdk-location-should-not-contain-whitespace-as-this-cause-problems-with

amanintech:
I have added that link.

andreban (approved):
LGTM, Thank you!

PR #321, opened 2020-09-03
Enable a ChromeOS Only Build

### LINKED ISSUE #320: Enable a ChromeOS Only Build
If we want to specify that the build should be only for Chrome OS, we would need to add a flag.  This could ensure that app developers only target Chrome OS devices.

A flag of value specified in the TWAManifest.

Manually adding the required uses-feature value to specify Chrome OS only.

**Additional context**
n/a

nohe427:
This changes the TWAManifest to specify whether the build should be
designated as a ChromeOS Only package for distribution through Play.
Otherwise, the build will be available to all devices.

Closes #320

### DIFF
--- a/packages/core/src/lib/TwaManifest.ts
+++ b/packages/core/src/lib/TwaManifest.ts
@@ -93,6 +93,8 @@ type Features = {
  *  ],
  * // The duration of fade out animation in milliseconds to be played when removing splash screen.
  * splashScreenFadeOutDuration: 300
+ * isChromeOSOnly: false, // Setting to true will enable a feature that prevents non-ChromeOS devices
+ *  from installing the app.
  *
  */
 export class TwaManifest {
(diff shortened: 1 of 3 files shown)


nohe427:
I do have one implementation question.  At the current time, I have to have a user update the isChromeOSOnly flag from the twa-manifest.json file.  Instead, should I add a flag to init and update to specify isChromeOSOnly and update the manifest values accordingly?

Many thanks!

andreban:
We have a few options:

1. Keep as it is (init, modify twa-manifest.json, then run update)
2. Add a flag to init (eg: `--chromeOSOnly` or something)
3. Ask the question to developers during `init`.

I think 2 is fine, but if we expect a large number of developer using ChromeOS only, we should probably implement 3. @PEConn thoughts?

andreban (approved):
LGTM, Thank you!

PEConn:
Code looks good - I'm wondering if you have a specific app in mind for this though.

PR #323, opened 2020-09-03
Add a version code prompt

### LINKED ISSUE #322: Add a version code prompt
This feature request is related to the fact that if I have an existing Android application, I need a version code that has not been used before (and preferably) a higher integer than the current version code.

A prompt for me to select the version code I would like on init.

A flag instead of a prompt.

**Additional context**
none.

nohe427:
This adds a version code prompt that prompts the developer for what
version code they would like to use when starting development of their
app. This is important because if the developer has an existing Android
app, they would need to use a new version number to be able to update
the Play Store Console.

Closes #322


packages/cli/src/lib/strings.ts
```diff
@@ -249,6 +254,7 @@ the PWA:
   promptKeystorePassword: 'Password for the Key Store:',
   promptKeyPassword: 'Password for the Key:',
   promptNewAppVersionName: 'versionName for the new App version:',
+  promptVersionCode: 'Starting Version code for the new app version',
```
PEConn, line 257: nit: Version should probably not be capitalized.
Also, for  consistency with the other prompts, could we have a colon at the end?

packages/cli/src/lib/cmds/init.ts
```diff
@@ -76,6 +76,13 @@ async function confirmTwaConfig(twaManifest: TwaManifest, prompt: Prompt): Promi
       validateColor,
   );
 
+  twaManifest.appVersionCode = await prompt.promptInput(
```
PEConn, line 79: I think it would make more sense for this question to come after the packageId question (they're both about Android specific details).

PEConn (commented):
Thanks for the pull request! Overall it's looking pretty good, just a few nits from me and some formatting stuff from the bot.

packages/cli/src/lib/cmds/init.ts
nohe427: Package ID on line 60?  moving to right after that.

packages/cli/src/lib/strings.ts
nohe427: Done

andreban (approved):
LGTM, Thanks!

packages/cli/src/lib/cmds/init.ts
```diff
@@ -63,6 +63,14 @@ async function confirmTwaConfig(twaManifest: TwaManifest, prompt: Prompt): Promi
...
+      messages.promptVersionCode,
+      twaManifest.appVersionCode.toString(),
+      validateInteger,
+  );
+  twaManifest.appVersionName = twaManifest.appVersionCode.toString();
+
+
```
PEConn, line 73: nit: random extra space.

PEConn (approved):

PR #324, opened 2020-09-08
Added the `rmdir` method

andreban:
This function is used to delete a file or directory hierarchy and
will be used to clean an existing project before regenerating it.

More context in #326

### DIFF
--- a/packages/core/src/lib/util.ts
+++ b/packages/core/src/lib/util.ts
@@ -17,6 +17,7 @@
 import * as extractZip from 'extract-zip';
 import fetch from 'node-fetch';
 import * as fs from 'fs';
+import {join} from 'path';
 import {promisify} from 'util';
 import {exec, execFile, spawn} from 'child_process';
 import {x as extractTar} from 'tar';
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/util.ts
```diff
@@ -224,3 +225,23 @@ export function validatePackageId(input: string): string | null{
   }
   return null;
 }
+
+/**
+ * Removes a file or directory. If the path is a directory, recursively deletes files and
+ * directories inside it.
+ */
+export async function rmdirs(path: string): Promise<void> {
```
PEConn, line 233: Shouldn't this be called `rmdir` instead of `rmdirs`? I would assume `rmdirs` would take a list of paths.

packages/core/src/lib/util.ts
```diff
@@ -224,3 +225,23 @@ export function validatePackageId(input: string): string | null{
...
+  const stat = await fs.promises.stat(path);
+
+  // This is a regular file. Just delete it.
+  if (stat.isFile()) {
+    await fs.promises.unlink(path);
+    return;
+  }
+
+  // This is a directory. We delete files and sub directories inside it, then delete the
+  // directory itself.
```
PEConn, line 243: Why? What breaks if you just delete the directory?

andreban: Node 12 has a recursive option when deleting directories. Node 10 doesn't, and rmdir fails.

packages/core/src/lib/util.ts
andreban: Renamed.

packages/core/src/lib/util.ts
PEConn: :facepalm:

PR #325, opened 2020-09-08
MockPromptForString + doctor to work with ValidatePathError.

chenlevy24:
Added 'MockPromptForString' and changed 'doctor' to work with 'ValidatePathError'.


packages/cli/src/spec/mock/MockPromptForStrings.ts
```diff
@@ -0,0 +1,60 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {Prompt} from '../../lib/Prompt';
+
+/**
+ * A class which usef for testing and which mocks user's input.
```
PEConn, line 20: nit: typo on "usef".

PEConn (commented):
I'm not quite sure how `MockPromptForStrings` is meant to be used, could you give an example?

packages/cli/src/spec/mock/MockPromptForStrings.ts
```diff
@@ -0,0 +1,60 @@
...
+export class MockPromptForStrings implements Prompt {
+  async printMessage(): Promise<void> {
+    // An empty function for testing.
+  }
+
+  /**
+   * Sets the output to be the given message.
+   * @param message the message to be returned.
+   */
+  async promptInput<T>(message: string): Promise<T> {
```
andreban, line 31: I'm wondering if we can replace `T` across the class with `string`. Have we tried that?

chenlevy24: I'm pretty sure it will work but I think I will change the implementation to Peter's

packages/cli/src/spec/mock/MockPromptForStrings.ts
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+import {Prompt, ValidateFunction} from '../../lib/Prompt';
+
+/**
+ * A class which used for testing and which mocks user's input.
+ */
+export class MockPromptForStrings implements Prompt {
+  private responses: string[] = [];
+  private lastMessageIndex = -1;
```
PEConn, line 24: Can you not replace `lastMessageIndex` everywhere with `responses.length`?

packages/cli/src/spec/mock/MockPromptForStrings.ts
```diff
@@ -0,0 +1,104 @@
...
+
+
+  /**
+   * Sets the output to be the given message.
+   * @param message the message to be returned.
+   * @param {string[]} choices a list of choices. Not important for testing.
+   * @param {string | null} defaultValue a default value or null.
+   * @param {ValidateFunction<T>} validateFunction a function to validate the input.
+   */
+
```
PEConn, line 65: nit: in a couple of places, you've got random newlines between the docs and the method.

packages/cli/src/spec/mock/MockPromptForStrings.ts
```diff
@@ -0,0 +1,104 @@
...
+   */
+  async promptInput<T>(message: string,
+      _defaultValue: string | null,
+      validateFunction: ValidateFunction<T>): Promise<T> {
+    if (this.lastMessageIndex < 0) {
+      throw new Error('No answer was given. Please use addMessage(NextResponse) before' +
+      ' using this function');
+    }
+    const nextResponse = this.responses[this.lastMessageIndex--];
+    this.responses.pop();
```
PEConn, line 53: Could extract this to a private method something like `getNextResponse`, then just use that here, in `promptChoice` and in `promptPassword`?

packages/cli/src/spec/mock/MockPromptForStrings.ts
chenlevy24: I can, I had problems with push and pop so I did what I did, but this is a good solution

packages/cli/src/spec/mock/MockPromptForStrings.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * Sets the output to be the given message.
+   * @param {ValidateFunction<T>} validateFunction a function to validate the input.
+   * @returns {string} which is the next message to be prompted`.
+   */
+  private getNextMessage(): string {
+    if (this.responses.length < 0) {
+      throw new Error('No answer was given. Please use addMessage(NextResponse) before' +
```
PEConn, line 100: nit: NextResponse should be nextResponse (or even message, since that's what the parameter to addMessage is called).

packages/cli/src/spec/mock/MockPromptForStrings.ts
```diff
@@ -0,0 +1,107 @@
...
+      ' using this function');
+    }
+    const nextResponse = this.responses[this.responses.length];
```
PEConn, line 103: You're adding responses with `this.responses.push(...)` which adds them to the end of the queue, then you're getting the next response, which is the last one, then you're calling pop, which removes the last element of the queue. I think the order is going to get messed up, for example:

1. Call `addMessage("hello")`, `responses = ["hello"]`.
2. Call `addMessage("world")`, `responses = ["hello", "world"]`.
3. Call `getNextMessage()`, it returns `"world"`, `responses = ["hello"]`.
4. Call `getNextMessage()`, it returns `"hello"`, `responses = []`.

So you're return the responses in reverse order.

Here are the [methods you have on an array][1] to help you find what operations to perform.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array

packages/cli/src/spec/mock/MockPromptForStrings.ts
```diff
@@ -0,0 +1,107 @@
...
+   * @param {ValidateFunction<T>} validateFunction a function to validate the input.
+   * @returns {string} which is the next message to be prompted`.
+   */
+  private getNextMessage(): string {
+    if (this.responses.length < 0) {
+      throw new Error('No answer was given. Please use addMessage(nextMessage) before' +
+      ' using this function');
+    }
+    const nextResponse = this.responses[this.responses.length];
+    this.responses.shift();
```
PEConn, line 104: I still don't think the logic is correct here, you're adding to the front of the list, removing from the front of the list and reading from the back of the list.

```
addMessage("hello") // responses = ["hello"]
addMessage("world") // responses = ["world", "hello"]
getMessage() // returns "hello", responses = ["hello"]
getMessage() // returns "hello", responses = []
```

Could you write a test? Just a simple Spec that adds a few things and calls one of the prompts to make sure the logic and ordering is correct.

(Actually come to think of it, won't this always crash? If you've got an array of length `1`, you're going to try to access `responses[1]`, which is out of bounds.)

packages/cli/src/spec/mock/MockPromptForStrings.ts
```diff
@@ -0,0 +1,107 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {Prompt, ValidateFunction} from '../../lib/Prompt';
+
+/**
+ * A class which used for testing and which mocks user's input.
+ */
+export class MockPromptForStrings implements Prompt {
```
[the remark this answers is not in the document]
chenlevy24: But the messages are always strings

packages/cli/src/spec/mock/MockPromptForStrings.ts
chenlevy24: You are right, I will change the logic

chenlevy24: Should I upload the test or do it just for me?

PEConn: Upload the test, it would be nice to have a proper Spec for it.

packages/cli/src/spec/mock/MockPromptForStrings.ts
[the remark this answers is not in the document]
PEConn: I had thought that the "ForStrings" was about the return type of the methods - for example that promptInput would always return a string (as opposed to a Color or something). The version we've got now works if it's used in a place that returns a Color.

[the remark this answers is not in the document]
chenlevy24: The "forStrings" was for the input we accept, and as for now it's only strings

andreban (changes requested):

packages/cli/src/spec/mock/MockPromptForStrings.ts
```diff
@@ -0,0 +1,107 @@
...
+   * @param {ValidateFunction<T>} validateFunction a function to validate the input.
+   * @returns {string} which is the next message to be prompted`.
+   */
+  private getNextMessage(): string {
+    if (this.responses.length === 0) {
+      throw new Error('No answer was given. Please use addMessage(nextMessage) before' +
+      ' using this function');
+    }
+    const nextResponse = this.responses[0];
+    this.responses.shift();
```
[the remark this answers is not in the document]
chenlevy24: Yup, if I do as you say, I get an error because the value could be `undefined` or `string`.. Even if I check that it isn't empty before.. this is the only way I could make this work

packages/cli/src/spec/mockPromptSpec.ts
```diff
@@ -0,0 +1,36 @@
+/*
```
andreban, line 1: Can we rename this file to `MockPromptSpec.ts` (with capital M). Also, let's ensure this is consistent to the name for mockprompt (currently `MockPromptForStrings`).

packages/cli/src/spec/mockPromptSpec.ts
```diff
@@ -0,0 +1,36 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {Result} from '@bubblewrap/core';
+import {MockPromptForStrings} from './mock/MockPromptForStrings';
+
+async function validationFunction(message: string): Promise<Result<string, Error>> {
```
andreban, line 20: Should we test with a validationFunction that returns something other than a string?

packages/cli/src/spec/mockPromptSpec.ts
chenlevy24: So, should I call it `MockPromptForStringsSpec.ts`? (assuming we don't change the classes name)

packages/cli/src/spec/mockPromptSpec.ts
chenlevy24: I don't see how it could change the bottom line.. every case depends on the implementation but with those tests I made sure that the logic of the class is good and functioning (At least I think so)

packages/cli/src/spec/mock/MockPrompt.ts
```diff
@@ -0,0 +1,107 @@
...
+   * @returns {string} which is the next message to be prompted`.
+   */
+  private getNextMessage(): string {
+    if (this.responses.length === 0) {
+      throw new Error('No answer was given. Please use addMessage(nextMessage) before' +
+      ' using this function');
+    }
+    const nextResponse = this.responses[0];
+    this.responses.shift();
+    return nextResponse;
```
PEConn, line 105: `shift` returns the value it takes out of the array, so these three lines can just be `return this.responses.shift();`.

packages/cli/src/spec/MockPromptSpec.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+import {Result} from '@bubblewrap/core';
+import {MockPrompt} from './mock/MockPrompt';
+
+async function validationFunction(message: string): Promise<Result<string, Error>> {
+  return Result.ok(message);
+}
+
+describe('MockPromptForStrings', () => {
```
PEConn, line 24: MockPromptForStrings should now be MockPrompt

PEConn (approved):
Cool, this is looking good. Thanks for all of your work on it.

Approved (with two small comments).

packages/cli/src/spec/mock/MockPrompt.ts
chenlevy24: If I do as you say, I get an error because the value could be undefined or string.. Even if I check that it isn't empty before.. this is the only way I could make this work

PEConn (approved):

andreban (approved):
LGTM

PR #327, opened 2020-09-08
Cleans project before updating

### LINKED ISSUE #326: Calling `update` on an old project causes `build` to fail.
`styles.xml` has been removed from the current template but still exists in older projects. The styles in the `styles.xml` depend on `androidx.appcompat`, which was removed. This causes the build to fail, even though the style itself is not used anymore. We should do one of:

- Deleting `styles.xml` as part of the update.
- Adding an updated `styles.xml` to the current template (even if empty), to overwrite the old one.
- Ensure the entire existing project structure is deleted before regenerating it.

Since this is likely to happen with other files in future template updates, the 3rd option seems to be the most future proof.

andreban:
- Removes files generated by TwaGenerator before updating the
  project, to avoid old, outdated files braking the build.

Closes #326


chenlevy24:
LGTM! 👍

chenlevy24 (approved):
LGTM! 👍

PEConn (approved):

PR #328, opened 2020-09-08
Fixed reading shortcuts from twa manifest

andreban:
`update` was crashing with method not found: `assetName`. Loading
the JSON from the file doesn't automatically assign methods to
the shortcut. So, we create proper shortcuts when creating the
TwaManifest from the JSON.

### DIFF
--- a/packages/core/src/lib/TwaManifest.ts
+++ b/packages/core/src/lib/TwaManifest.ts
@@ -153,7 +153,10 @@ export class TwaManifest {
     this.signingKey = data.signingKey;
     this.appVersionName = data.appVersion;
     this.appVersionCode = data.appVersionCode || DEFAULT_APP_VERSION_CODE;
-    this.shortcuts = data.shortcuts;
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/TwaManifest.ts
```diff
@@ -153,7 +153,10 @@ export class TwaManifest {
     this.signingKey = data.signingKey;
     this.appVersionName = data.appVersion;
     this.appVersionCode = data.appVersionCode || DEFAULT_APP_VERSION_CODE;
-    this.shortcuts = data.shortcuts;
+    this.shortcuts = (data.shortcuts || []).map((si) => {
```
PEConn, line 156: Huh, this line is quite neat.

PEConn (approved):

packages/core/src/lib/TwaManifest.ts
andreban: Thanks.

PR #331, opened 2020-09-10
Bug fix on JdkHelper + tests.

chenlevy24:
Fixed a bug on 'JdkHelper's' 'validatePath' function and added tests to it.

### DIFF
--- a/packages/core/src/spec/lib/jdk/JdkHelperSpec.ts
+++ b/packages/core/src/spec/lib/jdk/JdkHelperSpec.ts
@@ -75,7 +75,7 @@ describe('JdkHelper', () => {
       mock.restore();
     });
 
-    it('Creates a Windows environment and checks that a valid path will not pass', async () => {
+    it('Creates a Windows environment and checks that an invalid path will not pass', async () => {
       mock({
         'jdk': {
           'release': {},
--- a/packages/core/src/spec/lib/jdk/JdkHelperSpec.ts
+++ b/packages/core/src/spec/lib/jdk/JdkHelperSpec.ts
@@ -18,6 +18,7 @@
 
 import {JdkHelper} from '../../../lib/jdk/JdkHelper';
 import {Config} from '../../../lib/Config';
+import * as mock from 'mock-fs';
 
 describe('JdkHelper', () => {
   describe('getEnv()', () => {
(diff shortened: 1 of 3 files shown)


chenlevy24:
I didn't test the static method of `getJavaHome` because it just calls the original one, there is nothing new to test

andreban:
> I didn't test the static method of `getJavaHome` because it just calls the original one, there is nothing new to test

Isn't it the other way around? the `getJavaHome()` that's **not** static calls the static one using instance variables from `JdkHelper`.

chenlevy24:
Right, I got confused but I know how it works.. anyway the bottom line is the same, I don't think that a new test for this function is needed

andreban:
`getJavaHome()` itself doesn't seem to have any tests. `getEnv()` does have tests, and currently tests it indirectly. I don't think `getJavaHome()` was being used elsewhere (it should probably have been marked as private). We shouldn't depend on the `getEnv()` test to check it, as the implementation there can change. I'm happy with adding a test for the static version and skipping the instance version, but we should test one of the two.

packages/core/src/spec/lib/jdk/JdkHelperSpec.ts
```diff
@@ -63,4 +64,25 @@ describe('JdkHelper', () => {
       expect(env['JAVA_HOME']).toBe('C:\\Users\\user\\jdk8\\');
     });
   });
+
+  describe('validatePath', () => {
+    it('Checks that given a valid path, validatePath will pass', async () => {
```
andreban, line 69: Can you add tests for the path validation each platform, similar to what the `getEnv()` test does?

andreban (changes requested):

PEConn (approved):
Code LGTM once Andre approves.

packages/core/src/spec/lib/jdk/JdkHelperSpec.ts
chenlevy24: I can, but if `getJavaHome()` it proven to be working those tests will mean nothing.. `validatePath` uses `getJavaHome` and after we got to this folder it all looks the same

andreban: Sounds good, adding tests to `getJavaHome()` should be enough.

packages/core/src/spec/lib/jdk/JdkHelperSpec.ts
```diff
@@ -63,4 +64,86 @@ describe('JdkHelper', () => {
...
+    it('Creates a Windows environment and checks that a valid path will pass', async () => {
+      mock({
+        'jdk': {
+          'release': 'JAVA_VERSION="1.8',
+        }});
+      expect((await JdkHelper.validatePath('jdk')).isOk()).toBeTrue();
+      mock.restore();
+    });
+
+    it('Creates a Windows environment and checks that a valid path will not pass', async () => {
```
PEConn, line 78: "checks that a valid path will not pass" ->
"checks that an invalid path will not pass"

PEConn (approved):

andreban (approved):
LGTM, Thank you!

PR #336, opened 2020-09-15
Config.ts and its tests to work with Prompt + Minor addition to MockPrompt.ts

chenlevy24:
Changed config.ts and it's tests to work with Prompt + Changed promptConfirm function to be more versatile.

### DIFF
--- a/packages/cli/src/spec/configSpec.ts
+++ b/packages/cli/src/spec/configSpec.ts
@@ -80,9 +80,9 @@ describe('config', () => {
       const mockLog = new MockLog();
       const mockPrompt = new MockPrompt();
       // Since 'createConfig' will be called, we push 3 future answers to 'mockPrompt'.
-      mockPrompt.addMessage('false');
-      mockPrompt.addMessage('jdk');
-      mockPrompt.addMessage('sdk');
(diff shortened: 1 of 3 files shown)


andreban (approved):
Looks great, thank you!

packages/cli/src/spec/configSpec.ts
```diff
@@ -80,7 +78,12 @@ describe('config', () => {
         [homedir()]: {},
       });
       const mockLog = new MockLog();
-      await loadOrCreateConfig(mockLog);
+      const mockPrompt = new MockPrompt();
+      // Since 'createConfig' will be called, we push 3 future answers to 'mockPrompt'.
+      mockPrompt.addMessage('false');
+      mockPrompt.addMessage('jdk');
+      mockPrompt.addMessage('sdk');
```
PEConn, line 85: Do you mind adding comments saying what these messages are meant to be answering? Just something like:

```
mockPrompt.addMessage('false'); // Should bubblewrap to install the JDK?
mockPrompt.addMessage('jdk'); // What is the path to the JDK?
```

PEConn (approved):
This is looking good, thanks!

PR #344, opened 2020-09-23
Improved error messages and fixes tests running on MacOS

andreban:
- Fixed type.
- Added the JDK path to the error message.
- Fixed tests failing on MacOS due to platform differences.

### DIFF
--- a/packages/cli/src/spec/DoctorSpec.ts
+++ b/packages/cli/src/spec/DoctorSpec.ts
@@ -24,6 +24,24 @@ import {MockPrompt} from './mock/MockPrompt';
 import {enUS as messages} from '../lib/strings';
 
 describe('doctor', () => {
+  let originalProcess: NodeJS.Process;
+
+  beforeEach(() => {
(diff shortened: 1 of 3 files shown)


packages/cli/src/spec/DoctorSpec.ts
```diff
@@ -24,6 +24,24 @@ import {MockPrompt} from './mock/MockPrompt';
 import {enUS as messages} from '../lib/strings';
 
 describe('doctor', () => {
+  let originalProcess: NodeJS.Process;
+
+  beforeEach(() => {
```
PEConn, line 29: So I guess this at least makes it explicit that we don't have any doctor tests for Mac?

PEConn (approved):

packages/cli/src/spec/DoctorSpec.ts
andreban: Yeah, `validatePath()` already handles other platforms, I think.

PR #350, opened 2020-09-30
Added bubblewrap merge command

chenlevy24:
Added a static merge command to TwaManifest.ts.
For the next step, I will add a CLI's merge command and implement it using this function


packages/core/src/lib/TwaManifest.ts
```diff
@@ -315,6 +315,139 @@ export class TwaManifest {
...
+        if (shortcuts.length === 4) {
+          break;
+        }
+      }
+    } else {
+      shortcuts = oldTwaManifest.shortcuts;
+    }
+
+    if (!('name' in fieldsToIgnore)) {
+      name = webManifest['name'] || webManifest['short_name'] || DEFAULT_APP_NAME;
```
PEConn, line 379: In all of these blocks, should it not be:

```
name = webManifest['name'] || webManifest['short_name'] || oldTwaManifest.name;
```

We want new values to override old ones, but I don't think we should ever have a default value override an old value.

packages/core/src/lib/TwaManifest.ts
```diff
@@ -315,6 +315,139 @@ export class TwaManifest {
...
+    const navigationColor = oldTwaManifest.navigationColor.hex();
+    const navigationColorDark = oldTwaManifest.navigationColorDark.hex();
+    const navigationDividerColor = oldTwaManifest.navigationDividerColor.hex();
+    const navigationDividerColorDark = oldTwaManifest.navigationDividerColorDark.hex();
+    const appVersionName = oldTwaManifest.appVersionName;
+    const signingKey = oldTwaManifest.signingKey;
+    const splashScreenFadeOutDuration = oldTwaManifest.splashScreenFadeOutDuration;
+    const enableNotifications = oldTwaManifest.enableNotifications;
+    const twaManifestShortcuts = shortcuts;
+    const twaManifestWebManifestUrl = oldTwaManifest.webManifestUrl!.toString();
```
PEConn, line 353: These are just the variables that we're passing straight into the new manifest's constructor?

I think having variables for these is just noise, just do:

```
    const twaManifest = new TwaManifest({
      packageId: oldTwaManifest.packageId,
```

packages/core/src/lib/TwaManifest.ts
```diff
@@ -315,6 +315,139 @@ export class TwaManifest {
...
+        webManifest['name']?.substring(0, SHORT_NAME_MAX_SIZE) || DEFAULT_APP_NAME;
+    } else {
+      launcherName = oldTwaManifest.launcherName;
+    }
+
+    if (!('display' in fieldsToIgnore)) {
+      display = asDisplayMode(webManifest['display']!) || DEFAULT_DISPLAY_MODE;
+    } else {
+      display = oldTwaManifest.display;
+    }
```
PEConn, line 395: You're using this pattern a lot, I think we can simplify it:

```
function maybeUpdate(fieldName: string, fieldsToIgnore: string[],
    oldValue: T, newValue: T): T{
  if (fieldName in fieldsToIgnore) {
    return oldValue;
  }
  return newValue || oldValue;
}
```

(This is also making that change about ignoring default values I mentioned above.)

Then all of this code just becomes:

```
...
const display = maybeUpdate('display', fieldsToIgnore, oldTwaManifest.display
  asDisplayMode(webManifest['display']!));
const backgroundColor = maybeUpdate('backgroundColor', fieldsToIgnore,
  oldTwaManifest.backgroundColor.hex(), webManifest['background_color']);
...
```

chenlevy24: Nice! will do

packages/core/src/lib/TwaManifest.ts
```diff
@@ -315,6 +315,101 @@ export class TwaManifest {
...
+    if (!('icons' in fieldsToIgnore)) {
+      icon = findSuitableIcon(webManifest.icons, 'any', MIN_ICON_SIZE);
+      maskableIcon = findSuitableIcon(webManifest.icons, 'maskable', MIN_ICON_SIZE);
+      monochromeIcon =
+        findSuitableIcon(webManifest.icons, 'monochrome', MIN_NOTIFICATION_ICON_SIZE);
+    }
+
+    const twaManifest = new TwaManifest({
+      packageId: oldTwaManifest.packageId,
+      host: oldTwaManifest.host,
```
PEConn, line 383: These lines don't have to be in any specific order do they?

Could we have all the lines that are just `: oldManifest.thing` grouped together, and all the lines that call `getNewFieldValue` grouped together? (It will make it easier to see which things we try to update and which we don't bother with.)

packages/core/src/lib/TwaManifest.ts
```diff
@@ -315,6 +315,101 @@ export class TwaManifest {
...
+          if (shortcutInfo != null) {
+            shortcuts.push(shortcutInfo);
+          }
+        } catch (err) {
+          TwaManifest.log.warn(`Skipping shortcut[${i}] for ${err.message}.`);
+        }
+        if (shortcuts.length === 4) {
+          break;
+        }
+      }
```
PEConn, line 366: I feel like this bit of logic probably exists somewhere else already. Could we make a `function getShortcutInfos(manifest: WebManifestJson): ShortcutInfo[]` and use that in both places?

packages/core/src/lib/TwaManifest.ts
chenlevy24: I thought they had to but you are right, done

andreban:
Generally, I'm wonder if we could simplify the implementation by doing the following:

- add a `toJson(): TwaManifestJson` method to `TwaManifest`. We can extract the implementation from `saveTofile()`:
```typescript
    const json: TwaManifestJson = Object.assign({}, this, {
      themeColor: this.themeColor.hex(),
      navigationColor: this.navigationColor.hex(),
      navigationColorDark: this.navigationColorDark.hex(),
      navigationDividerColor: this.navigationDividerColor.hex(),
      navigationDividerColorDark: this.navigationDividerColorDark.hex(),
      backgroundColor: this.backgroundColor.hex(),
      appVersion: this.appVersionName,
      webManifestUrl: this.webManifestUrl ? this.webManifestUrl.toString() : undefined,
    });
```

- Create a new TwaManifest: `const newTwaManifest = new TwaManifest(oldManifest.toJson());` or `const newTwaManifest = oldManifest.clone();`, which invokes the above.

- Finally, only assign the fields we do want to touch on newTwaManifest.

There's also a syntax we might be able to take advantage. Something like:
```typescript
const twaManifestJson = oldManifest.toJson();
const newTwaManifest = new TwaManifest({
  name: this.getNewFieldValue('name', fieldsToIgnore, oldTwaManifest.name,
          webManifest['name'] || webManifest['short_name']!),
  // Add other fields from webManifest here...
  ...twaManifestJson, // this will use the fields from twaManifestJson not set above.
});
```

chenlevy24:
I have a little problem with my implementation but didn't find any alternative.
I'm using twaManifest.saveToFile twice(first time in merge, and second time in update).
so the user will see the save message twice when they will use merge command unless we will find an alternative

packages/cli/src/lib/cmds/merge.ts
```diff
@@ -0,0 +1,41 @@
...
+  let fieldsToIgnore = [];
+  if (args.ignore) {
+    fieldsToIgnore = args.ignore;
+  }
+  const manifestFile = path.join(process.cwd(), 'twa-manifest.json');
+  const twaManifest = await TwaManifest.fromFile(manifestFile);
+  const webManifestUrl: URL = twaManifest.webManifestUrl!;
+  const webManifest = await util.getWebManifest(webManifestUrl);
+  await TwaManifest.merge(fieldsToIgnore, webManifestUrl, webManifest, twaManifest);
+  // Update the app (args are not relevant in this case, bucause update's default values
```
andreban, line 37: nit: typo / final stop.
```suggestion
  // Update the app (args are not relevant in this case, because update's default values.
```

packages/cli/src/lib/cmds/help.ts
```diff
@@ -101,6 +103,16 @@ const HELP_MESSAGES = new Map<string, string>(
...
       ].join('\n')],
+      ['merge', [
+        'Usage:',
+        '',
+        '',
+        'bubblewrap merge',
+        '',
+        '',
+        'Options: ',
+        '--ignore [fields-list]................. the fields which you would like to keep the same',
```
andreban, line 114: Where can I find the fields I can add to `fields-list`?

andreban:
Can we add documentation on the `merge` command to `packages/cli/README.md`?

packages/cli/README.md
```diff
@@ -192,6 +192,46 @@ Usage:
 bubblewrap help
 ```
 
+## `doctor`
+
+Validates that the jdk and the androidSdk are located at the path specified and that
```
PEConn, line 197: Do you mind saying "specified in your config"?

packages/cli/README.md
```diff
@@ -192,6 +192,46 @@ Usage:
...
+```
+
+## `updateConfig`
+
+Sets the paths of the jdk or the androidSdk to the given paths.
+
+Usage:
+
+```
+bubblewrap updateConfig  --jdk="/path-to-jdk" --androidSdk="/path-to-androidSdk"
```
andreban, line 213: This command is incorrect. See #352

packages/cli/src/lib/cmds/help.ts
```diff
@@ -34,6 +34,7 @@ const HELP_MESSAGES = new Map<string, string>(
         'updateConfig ........ sets the paths of the jdk or the androidSdk to the given paths',
         'doctor ............. checks that the jdk and the androidSdk are in place and at the' +
             ' correct version',
+        'merge ........ merges your web manifest into twaManifest.json',
```
PEConn, line 37: Could you add enough `.` so that the description is aligned with the cases above? (`doctor` is also misaligned)

packages/core/src/lib/TwaManifest.ts
```diff
@@ -16,6 +16,7 @@
 
 'use strict';
 
+import {join} from 'path';
```
andreban, line 19: Please, check the warnings from the CI.

packages/core/src/lib/util.ts
```diff
@@ -248,3 +248,18 @@ export async function rmdir(path: string): Promise<void> {
...
+   * Given a web manifest's URL, the function retrns the web manifest.
+   *
+   * @param {URL} webManifestUrl the URL where the webManifest is available.
+   * @returns {Promise<WebManifestJson}
+   */
+export async function getWebManifest(webManifestUrl: URL): Promise<WebManifestJson> {
+  const response = await fetch(webManifestUrl);
+  if (response.status !== 200) {
+    throw new Error(`Failed to download Web Manifest ${webManifestUrl}.` +
+        `Responded with status ${response.status}`);
```
PEConn, line 262: nit: there should be a space in the string  between the two sentences - at the moment it will say:

Failed to download Web Manifest https://www.example.com/manifest.json.Responded with status...

packages/cli/src/lib/cmds/merge.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Updates an existing TWA Project using the `twa-manifest.json`.
+ * @param {string} [args.fieldsToIgnore] the fields that shouldn't be updated.
+ */
+export async function merge(args: ParsedArgs): Promise<boolean> {
+  // If there is nothing to ignore, continue with an empty list.
+  let fieldsToIgnore = [];
+  if (args.ignore) {
+    fieldsToIgnore = args.ignore;
+  }
```
PEConn, line 31: Can this not be `const fieldsToIgnore = args.ignore || []`?

packages/cli/src/lib/cmds/merge.ts
```diff
@@ -0,0 +1,43 @@
...
+  const manifestFile = path.join(process.cwd(), 'twa-manifest.json');
+  const twaManifest = await TwaManifest.fromFile(manifestFile);
+  const webManifestUrl: URL = twaManifest.webManifestUrl!;
+  const webManifest = await util.getWebManifest(webManifestUrl);
+  const newTwaManifest =
+      await TwaManifest.merge(fieldsToIgnore, webManifestUrl, webManifest, twaManifest);
+  // Update the app (args are not relevant in this case, because update's default values
+  // are valid for it. We just send something as an input).
+  await newTwaManifest.saveToFile(path.join(process.cwd(), 'twa-manifest.json'));
+  await update(args);
```
PEConn, line 41: If `args` aren't relevant, could we just pass null here?

packages/cli/src/lib/cmds/merge.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Updates an existing TWA Project using the `twa-manifest.json`.
+ * @param {string} [args.fieldsToIgnore] the fields that shouldn't be updated.
+ */
+export async function merge(args: ParsedArgs): Promise<boolean> {
+  // If there is nothing to ignore, continue with an empty list.
+  let fieldsToIgnore = [];
+  if (args.ignore) {
+    fieldsToIgnore = args.ignore;
+  }
+  const manifestFile = path.join(process.cwd(), 'twa-manifest.json');
```
PEConn, line 32: Could this be called `manifestPath` instead?

packages/cli/src/lib/cmds/merge.ts
```diff
@@ -0,0 +1,43 @@
...
+  const twaManifest = await TwaManifest.fromFile(manifestFile);
+  const webManifestUrl: URL = twaManifest.webManifestUrl!;
+  const webManifest = await util.getWebManifest(webManifestUrl);
+  const newTwaManifest =
+      await TwaManifest.merge(fieldsToIgnore, webManifestUrl, webManifest, twaManifest);
+  // Update the app (args are not relevant in this case, because update's default values
+  // are valid for it. We just send something as an input).
+  await newTwaManifest.saveToFile(path.join(process.cwd(), 'twa-manifest.json'));
```
PEConn, line 40: You can reuse `manifestFile` (or `manifestPath` if you took my rename suggestion).

packages/core/src/lib/TwaManifest.ts
```diff
@@ -183,6 +183,16 @@ export class TwaManifest {
...
     });
+  }
+
+  /**
+   * Saves the TWA Manifest to the file-system.
+   *
+   * @param {String} filename the location where the TWA Manifest will be saved.
+   */
+  async saveToFile(filename: string): Promise<void> {
+    console.log('Saving Config to: ' + filename);
```
PEConn, line 194: Did you mean to leave this in? If so, should we use some of our logging infrastructure instead?

packages/core/src/lib/TwaManifest.ts
```diff
@@ -315,6 +308,102 @@ export class TwaManifest {
...
+   * @param {string[]} fieldsToIgnore the fields which needs to be ignored.
+   * @param {URL} webManifestUrl the URL where the webManifest is available.
+   * @param {WebManifest} webManifest the Web Manifest, used as a base for the update of
+   *    the TWA Manifest.
+   * @param {TwaManifest} oldTwaManifest current Twa Manifest.
+   * @returns {Promise<TwaManifest>} the new and merged Twa manifest.
+   */
+  static async merge(fieldsToIgnore: string[], webManifestUrl: URL,
+      webManifest: WebManifestJson, oldTwaManifest: TwaManifest): Promise<TwaManifest> {
+    function resolveIconUrl(icon: WebManifestIcon | null): string | undefined {
```
PEConn, line 367: You could make this method `updateIconUrl` and do:

```
function updateIconUrl(icon: WebManifestIcon | null, oldUrl: string) : string {
    return icon ? new URL(icon.src, webManifestUrl).toString() : oldUrl;
}
```
And then it would simplify lines 400-402.

PEConn: Actually, you could also have it include the call to `findSuitableIcon`.

packages/cli/src/lib/cmds/merge.ts
chenlevy24: I tried that, it isn't working. it says: "Argument of type 'null' is not assignable to parameter of type 'ParsedArgs'."
same with undefined

packages/core/src/lib/TwaManifest.ts
chenlevy24: I didn't write this method. It was there and I just used it

packages/core/src/lib/TwaManifest.ts
chenlevy24: I think that calling `findSuitableIcon` will not do much, I still need to pass all of the arguments so the code will be the same

packages/cli/src/lib/cmds/merge.ts
andreban: Looking at the update implementation, the bits that seem to matter from that code are:

```ts
const newVersionInfo = await updateVersions(twaManifest, args.appVersionName, prompt);
const twaGenerator = new TwaGenerator();
await twaGenerator.removeTwaProject(targetDirectory);
await generateTwaProject(prompt, twaGenerator, targetDirectory, twaManifest);
```
The other bits seem to mainly related to handling args parameters (which we don't care about)

chenlevy24: it sends it to `updateVersions` and then we get the following code: 
```// If a version was passed as parameter, use it.
  if (appVersionNameArg) {
    return {
      appVersionCode: appVersionCode,
      appVersionName: appVersionNameArg,
    };
  }

  // Otherwise, try to upgrade automatically with the versionCode.
  if (twaManifest.appVersionName === previousAppVersionCode.toString()) {
    return {
      appVersionCode: appVersionCode,
      appVersionName: appVersionCode.toString(),
    };
  }

  // If not not possible, ask the user to input a new version.
  const appVersionName = await prompt.promptInput(
      messages.promptNewAppVersionName,
      null,
      createValidateString(1),
  );
```

in all of the situations we have a fine solution, and when I run it it went perfectly

andreban: Looking at the other cmds, I don't think we have an instance of once command invoking another command directly. One reason for this is that it makes it easier to reason when changing to code in one command - By invoking `update` from `merge`,  we now need to be careful when changing `update`, like adding a new flag to the command, as it may affect `merge`.  But this relationship is not obvious and folks working on the project are likely to forget.

Examples of similar problems where we created shared code: 
 - the shared code between `init` and `update` has been extracted to`shared.ts`. 
 - the shared code between `validate` and `build` has been extracted to `pwaValidationHelper.ts`.

packages/core/src/lib/TwaManifest.ts
andreban: Would there be less code duplication if we called `findSuitableIcon`?

packages/core/src/lib/TwaManifest.ts
```diff
@@ -372,15 +371,18 @@ export class TwaManifest {
...
+    const oldTwaManifestJson = oldTwaManifest.toJson();
+    const iconUrl = (fieldsToIgnore.includes('icons'))? oldTwaManifestJson.iconUrl:
+       resolveIconUrl(findSuitableIcon(webManifest.icons, 'any', MIN_ICON_SIZE));
+    const maskableIconUrl = (fieldsToIgnore.includes('maskableIcons'))?
+        oldTwaManifestJson.maskableIconUrl :
+        resolveIconUrl(findSuitableIcon(webManifest.icons, 'maskable', MIN_ICON_SIZE));
+    const monochromeIconUrl = (fieldsToIgnore.includes('monochromeIcons'))?
+        oldTwaManifestJson.monochromeIconUrl :
+        resolveIconUrl(findSuitableIcon(webManifest.icons, 'monochrome',
+            MIN_NOTIFICATION_ICON_SIZE));
```
PEConn, line 383: I feel these three statements duplicate a fair amount of code and I think we could extract them to another method (similar to `getNewFieldValue`) (although it would probably take a horrible amount of parameters), something like:

```
function getNewIconUrl(fieldsToIgnore, fieldName, oldUrl, icons, iconType, iconSize);
```

(feel free to order the parameters as you feel appropriate)

PEConn (commented):
This is looking good - just one more comment for code cleanup.

packages/cli/src/lib/cmds/merge.ts
chenlevy24: So should I move this function to `shared.ts`?
I don't see the difference this will make, we still have one function which both of the commands use, and we need to be careful when doing changes in one of them

packages/core/src/lib/TwaManifest.ts
chenlevy24: I think that the latest change solved the problem, let me know what do you think

packages/cli/src/lib/cmds/help.ts
chenlevy24: Added explanation on which fields are valid (all of the fields in the web manifest)

packages/cli/src/lib/cmds/merge.ts
andreban: Yes, please. It doesn't make a functional difference, but the code becomes easier to maintain. As a developer who's not familiar with the project (or even ourselves modifying the code in the future), it is not immediately obvious that one command would invoke the other (eg: `merge` invoke `update`). But when you see a command invoking a function in another file (especially if this file is `shared.ts`), it's pretty reasonable to assume that the function is used in multiple places, and that any changes should have this in mind.

packages/cli/src/lib/cmds/help.ts
andreban: Here's a list of fields available in the Web Manifest: https://www.w3.org/TR/appmanifest/#webappmanifest-dictionary - It seems the ones we care about are:
 - `name`,
 - `short_name`,
 - `icons`,
 - `start_url`,
 - `display`,
 - `orientation`,
 - `theme_color`,
 - `background_color`,
 - and `shortcuts`.

The following fields are currently not used by Bubblewrap at all :
 - `dir`,
 - `lang`,
 - `description`,
 - `screenshots`,
 - `categories`,
 - ` iarc_rating_id`,
 - `scope`,
 - `related_applications`,
 -  and `prefer_related_applications` 

It's probably OK to support the whole list, but may be useful to point out which ones are used by Bubblewrap (on a following PR)

andreban (approved):
LGTM, Thanks!

PR #351, opened 2020-10-01
Add the Chrome OS only flag to init.

nohe427:
It was brought to my attention that most developers pursuing bubblewrap
would already know that they want to specify a ChromeOS only build. This
allows those developers to specify ChromeOS only from the beginning,
preventing them from needing to go back and modify the TWA-Manifest file
later on.


andreban (approved):
LGTM, Thank you!

PR #352, opened 2020-10-05
chore: Update updateConfig usage message

AntonioStipic:
Error Message on updateConfigUsage misdirected users


andreban (approved):
Looks good to me. Thanks for the contribution!

andreban:
@chenlevy24 can you take a quick look? I just want to make sure we don't want to change the code to match the docs instead.

chenlevy24:
Yes, that looks good. Thanks!

PR #356, opened 2020-10-07
Added the firstRunFlag feature

### LINKED ISSUE #301: Add a "first-run" query parameter optional feature
Add an option feature to Bubblewrap that allows it to append an extra parameter to `launchUrl` when the Trusted Web Activity is running for the first time. The value should be persisted and the parameter not attached / having a different value on the 2nd launch and onwards.

andreban:
-  When enabled, the `firstRunFlag` feature appends an extra
   query parameter to the the launch URL. The first time the app
   is launched, this parameter will contain the value `true`. On
   the following launches, the parameter will have the value
   `false`.
- This commit adds the feature, tests and modifies the template
  project
- Closes #301

### DIFF
--- a/packages/core/src/lib/features/FeatureManager.ts
+++ b/packages/core/src/lib/features/FeatureManager.ts
@@ -17,6 +17,7 @@
 import {Feature} from './Feature';
 import {AppsFlyerFeature} from './AppsFlyerFeature';
 import {TwaManifest} from '../TwaManifest';
+import {FirstRunFlagFeature} from './FirstRunFlagFeature';
 
 /**
  * Analyzes a TwaManifest to collect enable features and aggregates all customizations that will
@@ -38,6 +39,8 @@ export class FeatureManager {
   };
   launcherActivity = {
     imports: new Set<string>(),
+    methods: new Set<string>(),
+    variables: new Set<string>(),
     launchUrl: new Array<string>(),
   };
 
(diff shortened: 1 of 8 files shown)


packages/core/src/lib/features/FeatureManager.ts
```diff
@@ -49,6 +52,10 @@ export class FeatureManager {
       this.addFeature(new AppsFlyerFeature(twaManifest.features.appsFlyer));
     }
 
+    if (twaManifest.features.firstRunFlag !== undefined) {
```
PEConn, line 55: I know this isn't the first PR with that pattern, but I feel a little weird that we're using an empty object to signify we should turn the feature on.

PEConn (approved):
Looks good, with one vague thought rather than a comment that needs change.

Also, I think it would be good if we had a test that took a config and compared the generated LauncherActivity.java against some golden.

packages/core/src/lib/features/FeatureManager.ts
andreban: We could add a `enabled: boolean` to the config. The test would then become:

```ts
if (twaManifest.features.firstRunFlag && twaManifest.features.firstRunFlag.enabled)  {
```

PEConn: Yeah, that works well for me.

PR #357, opened 2020-10-08
Added tests to `merge` function on `core`.

chenlevy24:
Added 2 tests to `twaManifestSpec.ts` to tests the new function `merge`.

### DIFF
--- a/packages/core/src/spec/lib/TwaManifestSpec.ts
+++ b/packages/core/src/spec/lib/TwaManifestSpec.ts
@@ -323,4 +323,115 @@ describe('TwaManifest', () => {
       expect(asDisplayMode('')).toBeNull();
     });
   });
+  describe('#merge', () => {
+    it('Validates that the merge is done correctly in case which' +
+        ' there are no files to ignore', async () => {
+      const webManifest: WebManifestJson = {
+        'display': 'fullscreen',
+        'name': 'name',
+        'short_name': 'different_name',
+        'start_url': 'https://name.github.io/',
+        'icons': [{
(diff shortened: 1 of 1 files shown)


andreban (approved):
LGTM, Thanks!

packages/core/src/spec/lib/TwaManifestSpec.ts
```diff
@@ -323,4 +323,115 @@ describe('TwaManifest', () => {
       expect(asDisplayMode('')).toBeNull();
     });
   });
+  describe('#merge', () => {
+    it('Validates that the merge is done correctly in case which' +
+        ' there are no files to ignore', async () => {
```
PEConn, line 328: Should this be "files" or "fields"?

packages/core/src/spec/lib/TwaManifestSpec.ts
```diff
@@ -323,4 +323,115 @@ describe('TwaManifest', () => {
...
+        'launcherName': 'name',
+        'display': 'standalone',
+        'themeColor': '#FFFFFF',
+        'navigationColor': '#000000',
+        'navigationColorDark': '#000000',
+        'navigationDividerColor': '#000000',
+        'navigationDividerColorDark': '#000000',
+        'backgroundColor': '#FFFFFF',
+        'enableNotifications': false,
+        'startUrl': '/',
```
PEConn, line 354: Can you add a comment somewhere that the start_urls are different, but since they both resolve the same relative to the host url, nothing changes.

PEConn (approved):

PR #358, opened 2020-10-08
Introduced AbstractFeature

andreban:
- AbstractFeature allows removing boilerplate when implementing a Feature and only adding code for the fields we actually want to add details to.
    
- Updated AppsFlyerFeature and FirsRunFlagFeature to use the AbstractFeature.

### DIFF
--- a/packages/core/src/lib/features/AbstractFeature.ts
+++ b/packages/core/src/lib/features/AbstractFeature.ts
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2020 Google Inc. All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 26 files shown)


packages/core/src/lib/features/AbstractFeature.ts
```diff
@@ -0,0 +1,60 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {Feature} from './Feature';
+
+export class AbstractFeature implements Feature {
```
PEConn, line 19: I'm not quite sure I'd call this AbstractFeature, to me "abstract" kinda means "somewhat implemented, but left some gaps for subclasses to implement". Following examples from the Chrome codebase, I'd probably call this "EmptyFeature". It reads quite nicely too, "AppsFlyerFeature extends EmptyFeature but with these bits filled in".

PEConn (commented):
Code looks good, just a comment about the naming (also you have a typo in your first comment).

packages/core/src/lib/features/AbstractFeature.ts
andreban: SGTM - naming is hard :P

andreban: Renamed.

PEConn (approved):

PR #359, opened 2020-10-09
Allows enabling/disable a feature in the twa-manifest.

andreban:
- Features were automatically enabled when added to the twa-manifest.
- This allows finer control and disabling features without removing
  them from the Manifest.


PEConn (approved):
Thanks for this!

PR #364, opened 2020-10-13
Added alphaDependencies

chenlevy24:
Added alphaDependencies to `twaManifest.json` and added a check of that field in `app/build.gradle`. If it turned on, we will use version 2.0 of androidBrowserHelper, else we will use the current version.

### DIFF
--- a/packages/core/src/spec/lib/features/FeatureManagerSpec.ts
+++ b/packages/core/src/spec/lib/features/FeatureManagerSpec.ts
@@ -72,11 +72,22 @@ describe('FeatureManager', () => {
       expect(features.applicationClass.imports).toEqual(emptySet);
       expect(features.applicationClass.onCreate).toEqual([]);
       expect(features.applicationClass.variables).toEqual([]);
-      expect(features.buildGradle.dependencies).
-          toEqual((new Set().add('com.google.androidbrowserhelper:androidbrowserhelper:2.0.0')));
+      expect(features.buildGradle.dependencies).toEqual((new Set().
+          add('com.google.androidbrowserhelper:androidbrowserhelper:2.0.0')));
       expect(features.buildGradle.repositories).toEqual(emptySet);
       expect(features.launcherActivity.imports).toEqual(emptySet);
       expect(features.launcherActivity.launchUrl).toEqual([]);
+    });
+
+    it('Creates from empty features with alpha features unabled', () => {
+      const manifest = {
+        features: {},
+        fallbackType: 'customtabs',
+        alphaDependencies: {enabled: true},
+      } as TwaManifest;
+      const features = new FeatureManager(manifest);
+      expect(features.buildGradle.dependencies).toEqual((new Set().
+          add('com.google.androidbrowserhelper:androidbrowserhelper:1.4.0-alpha01')));
     });
 
     it('Adds INTERNET permission when WebView fallback is enabled', () => {
--- a/packages/core/src/spec/lib/features/FeatureManagerSpec.ts
+++ b/packages/core/src/spec/lib/features/FeatureManagerSpec.ts
@@ -72,8 +72,8 @@ describe('FeatureManager', () => {
       expect(features.applicationClass.imports).toEqual(emptySet);
       expect(features.applicationClass.onCreate).toEqual([]);
       expect(features.applicationClass.variables).toEqual([]);
-      expect(features.buildGradle.dependencies).toEqual((new Set().
-          add('com.google.androidbrowserhelper:androidbrowserhelper:2.0.0')));
+      expect(features.buildGradle.dependencies).toContain(
+          'com.google.androidbrowserhelper:androidbrowserhelper:2.0.0');
       expect(features.buildGradle.repositories).toEqual(emptySet);
       expect(features.launcherActivity.imports).toEqual(emptySet);
       expect(features.launcherActivity.launchUrl).toEqual([]);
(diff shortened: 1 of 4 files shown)


packages/core/template_project/app/build.gradle
```diff
@@ -196,5 +196,11 @@ dependencies {
     <% for(const dep of buildGradle.dependencies) { %>
         implementation '<%= dep %>'
     <% } %>
-    implementation 'com.google.androidbrowserhelper:androidbrowserhelper:2.0.0'    
+     if (twaManifest.alphaFeatures && twaManifest.alphaFeatures.enabled) {
+      this.buildGradle.dependencies.add(
+          'com.google.androidbrowserhelper:androidbrowserhelper:1.4.0-alpha01');
```
PEConn, line 201: I'm not 100% sure how this works, maybe @andreban can help me, but I don't think this will add the lines to build.gradle.

We're adding the dependencies to the list of dependencies, but we added the list of dependencies to the file up on line 196-198. So we're adding this *after* we've read from the list.

andreban: I'm surprised running `init` or `update` on this template works and would expect an error on `twaManifest.alphaFeatures`, as `twaManifest` would be undefined. The code should look like something like this:
```
<% if (alphaFeatures && alphaFeatures.enabled { %>
    implementation 'com.google.androidbrowserhelper:androidbrowserhelper:1.4.0-alpha01' 
<% } else  { %> 
    implementation 'com.google.androidbrowserhelper:androidbrowserhelper:2.0.0' 
<% } %> 
```

packages/core/template_project/app/build.gradle
```diff
@@ -196,5 +196,9 @@ dependencies {
     <% for(const dep of buildGradle.dependencies) { %>
         implementation '<%= dep %>'
     <% } %>
-    implementation 'com.google.androidbrowserhelper:androidbrowserhelper:2.0.0'    
+    if (twaManifest.alphaDependencies && twaManifest.alphaDependencies.enabled) {
+        implementation 'com.google.androidbrowserhelper:androidbrowserhelper:1.4.0-alpha01' 
+    } else {
+        implementation 'com.google.androidbrowserhelper:androidbrowserhelper:2.0.0' 
+    }
```
PEConn, line 203: Do we need both this and the changes in FeatureManager? Surely we just need the one.

chenlevy24: Ah, so I will remove the changes from `build.gradle`

packages/core/template_project/app/build.gradle
```diff
@@ -197,4 +197,4 @@ dependencies {
         implementation '<%= dep %>'
     <% } %>
     implementation 'com.google.androidbrowserhelper:androidbrowserhelper:2.0.0'
```
andreban, line 199: When adding the dependency via FeatureManager, this line will duplicate the dependency.

packages/core/template_project/app/build.gradle
```diff
@@ -197,4 +197,4 @@ dependencies {
         implementation '<%= dep %>'
     <% } %>
     implementation 'com.google.androidbrowserhelper:androidbrowserhelper:2.0.0'    
-}
+}
```
andreban, line 200: nit: Missing end of line.

packages/core/src/spec/lib/features/FeatureManagerSpec.ts
```diff
@@ -72,7 +72,8 @@ describe('FeatureManager', () => {
       expect(features.applicationClass.imports).toEqual(emptySet);
       expect(features.applicationClass.onCreate).toEqual([]);
       expect(features.applicationClass.variables).toEqual([]);
-      expect(features.buildGradle.dependencies).toEqual(emptySet);
+      expect(features.buildGradle.dependencies).
```
andreban, line 75: Can we add a test for  `"alphaDependencies": {"enabled": true}`?

packages/core/src/spec/lib/features/FeatureManagerSpec.ts
```diff
@@ -72,13 +72,24 @@ describe('FeatureManager', () => {
...
 
+    it('Creates from empty features with alpha features unabled', () => {
+      const manifest = {
+        features: {},
+        fallbackType: 'customtabs',
+        alphaDependencies: {enabled: true},
+      } as TwaManifest;
+      const features = new FeatureManager(manifest);
+      expect(features.buildGradle.dependencies).toEqual((new Set().
+          add('com.google.androidbrowserhelper:androidbrowserhelper:1.4.0-alpha01')));
```
andreban, line 90: A note: this ensures that the **only** dependency in the apps is androidbrowserhelper. Is this indeed what we want to test? If we only want to check if the correct dependency was added, we could do something like this (same thing for the validation above):
```suggestion
      expect(features.buildGradle.dependencies)
          .toContain('com.google.androidbrowserhelper:androidbrowserhelper:1.4.0-alpha01');
```

packages/core/src/spec/lib/features/FeatureManagerSpec.ts
```diff
@@ -72,12 +72,24 @@ describe('FeatureManager', () => {
...
       expect(features.applicationClass.variables).toEqual([]);
-      expect(features.buildGradle.dependencies).toEqual(emptySet);
+      expect(features.buildGradle.dependencies).toContain(
+          'com.google.androidbrowserhelper:androidbrowserhelper:2.0.0');
       expect(features.buildGradle.repositories).toEqual(emptySet);
       expect(features.launcherActivity.imports).toEqual(emptySet);
       expect(features.launcherActivity.launchUrl).toEqual([]);
     });
 
+    it('Creates from empty features with alpha features unabled', () => {
```
andreban, line 82: nit:
```suggestion
    it('Creates from empty features with alpha features disabled', () => {
```

andreban (approved):
LGTM, Thank you

PEConn (approved):
Also LGTM

PR #366, opened 2020-10-14
Location delegation

chenlevy24:
Added location delegation feature, as was done in PR #209 .


packages/core/src/lib/TwaManifest.ts
```diff
@@ -57,12 +58,14 @@ const DEFAULT_APP_VERSION_NAME = DEFAULT_APP_VERSION_CODE.toString();
 const DEFAULT_SIGNING_KEY_PATH = './android.keystore';
 const DEFAULT_SIGNING_KEY_ALIAS = 'android';
 const DEFAULT_ENABLE_NOTIFICATIONS = false;
+const DEFAULT_ENABLE_LOCATIONS = false;
```
PEConn, line 61: nit: This should be LOCATION not LOCATIONS.

packages/core/src/lib/features/FeatureManager.ts
```diff
@@ -43,11 +44,19 @@ export class FeatureManager {
...
+  delegationService = {
+    imports: new Set<string>(),
+    constructor: new Array<string>(),
+  };
 
   /**
    * Builds a new intance from a TwaManifest.
    */
   constructor(twaManifest: TwaManifest) {
+    if (twaManifest.features.locationDelegation) {
```
PEConn, line 56: Can this be `twaManifest.features.locationDelegation && twaManifest.features.locationDelegation.enabled`?

packages/core/template_project/app/src/main/java/DelegationService.java
```diff
@@ -0,0 +1,15 @@
+package <%= packageId %>;
+
+<% for(const imp of delegationService.imports) { %>
+    import <%= imp %>;
```
PEConn, line 4: nit: No indentation here (to make the generated code look normal).

packages/core/src/lib/features/FeatureManager.ts
```diff
@@ -81,16 +90,18 @@ export class FeatureManager {
...
-      this.applicationClass.imports.add(imp);
-    });
-
-    feature.applicationClass.variables.forEach((imp) => {
-      this.applicationClass.variables.push(imp);
-    });
-
-    if (feature.applicationClass.onCreate) {
-      this.applicationClass.onCreate.push(feature.applicationClass.onCreate);
+    if (feature.applicationClass !== undefined) {
```
PEConn, line 93: Why does this need to be checked? I thought `applicationClass` can't be null/undefined.

packages/core/src/lib/features/FeatureManager.ts
```diff
@@ -103,20 +114,32 @@ export class FeatureManager {
...
-    });
-
-    feature.launcherActivity.variables.forEach((imp) => {
-      this.launcherActivity.variables.add(imp);
-    });
-
-    feature.launcherActivity.methods.forEach((imp) => {
-      this.launcherActivity.methods.add(imp);
-    });
+    if (feature.launcherActivity !== undefined) {
```
PEConn, line 117: (same as above)

packages/core/src/lib/features/LocationDelegationFeature.ts
```diff
@@ -0,0 +1,62 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {EmptyFeature} from './EmptyFeature';
+
+export type LocationDelegationConfig = {
+}
+
+export class LocationDelegationFeature implements EmptyFeature {
```
PEConn, line 22: Can we make this class look like `FirstRunFlagFeature` or `AppsFlyerFeature`? They both extend `EmptyFeature` but instead of defining everything they just set the things that are non-empty in the constructor.

packages/core/src/lib/features/LocationDelegationFeature.ts
```diff
@@ -0,0 +1,62 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {EmptyFeature} from './EmptyFeature';
+
+export type LocationDelegationConfig = {
+}
```
andreban, line 20: We are adding an `enabled` attribute to every config.

```suggestion
export type LocationDelegationConfig = {
  enabled: boolean;
}
```

andreban (changes requested):

chenlevy24:
I changed the implementation of `LocationDelegationFeature` according to the assumption that `delegationService` isn't a mandatory field in `Feature`. If it should be, I will change it accordingly.

packages/core/src/lib/TwaManifest.ts
```diff
@@ -289,6 +292,9 @@ export class TwaManifest {
       enableNotifications: DEFAULT_ENABLE_NOTIFICATIONS,
       shortcuts: shortcuts,
       webManifestUrl: webManifestUrl.toString(),
+      features: {locationDelegation: {
+        enabled: DEFAULT_ENABLE_LOCATION},
+      },
```
PEConn, line 297: nit: could you format this?

```
features: {
  locationDelegation: { enabled: DEFAULT_ENABLE_LOCATION },
}
```

(or having enabled on its own line is OK too)

packages/core/src/lib/features/FeatureManager.ts
```diff
@@ -43,11 +44,20 @@ export class FeatureManager {
...
+    imports: new Set<string>(),
+    constructor: new Array<string>(),
+  };
 
   /**
    * Builds a new intance from a TwaManifest.
    */
   constructor(twaManifest: TwaManifest) {
+    if (twaManifest.features.locationDelegation &&
+        twaManifest.features.locationDelegation.enabled) {
```
PEConn, line 57: I just realized that this can be written as:

```
if (twaManifest.features.locationDelegation?.enabled)
```

Look at this - https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

I've always found that feature pretty cool (you can also change line 61).

packages/core/src/lib/features/FeatureManager.ts
```diff
@@ -106,17 +114,25 @@ export class FeatureManager {
...
       this.launcherActivity.methods.add(imp);
     });
 
-    if (feature.launcherActivity?.launchUrl) {
+    if (feature.launcherActivity.launchUrl) {
       this.launcherActivity.launchUrl.push(feature.launcherActivity.launchUrl);
     }
+
+    // Adds properties to delegationService
+    if (feature.delegationService !== undefined) {
```
PEConn, line 129: `delegationService` is non nullable, so this isn't necessary.

PEConn (commented):
This is looking good, thanks!

EiraGe:
@googlebot I consent.

PEConn (approved):
Looks good to me

packages/core/src/lib/features/EmptyFeature.ts
```diff
@@ -54,6 +54,14 @@ export class EmptyFeature implements Feature {
     methods: new Array<string>(),
   };
 
+  delegationService: {
+    imports: string[];
+    constructor: string;
```
andreban, line 59: Shouldn't this one be optional (like `launchUrl` on `LauncherActivity`?

packages/core/src/lib/features/EmptyFeature.ts
```diff
@@ -54,6 +54,14 @@ export class EmptyFeature implements Feature {
...
+  } = {
+    imports: new Array<string>(),
+    constructor: '',
```
andreban, line 62: Then, this line can be removed.

andreban (approved):
LGTM, Thank you!

PR #372, opened 2020-10-22
Web share target support

chenlevy24:
Added an optional share_target field to the webManifest


packages/core/src/lib/types/WebManifest.ts
```diff
@@ -32,6 +32,19 @@ export interface WebManifestShortcutJson {
 
 type WebManifestDisplayMode = 'browser' | 'minimal-ui' | 'standalone' | 'fullscreen';
 
+export interface ShareTargetParams {
```
PEConn, line 35: It may be worth adding a comment that these two data structures follow the definitions from here:

https://w3c.github.io/web-share-target/

PEConn (approved):
Looking good.

andreban (approved):
LGTM

PEConn (approved):

PR #377, opened 2020-10-27
Web share target support - string resources save

chenlevy24:
Now we save the share target data as a string resource.


PEConn (approved):
LGTM

PR #378, opened 2020-10-27
Draft: Makes svg2img dependency optional

andreban:


packages/core/src/lib/util.ts
```diff
@@ -140,7 +142,17 @@ export function findSuitableIcon(
 
   let largestIcon: WebManifestIcon | null = null;
   let largestIconSize = 0;
+  
   for (const icon of icons) {
+    // Use the mime-type from the icon or look up from the URL if one is not provided.	
+    const mimeType = icon.mimeType || lookup(icon.src);
+
+    // We don't support SVG images, so skip SVG icons.
```
PEConn, line 150: This comment is a little misleading - maybe something like "ignore SVG icons if SVG support isn't set up".

packages/core/src/lib/util.ts
```diff
@@ -140,7 +142,17 @@ export function findSuitableIcon(
...
   let largestIcon: WebManifestIcon | null = null;
   let largestIconSize = 0;
+  
   for (const icon of icons) {
+    // Use the mime-type from the icon or look up from the URL if one is not provided.	
+    const mimeType = icon.mimeType || lookup(icon.src);
+
+    // We don't support SVG images, so skip SVG icons.	
+    if (mimeType && mimeType.startsWith('image/svg') && !isSvgSupported()) {	
+      console.debug(`Skipping SVG: ${icon.src}`);
```
PEConn, line 152: Probably also tell users how they can fix this?

PEConn (commented):
This is looking good, just a few wording comments.

PR #382, opened 2020-10-28
Web share target support - fixed bug

chenlevy24:
Fixed the web share target's string resource save on `build.gradle`

### DIFF
--- a/packages/core/template_project/app/build.gradle
+++ b/packages/core/template_project/app/build.gradle
@@ -72,9 +72,9 @@ android {
             resValue "string", "webManifestUrl", '<%= webManifestUrl %>'
         <% } %>
 
-        <% if (twaManifest.webShare) { %>
+        <% if (shareTarget) { %>
             // The data for the app to support web share target.
-            resValue "string", "webShare", twaManifest.webShare
+            resValue "string", "shareTarget", '<%= JSON.stringify(shareTarget) %>'
         <% } %>
 
         // The hostname is used when building the intent-filter, so the TWA is able to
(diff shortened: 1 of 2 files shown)


packages/core/template_project/app/build.gradle
```diff
@@ -72,9 +72,9 @@ android {
             resValue "string", "webManifestUrl", '<%= webManifestUrl %>'
         <% } %>
 
-        <% if (twaManifest.webShare) { %>
+        <% if (shareTarget) { %>
             // The data for the app to support web share target.
-            resValue "string", "webShare", twaManifest.webShare
+            resValue "string", "shareTarget", '<%= JSON.stringify(shareTarget) %>'
```
andreban, line 77: Alternatively, we could add a `webShareTargetJson(): string` to `TwaManifest` and replace the call with that. Also maybe replace the `if` test with `hasShareTarget(): boolean`.

andreban (approved):
LGTM. Add a note, not as a requirement, but something to consider.

packages/core/template_project/app/build.gradle
chenlevy24: About the JSON, done. 
About the if: It wouldn't make the code better I think.

PEConn (commented):
Wait, so now we're stringifying it twice?

packages/core/template_project/app/build.gradle
chenlevy24: I'm fine with both of the options. I think they are both serves us the same

andreban: @PEConn the reason I usually add to `twaManifest` because the template is more error prone - code is harder to read and we don't get any kind of type checking. It's also easier if we want to change the implementation in the future, but still have a method that returns a `string`. As I mentioned when I approved the request, this is more a suggestion and I'm happy either way.

JudahGabriel:
Let us know when this fix is available in a stable release of Bubblewrap.

Issue #21, opened 2019-11-15
Support Web Share Target


NotWoods:
Similar to shortcuts, generate XML entries for the [Web Share Target](https://wicg.github.io/web-share-target/) API. These entries would probably be intent filters in the Android Manifest.

andreban:
Thanks for filing this.
The [shortcuts section](https://github.com/GoogleChromeLabs/llama-pack/blob/master/template_project/app/build.gradle#L31-L39) of the template project can be used to add this support.

We are missing code to extract details from an existing Web Manifest, confirm those details with the developer and then generate the shortcut section.

andreban:
This has now been implemented.

anuragvohraec:
> This has now been implemented.
It seems to be not working though.
Here is my pwa manifest file: https://yeio.app/manifest.json
It works all find when installed directly using "Add to homescreen".
However ever since pubslishing this to Play store, it do not works.
Please help.

PR #383, opened 2020-10-29
Commented out share_target implementation

andreban:
The share_target implementation is breaking the build. This PR
fixes what is breaking it.

We also set the shareTarget variable to `undefined`, effectively
disabling the feature. The reason being that no validation is
being done in the value read from the Web Manifest, which can
potentially generate an invalid TWA Manifest.

Related to #21

### DIFF
--- a/packages/core/src/lib/TwaManifest.ts
+++ b/packages/core/src/lib/TwaManifest.ts
@@ -297,7 +297,7 @@ export class TwaManifest {
       features: {
         locationDelegation: {enabled: DEFAULT_ENABLE_LOCATION},
       },
-      shareTarget: undefined,//webManifest['share_target'],
+      shareTarget: undefined, //webManifest['share_target'],
     });
     return twaManifest;
   }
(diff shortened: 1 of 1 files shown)


packages/core/src/lib/TwaManifest.ts
```diff
@@ -297,7 +297,7 @@ export class TwaManifest {
       features: {
         locationDelegation: {enabled: DEFAULT_ENABLE_LOCATION},
       },
-      shareTargetJson: JSON.stringify(webManifest['share_target']),
+      shareTarget: undefined, // webManifest['share_target'],
```
PEConn, line 300: I don't really like checking in commented out code. Potentially what we could have is a comment linking to a bug or issue explaining why it's undefined.

PEConn (approved):
I'd prefer if we got rid of the commented out code, but if we're in a rush, LGTM.

packages/core/src/lib/TwaManifest.ts
andreban: Removed the commented-out code and added a comment. I want to avoid having broken things in `master`.

PEConn: LGTM

PR #384, opened 2020-10-30
Added progress bar when downloading SDKs

### LINKED ISSUE #349: Add a Progress bar to JDK and Android SDK downloads
**Context:**
The JDK and Android SDK downloads are quite large, and we currently don't give users any feedback on progress, and it may look like the download is frozen.

**Work**
- [x] Modify `util.downloadFile` to receive a callback function and invoke it for every chunk received (#348).
- [ ] Modify `JdkInstaller#install()` to receive a callback function, and enable the CLI to display progress.
- [ ] Modify `AndroidSdkInstaller#install()` to receive a callback function, and enable the CLI to display progress.
- [ ] Modify the CLI code to pass a callback function to JdkInstaller and display a progress bar with the Jdk download.
- [ ] Modify the CLI code to pass a callback function to AndroidSdk Installer and display a progress bar with the Jdk download.

CC @joycetoh8

andreban:
- Moves JdkInstalled and AndroidSdkInstaller to `cli`, as apps
  using `core` should build their own configs.
- Displays a progress bar when downloading the JDK Sources, the
  JDK Binaries and the Android SDK.
- Displays messages when decompressing JDK Sources, JDK Binaries
  and the Android SDK.

Screen Shot 2020-11-03 at 19 46 41

Closes #349

### DIFF
--- a/packages/cli/src/lib/AndroidSdkToolsInstaller.ts
+++ b/packages/cli/src/lib/AndroidSdkToolsInstaller.ts
@@ -15,28 +15,36 @@
  */
 
 import * as path from 'path';
-import * as util from '../util';
+import {util} from '@bubblewrap/core';
+import {Prompt} from './Prompt';
+import {Presets, Bar} from 'cli-progress';
+import {green} from 'colors';
+import {enUS as messages} from './strings';
 
 const SDK_VERSION = '6609375';
 const DOWNLOAD_SDK_ROOT = 'https://dl.google.com/android/repository/';
 const WINDOWS_URL = `commandlinetools-win-${SDK_VERSION}_latest.zip`;
 const MAC_URL = `commandlinetools-mac-${SDK_VERSION}_latest.zip`;
 const LINUX_URL = `commandlinetools-linux-${SDK_VERSION}_latest.zip`;
+const ANDROID_SDK_SIZE = 86532338;
 
 /**
  * Install Android Command Line Tools by downloading the zip and
  * decompressing it.
  */
 export class AndroidSdkToolsInstaller {
+  constructor(private process: NodeJS.Process, private prompt: Prompt) {
+  }
+
   /**
    * Downloads the platform-appropriate version of Android
    * Command Line Tools.
    *
    * @param installPath {string} path to install SDK at.
    */
-  static async install(installPath: string): Promise<void> {
+  async install(installPath: string): Promise<void> {
     let downloadFileName;
-    switch (process.platform) {
+    switch (this.process.platform) {
       case 'darwin': {
         downloadFileName = MAC_URL;
         break;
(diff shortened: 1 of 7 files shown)


packages/cli/src/lib/JdkInstaller.ts
```diff
@@ -79,12 +85,30 @@ export class JdkInstaller {
...
+    }, Presets.shades_classic);
+
+    this.prompt.printMessage(messages.messageDownloadJdkSrc);
+    progressBar.start(Math.round(JDK_SOURCE_SIZE / 1024), 0);
+    await util.downloadFile(downloadSrcUrl, localSrcZipPath, (current) => {
+      progressBar.update(Math.round(current / 1024));
+    });
+    progressBar.stop();
+
+    this.prompt.printMessage(messages.messageDecompressJdkSrc);
```
PEConn, line 100: I feel you could extract this and the chunk you added below into some method.

Potentially even something static that could also be used in the Android downloader.

packages/cli/src/lib/AndroidSdkToolsInstaller.ts
```diff
@@ -15,28 +15,36 @@
...
+import {Presets, Bar} from 'cli-progress';
+import {green} from 'colors';
+import {enUS as messages} from './strings';
 
 const SDK_VERSION = '6609375';
 const DOWNLOAD_SDK_ROOT = 'https://dl.google.com/android/repository/';
 const WINDOWS_URL = `commandlinetools-win-${SDK_VERSION}_latest.zip`;
 const MAC_URL = `commandlinetools-mac-${SDK_VERSION}_latest.zip`;
 const LINUX_URL = `commandlinetools-linux-${SDK_VERSION}_latest.zip`;
+const ANDROID_SDK_SIZE = 86532338;
```
PEConn, line 29: I think this is OK, but I wonder if there's anything we can do to make sure it's always updated when SDK_VERSION is updated.

PEConn (commented):
Looking good.

packages/cli/src/lib/AndroidSdkToolsInstaller.ts
andreban: My first attempt was to use the `content-length` header, so I wouldn't have to hard-code the size, but it's not always there. As a matter of fact, this PR is missing the fact that Windows, MacOS and Linux versions likely have different sizes.

packages/cli/src/lib/JdkInstaller.ts
joycetoh8: +1

joycetoh8 (approved):
This looks great! Thanks for adding this. It'll be great to show users the progress bars, especially for the JDK source and binaries as those can take quite long.

packages/cli/src/lib/JdkInstaller.ts
joycetoh8: +1

andreban: done.

packages/cli/src/lib/AndroidSdkToolsInstaller.ts
andreban: I'm only using the fixed size for the JDK sources. Using Content-Length for JDK Bin and Android SDK.

joycetoh8 (approved):

PEConn (approved):
LGTM

PR #388, opened 2020-11-04
Web share target support

chenlevy24:


chenlevy24:
Extended the webManifest interface to support web share target completely. Moreover, added `escapeJsonString` method for formatting the shareTarget JSON on `build.gradle`.

packages/core/template_project/app/src/main/AndroidManifest.xml
```diff
@@ -55,14 +55,27 @@
             android:name="twa_generator"
             android:value="@string/generatorApp" />
 
+        <meta-data
+            android:name="android.support.customtabs.trusted.METADATA_SHARE_TARGET"
+            android:resource="@string/shareTarget"/>
+
```
andreban: We only want to add this when Share Target is enabled:
```suggestion
 <% if (shareTarget) { %>
        <meta-data
            android:name="android.support.customtabs.trusted.METADATA_SHARE_TARGET"
            android:resource="@string/shareTarget"/>
<% } %>
```

packages/core/template_project/app/src/main/AndroidManifest.xml
```diff
@@ -55,14 +55,27 @@
...
-        
+
+        <intent-filter>
+            <action android:name="android.intent.action.SEND" />
+            <action android:name="android.intent.action.SEND_MULTIPLE" />
+            <category android:name="android.intent.category.DEFAULT" />
+            <data android:mimeType="audio/*" />
+            <data android:mimeType="image/*" />
+            <data android:mimeType="video/*" />
+        </intent-filter>
```
andreban: Same for the `intent-filter`:

```suggestion
 <% if (shareTarget) { %>
        <intent-filter>
            <action android:name="android.intent.action.SEND" />
            <action android:name="android.intent.action.SEND_MULTIPLE" />
            <category android:name="android.intent.category.DEFAULT" />
            <data android:mimeType="audio/*" />
            <data android:mimeType="image/*" />
            <data android:mimeType="video/*" />
        </intent-filter>
<% } %>
```

andreban (changes requested):
A couple of changes around only including the meta-tag and intent-filter when shareTarget is enabled.

packages/core/template_project/app/src/main/AndroidManifest.xml
```diff
@@ -105,11 +105,28 @@
             <meta-data android:name="android.support.customtabs.trusted.FALLBACK_STRATEGY"
                 android:value="@string/fallbackType" />
 
+            <% if (shareTarget) { %>
```
PEConn, line 108: Why not combine this section and the one you added below? Put both bits inside a single `if (shareTarget)`?

packages/core/src/lib/util.ts
```diff
@@ -276,3 +276,16 @@ export async function getWebManifest(webManifestUrl: URL): Promise<WebManifestJs
   }
   return await response.json();
 }
+
+/**
+ * Given a string of a JSON, the function retrns an escaped string representing that string.
```
PEConn, line 281: nit: retrns -> returns

packages/core/src/lib/util.ts
```diff
@@ -276,3 +276,16 @@ export async function getWebManifest(webManifestUrl: URL): Promise<WebManifestJs
...
+ * eg: Turns every " instance into \\".
+ *
+ * @param {string} stringToReplace the string before the manipulation.
+ * @returns {string} the string after the manipulation.
+ */
+export function escapeJsonString(stringToReplace: string): string {
+  // The 'g' flag is for replacing all of the instances.
+  const regExp = new RegExp('\"', 'g');
+  return stringToReplace.replace(regExp, '\\\\\"');
```
PEConn, line 290: I am confused what this bit does - the `\\\\\"` specifically?

andreban: It replaces `"` with `\\"`. We end up with quite a few escapes for that.

chenlevy24: I figured that will be confusing so I mentioned it in the function's documentation.

andreban (approved):
LGTM

PEConn (approved):

PR #391, opened 2020-11-06
shareTarget derives full action URL

andreban:
- Derive the full action URL from the Web Manifest URL.
- Extracts mime-types from share-target.

Improves #21

### DIFF
--- a/packages/core/src/lib/TwaManifest.ts
+++ b/packages/core/src/lib/TwaManifest.ts
@@ -294,13 +294,28 @@ export class TwaManifest {
       shortcuts: shortcuts,
       webManifestUrl: webManifestUrl.toString(),
       features: {},
-      // We're leaving `shareTarget` as undefined for now, as the shareTarget is not complete
-      // and we want to avoid creating manifests with invalid values for now. The broader Web Share
-      // Target implementation is being tracked at
-      // https://github.com/GoogleChromeLabs/bubblewrap/issues/21.
-      shareTarget: undefined,
+      shareTarget: TwaManifest.verifyShareTarget(webManifestUrl, webManifest.share_target),
     });
     return twaManifest;
+  }
+
+  private static verifyShareTarget(
+      webManifestUrl: URL, shareTarget?: ShareTarget): ShareTarget | undefined {
+    if (!shareTarget || !shareTarget.action || !shareTarget.params || !shareTarget.params.files) {
+      return undefined;
+    }
+
+    for (const file of shareTarget.params.files) {
+      if (!file.accept) {
+        return undefined;
+      }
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/TwaManifest.ts
```diff
@@ -294,15 +294,30 @@ export class TwaManifest {
...
-      // https://github.com/GoogleChromeLabs/bubblewrap/issues/21.
-      shareTarget: undefined,
+      shareTarget: TwaManifest.verifyShareTarget(webManifestUrl, webManifest.share_target),
     });
     return twaManifest;
   }
 
+  private static verifyShareTarget(
+      webManifestUrl: URL, shareTarget?: ShareTarget): ShareTarget | undefined {
+    if (!shareTarget || !shareTarget.action || !shareTarget.params || !shareTarget.params.files) {
```
PEConn, line 304: Can this not be `if (!shareTarget?.action || !shareTarget.params?.files)`?

PEConn (commented):
I think this looks good - I'm a little unsure about why we're using undefined instead of null though.

packages/core/src/lib/TwaManifest.ts
andreban: Good point. done.

andreban:
Unfortunately, when a field is defined as optional like `shareTarget? : ShareTarget`), that means `ShareTarget | undefined` and returning `null` causes a compilation error. (`Type 'ShareTarget | null' is not assignable to type 'ShareTarget | undefined'.`)

packages/core/src/lib/TwaManifest.ts
```diff
@@ -294,15 +294,30 @@ export class TwaManifest {
...
-      // Target implementation is being tracked at
-      // https://github.com/GoogleChromeLabs/bubblewrap/issues/21.
-      shareTarget: undefined,
+      shareTarget: TwaManifest.verifyShareTarget(webManifestUrl, webManifest.share_target),
     });
     return twaManifest;
   }
 
+  private static verifyShareTarget(
+      webManifestUrl: URL, shareTarget?: ShareTarget): ShareTarget | undefined {
```
PEConn, line 303: The return type can be `ShareTarget?` right?

PEConn (approved):
Looks good with nit.

packages/core/src/lib/TwaManifest.ts
andreban: The optional syntax only works for attributes, variable declarations and parameters. Not for return types.

PEConn:
Oh, it seems I still don't understand TypeScript. LGTM again.

PR #392, opened 2020-11-06
Handles orientation in twa-manifest.json, AndroidManifest.xml and init

andreban:
- Handles orientation in twa-manifest.json, AndroidManifest.xml and init
- Adds an `alphaDependencies` parameter to `init`.
- Prompt for orientation only happens when that's enabled.

### DIFF
--- a/packages/cli/src/lib/cmds/init.ts
+++ b/packages/cli/src/lib/cmds/init.ts
@@ -15,19 +15,21 @@
  */
 
 import * as fs from 'fs';
-import {Config, DisplayModes, JdkHelper, KeyTool, TwaGenerator, TwaManifest}
+import {Config, DisplayModes, JdkHelper, KeyTool, Orientations, TwaGenerator, TwaManifest}
   from '@bubblewrap/core';
 import {validateHost, validateColor, createValidateString, validateDisplayMode, validatePackageId,
-  validateImageUrl, validateOptionalImageUrl, validateInteger} from '../inputHelpers';
-import {ParsedArgs} from 'minimist';
+  validateImageUrl, validateOptionalImageUrl, validateInteger, validateOrientation}
+  from '../inputHelpers';
 import {APP_NAME} from '../constants';
 import {Prompt, InquirerPrompt} from '../Prompt';
 import {enUS as messages} from '../strings';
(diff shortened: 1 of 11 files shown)


packages/core/src/lib/TwaManifest.ts
```diff
@@ -45,6 +45,18 @@ export function asDisplayMode(input: string): DisplayMode | null {
   return DISPLAY_MODE_VALUES.includes(input) ? input as DisplayMode : null;
 }
 
+const ORIENTATION_VALUES = ['any', 'natural', 'landscape', 'portrait', 'portrait-primary',
```
PEConn, line 48: Could you add a comment linking to the web manifest spec for orientation values?

packages/cli/src/lib/cmds/init.ts
```diff
@@ -77,6 +79,15 @@ async function confirmTwaConfig(twaManifest: TwaManifest, prompt: Prompt): Promi
       validateDisplayMode,
   );
 
+  // TODO(andreban): Remove the alpha check when androidx.browser becomes stable.
+  if (twaManifest.alphaDependencies) {
```
PEConn, line 83: So where does this get set? What does a user have to do to get this prompt to appear?

packages/core/src/lib/TwaManifest.ts
andreban: This is actually from android-browser-helper: https://github.com/GoogleChrome/android-browser-helper/blob/alpha/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java#L191-L216. I'll add a link to that.

packages/cli/src/lib/cmds/init.ts
andreban: Adding `--alphaDependencies` when running `init` does it. Will add to the docs.

andreban: done.

packages/core/src/lib/TwaManifest.ts
andreban: done.

PEConn (approved):

PR #399, opened 2020-11-15
Adds screen orientation to LauncherActivity

andreban:
Adds screen orientation to LauncherActivity so orientation is locked on the splash screen.

### DIFF
--- a/packages/core/template_project/app/build.gradle
+++ b/packages/core/template_project/app/build.gradle
@@ -143,6 +143,17 @@ android {
 
         resValue "bool", "enableSiteSettingsShortcut", twaManifest.enableSiteSettingsShortcut
         resValue "string", "orientation", twaManifest.orientation
+
+        // This is the value of the screenOrientation for the LauncherActivity which determines the
+        // orientation of the splash screen. The "landscape" and "portrait" values from the TWA
+        // manifest directly maps ot "landscape" and "portrait" screenOrientation. Other values
+        // can't be directly mapped and are left as "unspecified". See more at:
+        // https://developer.android.com/guide/topics/manifest/activity-element#screen
+        def splashOrientation = "unspecified";
+        if (twaManifest.orientation == "landscape" || twaManifest.orientation == "portrait") {
+            splashOrientation.orientation
+        }
+        resValue "string", "splashOrientation", splashOrientation
     }
     buildTypes {
         release {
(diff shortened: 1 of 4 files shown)


packages/core/template_project/app/build.gradle
```diff
@@ -143,6 +143,17 @@ android {
 
         resValue "bool", "enableSiteSettingsShortcut", twaManifest.enableSiteSettingsShortcut
         resValue "string", "orientation", twaManifest.orientation
+
+        // This is the value of the screenOrientation for the LauncherActivity which determines the
+        // orientation of the splash screen. The "landscape" and "portrait" values from the TWA
+        // manifest directly maps ot "landscape" and "portrait" screenOrientation. Other values
```
PEConn, line 149: typo: "ot" -> "to"

packages/core/template_project/app/build.gradle
```diff
@@ -143,6 +143,17 @@ android {
...
+        // can't be directly mapped and are left as "unspecified". See more at:
```
PEConn, line 150: It feels a bit odd that `portrait-primary` and `portrait-secondary` can't be mapped to `portrait`.

PEConn (approved):
LGTM with comments

packages/core/template_project/app/build.gradle
andreban: It may be possible, I just didn't look into it yet. I didn't really test his change on a a device / emulator yet.

andreban: From the [Android Manifest](https://developer.android.com/guide/topics/manifest/activity-element#screen) and [web lockOrientation](https://developer.mozilla.org/en-US/docs/Web/API/Screen/lockOrientation], it seems we would map:

|Web Manifest|Android Manifest|
|----|:----|
|`portrait`|`userPortrait`|
|`portrait-primary`|`portrait`|
|`portrait-secondary`|`reversePortrait`|
|`landscape`|`userLandscape`|
|`landscape-primary`|`landscape`|
|`landscape-secondary`|`reverseLandscape`|

Does this make sense?

PEConn: Yeah, I think that makes sense.

andreban:
@PEConn moved the code that maps the web manifest orientation to the Android orientation to TS instead of inside the build.grade. PTAL.

PEConn (approved):
LGTM

PR #409, opened 2020-11-26
Turns Notification Delegation on by default

andreban:


andreban:
Hey @JudahGabriel, we are considering changing the default value here and set `enableNotifications` to `true` by default. One of the reasons for that is that it will be one less thing developers need to know to enable notifications. The other reason is that features like Web Share Target, Location Delegation and the upcoming Play Billing need to export the Delegation Service, which needs this set to true (so, the field is more of a hint right now than a flag). Wondering how this affects PWA Builder. Would you mind taking a look?

FluorescentHallucinogen:
Do notifications work in Firefox or other browsers different to Chome while notification delegation is enabled? If not, enabling notification delegation by default will affect them.

andreban:
@FluorescentHallucinogen browsers have to implement code for routing the notification to their app on their side, so if a browser doesn't support delegation their behaviour will remain the same if enabled in the app (notification will happen through the browser UI itself).

JudahGabriel:
Hey guys. If enableNotifications is set to true by default, PWABuilder would be cool with that. We'll also enable it by default on our end. Let me know when this feature is pushed out in public release.

PR #410, opened 2020-11-26
Move the DelegationService constructor fields into onCreate.

PEConn:
I probably will want to change the name of the classConstructor field, but with this in place, your orientation fix and with notifications enabled, TWA Play Billing integration finally works.


andreban (approved):
LGTM - It seems the only other place where code is added to `classConstructor` is in the LocationDelegationFeature at https://github.com/GoogleChromeLabs/bubblewrap/blob/master/packages/core/src/lib/features/LocationDelegationFeature.ts#L33-L34

It does make sense to use `onCreate` there as well.

PEConn:
Thanks for the merge!

I'll rename the variable in a follow up PR.

PR #412, opened 2020-11-27
Rename classConstructor to onCreate.

PEConn:
This PR follows up from #410 and makes the variable name a bit more apt.


andreban (approved):
LGTM, Thanks for the follow-up PR!

Issue #416, opened 2020-11-30
Log output during bubblewrap update is ugly.


PEConn:
Create an invalid twa-manifest.json file, for example enable play billing but not alpha dependencies, so your file contains:

```
  "features": {
    "playBilling": {
      "enabled": true
    }
  },
  "alphaDependencies": {
    "enabled": false
  },
```

The output to bubblewrap update is:

```
Generating Android Project.
 >> [░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░] 0%FeatureManager WARNING Skipping PlayBil.
```

(the string got cut in half because my terminal wasn't wide enough.)

So it seems that the progress bar doesn't play nicely with outputting warnings.

PEConn:
(the above is the output after the command completed, so not only was the warning printed on the same line, the progress bar should display 100%, not 0%).

joycetoh8:
@PEConn @andreban can you confirm that this issue has been resolved by Peter's PR #417 ? If so, we can go ahead and close this. Thanks!

PEConn:
Yeah, I think my change fixed this issue.

PR #417, opened 2020-11-30
Add BufferedLog and use it when generating the TWA.

PEConn:
This PR adds the `BufferedLog` class, a log that wraps another log and delays calls to any of the logging methods until `flush` is called.

This is used when we want to log something while showing a progress bar. If we logged while showing a progress bar we'll get mess output (Issue #416). Now we can buffer all the log messages until the progress bar has finished.

It also changes the warning presented when the user is trying to build with location delegation and play billing, but not with alpha dependencies into an error. In a follow up PR I'll add a check to ensure `enableNotifications` is `true`. I also believe that we should make the bubblewrap return code be non-zero as well.

### DIFF
--- a/packages/core/src/lib/BufferedLog.ts
+++ b/packages/core/src/lib/BufferedLog.ts
@@ -23,13 +23,14 @@ enum LogLevel {
   Error,
 }
 
+/** Data class to store level and message of pending */
 class PendingLog {
   constructor(public level: LogLevel, public message: string) {}
 }
 
 /**
  * A Log that wraps another Log, saving up all the log calls to be applied when flush is called.
- * 
+ *
  * It doesn't currently support arguments to the log messages.
  */
 export class BufferedLog implements Log {
@@ -37,16 +38,18 @@ export class BufferedLog implements Log {
 
   constructor(private innerLog: Log) {}
 
+  /* eslint-disable no-invalid-this */
   debug = this.addLogFunction(LogLevel.Debug);
   info = this.addLogFunction(LogLevel.Info);
   warn = this.addLogFunction(LogLevel.Warn);
   error = this.addLogFunction(LogLevel.Error);
+  /* eslint-enable no-invalid-this */
 
   /** Creates a function that adds a log at the given level. */
   private addLogFunction(level: LogLevel): (message: string) => void {
-    return (message: string) => {
+    return (message: string): void => {
       this.pendingLogs.push(new PendingLog(level, message));
-    }
+    };
   }
 
   setVerbose(verbose: boolean): void {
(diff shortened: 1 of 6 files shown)


packages/core/src/lib/BufferedLog.ts
```diff
@@ -0,0 +1,82 @@
...
+  warn = this.addLogFunction(LogLevel.Warn);
+  error = this.addLogFunction(LogLevel.Error);
+  /* eslint-enable no-invalid-this */
+
+  /** Creates a function that adds a log at the given level. */
+  private addLogFunction(level: LogLevel): (message: string) => void {
+    return (message: string): void => {
+      this.pendingLogs.push(new PendingLog(level, message));
+    };
+  }
```
andreban, line 53:
```suggestion
  private pendingLogs: Array<PendingLog> = [];
  public debug: (message: string) => void;
  public info: (message: string) => void;
  public warn: (message: string) => void;
  public error: (message: string) => void;

  constructor(private innerLog: Log) {
    this.debug = this.addLogFunction(LogLevel.Debug);
    this.info = this.addLogFunction(LogLevel.Info);
    this.warn = this.addLogFunction(LogLevel.Warn);
    this.error = this.addLogFunction(LogLevel.Error);
  }

  /** Creates a function that adds a log at the given level. */
  private addLogFunction(level: LogLevel): (message: string) => void {
    return (message: string): void => {
      this.pendingLogs.push(new PendingLog(level, message));
    };
  }
```

PEConn: Yeah, that's a good suggestion.

So as far as I can tell (purely just because it compiles and seems to work), what I've written is valid code. So this means that the lint check is wrong. In which case is it better to work around the faulty lint tool or to just disable it temporarily so we can carry on?

(Or maybe there is something actually wrong with my code that I don't understand.)

PEConn: Because I'm curious to know if the problem is with eslint or with my code, I [filed an issue with eslint](https://github.com/eslint/eslint/issues/13894).

andreban: FWIW, this is what the transpiled code looks like:
```typescript
class BufferedLog {
    constructor(innerLog) {
        this.innerLog = innerLog;
        this.pendingLogs = [];
        /* eslint-disable no-invalid-this */
        this.debug = this.addLogFunction(LogLevel.Debug);
        this.info = this.addLogFunction(LogLevel.Info);
        this.warn = this.addLogFunction(LogLevel.Warn);
        this.error = this.addLogFunction(LogLevel.Error);
        // This is the code inside the constructor.
        console.log('Where does the constructor code go?');
    }
    /* eslint-enable no-invalid-this */
    /** Creates a function that adds a log at the given level. */
    addLogFunction(level) {
        return (message) => {
            this.pendingLogs.push(new PendingLog(level, message));
        };
    }
```
I was wondering if that code could lead to some kind of undefined behaviour, but doesn't look like the case. Would you mind adding a link to the issue in the comment and a TO-DO to checkout out the issue response in the future? Besides that, LGTM then.

andreban (approved):
LGTM with the note on adding a comment to the eslint issue.

PR #418, opened 2020-12-01
Display error when Play Billing or Location Delegation are enabled but notifications aren't.

PEConn:


andreban:
Latest commit seems to have broken tests. Otherwise, LGTM (this change does seem to include code from #417)

andreban (approved):
LGTM, Thank you!

PR #420, opened 2020-12-07
Updates project dependencies

andreban:
- Uses androidbrowserhelper:2.1.0.
- Uses locationdelegation:1.0.0.
- Uses billing:1.0.0-alpha05.
- Updates CLI to always ask the orientation question.
- Updates LocationDelegation feature to not depend on alphaDependencies

### DIFF
--- a/packages/core/src/lib/features/FeatureManager.ts
+++ b/packages/core/src/lib/features/FeatureManager.ts
@@ -56,12 +56,7 @@ export class FeatureManager {
    */
   constructor(twaManifest: TwaManifest, log: Log = new ConsoleLog('FeatureManager')) {
     if (twaManifest.features.locationDelegation?.enabled) {
-      if (twaManifest.alphaDependencies?.enabled) {
-        this.addFeature(new LocationDelegationFeature());
-      } else {
-        log.error('Skipping LocationDelegationFeature. '+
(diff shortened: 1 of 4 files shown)


packages/core/src/lib/features/FeatureManager.ts
```diff
@@ -56,12 +56,7 @@ export class FeatureManager {
    */
   constructor(twaManifest: TwaManifest, log: Log = new ConsoleLog('FeatureManager')) {
     if (twaManifest.features.locationDelegation?.enabled) {
-      if (twaManifest.alphaDependencies?.enabled) {
-        this.addFeature(new LocationDelegationFeature());
-      } else {
-        log.error('Skipping LocationDelegationFeature. '+
-            'Enable alphaDependencies to add LocationDelegationFeature.');
-      }
+      this.addFeature(new LocationDelegationFeature());
```
PEConn, line 59: We can get rid of the Play Billing check below too, right?

PEConn (approved):

packages/core/src/lib/features/FeatureManager.ts
andreban: Technically, billing is bringing in an alpha dependency, so I thought it made sense to leave this in to make it clear developers are aware?

PEConn (approved):
Sounds good, LGTM

PR #424, opened 2020-12-10
Added .bat to sdkmanager before checking to see if the file exists.

david-fry:
Otherwise it will fail on initial use on Windows 10.


andreban (approved):
LGTM!, Thanks for catching this and submitting the PR!

PR #427, opened 2020-12-17
Retain task state by default

fpetitit:
Hello and thanks for this great work.

I spent a few time to figure out why my app was restarting at the `starturl` each time it was relaunched from the homescreen, even a few second after switching with an other app.

It seems that adding `android:alwaysRetainTaskState="true"` in `AndroidManifest.xml` is the standard solution to prevent that, and that it is a generally observed behaviours on other Android app (as well as for Chrome or other browsers too).

What do you think about adding this parameter by default ?


andreban:
@PEConn for thoughts

andreban (approved):
LGTM, thanks for adding this

PR #432, opened 2021-01-05
Remove mention to slack channel

andreban:
The ChromiumDev slack channel is being deprecated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,12 +39,10 @@ compare Trusted Web Activity projects against the quality criteria.
 ## Community
 
 We welcome anyone who wants to contribute with issues, feedback, feature requests or just
-generally discuss Bubblewrap. Those are a alternative ways developers can participate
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,14 +39,12 @@ compare Trusted Web Activity projects against the quality criteria.
 ## Community
 
 We welcome anyone who wants to contribute with issues, feedback, feature requests or just
-generally discuss Bubblewrap. There are a couple of other ways developers can participate
+generally discuss Bubblewrap. Those are a alternative ways developers can participate
```
PEConn, line 42: "Those are a alternative way" doesn't sound good (It should probably be "This is an alternative way").

However, I think it would read better as "Alternatively, developers can ..."

andreban: updated

PEConn (approved):

PR #440, opened 2021-01-15
Ensure embedded and twa start-urls match

andreban:


packages/core/src/lib/TwaGenerator.ts
```diff
@@ -250,16 +250,23 @@ export class TwaGenerator {
     }));
   }
 
-  private async writeWebManifest(webManifestUrl: URL, targetDirectory: string): Promise<void> {
-    const response = await fetch(webManifestUrl);
+  private async writeWebManifest(twaManifest: TwaManifest, targetDirectory: string): Promise<void> {
+    if (!twaManifest.webManifestUrl) {
```
PEConn, line 254: Are we just failing silently in this case?

packages/core/src/lib/TwaGenerator.ts
```diff
@@ -250,16 +250,23 @@ export class TwaGenerator {
     }));
   }
 
-  private async writeWebManifest(webManifestUrl: URL, targetDirectory: string): Promise<void> {
-    const response = await fetch(webManifestUrl);
+  private async writeWebManifest(twaManifest: TwaManifest, targetDirectory: string): Promise<void> {
```
PEConn, line 253: I'm not quite sure what's going on with the typing here - how have you changed the first parameter of this method from a URL to a TwaManifest but you haven't changed any of the callers?

PEConn (commented):
If this is a WIP, LGTM, one or two points if this is meant for landing.

packages/core/src/lib/TwaGenerator.ts
andreban: IIRC, this is only called here: https://github.com/GoogleChromeLabs/bubblewrap/pull/440/files#diff-6020d7a7812abc3186bdd457af35e210e130e887a016dfcbdfffe9b2e5fbc304R403 and the line invoking this was changed from `this.writeWebManifest(twaManifest.webManifestUrl, targetDirectory);` to `this.writeWebManifest(twaManifest, targetDirectory);`

packages/core/src/lib/TwaGenerator.ts
andreban: Hum, in fact, this is being checked just before calling the method. Since this is an internal method, I can probably just remove the check.

packages/core/src/lib/TwaGenerator.ts
PEConn: Oh right, sorry, I didn't twig that line 403 was in a different method that called this one. (I didn't see the GitHub code collapsing.)

PEConn (approved):
LGTM % the check on line 254 either being removed or throwing an exception.

packages/core/src/lib/TwaGenerator.ts
andreban: Added an error instead of just removing the check, as the linter will be unhappy if just removing it.

PEConn (approved):

PR #445, opened 2021-01-22
Fixes support for simple Web Share Target

### LINKED ISSUE #448: Adding share target results in Gradle exception
On Bubblewrap 1.8.2, is share target really supported? I see it in `TwaManifest.shareTarget`, but if I try to use it, it breaks.

Specifically, if I create a `TwaManifest` with the following `shareTarget`, Bubblewrap will fail to generate the package.

```typescript
const manifest: TwaManifest = {
   "host": "mathbymiles.com",
   "shareTarget": {
         "action":"/new-topic",
         "method":"GET",
         "enctype":"application/x-www-form-urlencoded",
         "params": {
              "title":"title",
              "text":"body"
          }
   },
   ...omitted for brevity
}
```

If I try to build the package, I receive the following error:

> Details: Error generating app package: Error: Command failed: ./gradlew assembleRelease --stacktrace
> FAILURE: Build failed with an exception.
>
> What went wrong:
> Execution failed for task ':app:generateReleaseBuildConfig'.
> Manifest file does not exist: /tmp/pwabuilder-cloudapk--36-B2nqEkzP1ZR3/app/src/main/AndroidManifest.xml

This error was originally reported to PWABuilder [here](https://github.com/pwa-builder/PWABuilder/issues/1386). Others are experiencing the same issue with different URLs, e.g. https://github.com/pwa-builder/PWABuilder/issues/1215#issuecomment-758593296.

andreban:
Closes #448

### DIFF
--- a/packages/core/template_project/app/src/main/AndroidManifest.xml
+++ b/packages/core/template_project/app/src/main/AndroidManifest.xml
@@ -124,12 +124,22 @@
             <% if (shareTarget) { %>
                 <intent-filter>
                     <action android:name="android.intent.action.SEND" />
-                    <action android:name="android.intent.action.SEND_MULTIPLE" />
+                    <% 
+                       // SEND_MULTIPLE is added when handling images, videos, etc.  
+                       if (shareTarget.params.files) { %>
+                        <action android:name="android.intent.action.SEND_MULTIPLE" />
+                    <% } %>
                     <category android:name="android.intent.category.DEFAULT" />
-                    <% for (const file of shareTarget.params.files) { %>
-                        <% for (const accept of file.accept) { %>
-                            <data android:mimeType="<%= accept %>" />
+                    <% 
+                       if (shareTarget.params.files) { %>
+                        <% for (const file of shareTarget.params.files) { %>
+                            <% for (const accept of file.accept) { %>
+                                <data android:mimeType="<%= accept %>" />
+                            <% } %>
                         <% } %>
+                    <% } else {%>
+                        // When params.files is not available, use "text/plain" the mime-type.  
+                        <data android:mimeType="text/plain" />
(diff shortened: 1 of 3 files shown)


packages/core/template_project/app/src/main/AndroidManifest.xml
```diff
@@ -124,12 +124,22 @@
             <% if (shareTarget) { %>
                 <intent-filter>
                     <action android:name="android.intent.action.SEND" />
-                    <action android:name="android.intent.action.SEND_MULTIPLE" />
+                    <% 
+                       // SEND_MULTIPLE is added when handling images, videos, etc.  
+                       if (shareTarget.params.files) { %>
+                        <action android:name="android.intent.action.SEND_MULTIPLE" />
```
PEConn, line 130: Is there any reason not to combine this block with the one below?

andreban: Just keeping a nice order in the `AndroidManifest.xml`. FWIW, I'm leaving this as a draft as I'm thinking I'd rather move this logic somewhere else and just have 2 for loops (one of actions and another for mime types)

PEConn: Ah fair enough - in which case I'll also point out how seeing you use C-style comments in the template section of an XML file also confuses me.

andreban: I honestly find this whole section hard to read... (that's why I want to move the logic)

andreban: Rewrote this bit - still need to test better, so leaving this as a draft for now.

PEConn (approved):

PR #447, opened 2021-01-25
Forces locale to en for keytool list

### LINKED ISSUE #446: Extra colon and space in assetlinks.json
Hi everyone
First of all, I would like to thank you for this great tool!

**Describe the bug**
When I run `bubblewrap build --skipPwaValidation`, it generates the `assetlinks.json` file well, but it adds a colon followed by a space at the beginning of the `sha256_cert_fingerprints`.
I have to remove it manually, otherwise the correspondence between the application and assetlinks is not ok.

**Output example:**
`"sha256_cert_fingerprints": [": 85:A0...`

**Expected output example:**
`"sha256_cert_fingerprints": ["85:A0...`

**Steps to reproduce:**
- `bubblewrap init --manifest=https://thewebsite`
- `bubblewrap build --skipPwaValidation`

**Environment**
- macOS Mojave version 10.14.5.
- @bubblewrap/cli@1.8.2
- node v10.20.0
- `"jdkPath":"/Users/username/.bubblewrap/jdk/jdk8u265-b01"`
- `"androidSdkPath":"/Users/username/.bubblewrap/android_sdk"`

andreban:
Fixes #446

### DIFF
--- a/packages/core/src/lib/jdk/KeyTool.ts
+++ b/packages/core/src/lib/jdk/KeyTool.ts
@@ -100,6 +100,11 @@ export class KeyTool {
     }
     const keyListCmd = [
       'keytool',
+      // Forces the language to 'en' in order to get the expected formatting.
+      // The JVM seems to ignore the LANG and LC_ALL variables, so we set the value
+      // when invoking the command. See https://github.com/GoogleChromeLabs/bubblewrap/issues/446
+      // for more.
+      '-J-Duser.language=en',
       '-list',
       '-v',
       `-keystore \"${keyOptions.path}\"`,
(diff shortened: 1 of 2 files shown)


cdreydemy-vp (approved):

packages/core/src/lib/jdk/KeyTool.ts
```diff
@@ -100,6 +100,11 @@ export class KeyTool {
     }
     const keyListCmd = [
       'keytool',
+      // Forces the language to 'en' in order to get the expected formatting.
+      // The JVM seems to ignore the LANG and LC_ALL variables, so we set the value
```
PEConn, line 104: I'm assuming the JVM wants something that is consistent across OSes? (Do `LANG` and `LC_ALL` exist on Windows?)

PEConn (approved):

packages/core/src/lib/jdk/KeyTool.ts
andreban: It seems `LANG` does exist on Windows. Nevertheless, this seems to the the appropriate way to do this.

PR #449, opened 2021-01-27
Adds support for sticky-immersive to Bubblewrap

### LINKED ISSUE #430: Fullscreen mode as default without immersive?
Could we add an enhancement to have fullscreen optional as either default or immersive?

So that fullscreen mode can still have the status and nav bar without the need to touch and toggle to display these items?

With the status bar being an overlay of the app content?

Maybe I'm wrong but looks like the 
packages/core/template_project/app/src/main/AndroidManifest.xml forces immersive mode if display is fullscreen?

Thanks.

_Originally posted by @JohnRSim in https://github.com/GoogleChromeLabs/bubblewrap/issues/175#issuecomment-751040575_

andreban:
Closes #430

### DIFF
--- a/packages/core/src/lib/TwaManifest.ts
+++ b/packages/core/src/lib/TwaManifest.ts
@@ -38,7 +38,7 @@ const SHORT_NAME_MAX_SIZE = 12;
 const MIN_NOTIFICATION_ICON_SIZE = 48;
 
 // Supported display modes for TWA
-const DISPLAY_MODE_VALUES = ['standalone', 'fullscreen'];
+const DISPLAY_MODE_VALUES = ['standalone', 'fullscreen', 'fullscreen-sticky'];
 export type DisplayMode = typeof DISPLAY_MODE_VALUES[number];
 export const DisplayModes: DisplayMode[] = [...DISPLAY_MODE_VALUES];
 
--- a/packages/core/src/lib/features/FeatureManager.ts
+++ b/packages/core/src/lib/features/FeatureManager.ts
@@ -82,7 +82,7 @@ export class FeatureManager {
     }
 
     this.buildGradle.dependencies.add(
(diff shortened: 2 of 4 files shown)


PEConn (approved):

JohnRSim:
@andreban this is great thanks - although would it be possible to have full screen with the android action bar/clock battery as a permanent overlay on the webview/browsertabs view.

Stick-immersive hides the top bar and only show when user swipes from the top/bottom?

I'd like to replicate the following experience with a TWA
[https://user-images.githubusercontent.com/364208/103466153-39e9d600-4d3a-11eb-9b1d-3e1766d4ea4e.jpg](url)

andreban:
`sticky-immersive` is supported by Chrome, but wasn't enabled in `android-browser-helper` and Bubblewrap. This fixed that.

It seems that what you are looking for is a [translucent status bar](https://medium.com/androiddevelopers/translucent-systembars-the-right-way-across-api-levels-and-themes-6d7ddda21396), which is not currently supported by the browser, so there isn't much we can do in `android-browser-helper` and Bubblewrap. The solution here is to file a feature-request for Chrome on https://crbug.com.

JohnRSim:
Tried out the sticky-immersive all looks great - but I did come across a UX issue not sure if it is resolvable.

1. The app I have has slide out menus like twitter you can drag your finger across the screen from left edge to right.
However on doing this it shows the status and nav bar. 

Is it possible to only enable the drag option from where the status or nav bar is displayed ie if phone is vertical only enable top bottom drag to show header/footer and disable side drag to show header/footer? 

2. The other issue I noticed is that the status bar keeps appearing and disappearing which makes the app look a little clunky.
Here are some examples. 

- Launch app going from splash screen to landing page
- Launching oAuth ie Azure and then redirecting back to landing page
- Launching biometrics
- Opening an external link

PR #452, opened 2021-02-01
Use url.host property for input validation

lennartkoester:
The current validation does not accept port numbers. However, hosts with a portnumber are valid hosts.
By testing URL creation first and using the .host property, the port does not invalidate the input.

since it gets concatinated in the build.gradle anyways:
"def launchUrl = "https://" + twaManifest.hostName + twaManifest.launchUrl"

it should not have any breaking changes other than improving QoL

### DIFF
--- a/packages/cli/src/lib/inputHelpers.ts
+++ b/packages/cli/src/lib/inputHelpers.ts
@@ -172,19 +172,18 @@ export async function validateHost(input: string): Promise<Result<string, Error>
     if (parts[0] !== 'https') {
       return Result.error(new Error(messages.errorRequireHttps));
     }
-    host = parts[1];
+    
   }
-
-  // Verify if the characters added to the domain are valid. This functions returns an empty
-  // string when the input is invalid.
-  const ascIIInput = domainToASCII(host);
-  if (ascIIInput.length === 0) {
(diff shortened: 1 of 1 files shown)


andreban:
@lennartkoester 

Thanks for the pull request. Before proceeding, can you please do a few things:

- Create an issue with [this template](https://github.com/GoogleChromeLabs/bubblewrap/issues/new?assignees=&labels=&template=bug_report.md&title=) describing the problem your PR is supposed to solve. I'm assuming the issue is related to not accepting something like `https://example.com:8080/`, but that's not clear to me.
- Update the test case at `inputHelperSpec.ts` to include the cases that your solution would be fixing, so we ensure the code works as intended and avoid regressions.
- Make sure your pull request passes the tests (`npm run test`) and lint checks (`npm run lint`). Your PR is currently failing on the CI, due to lint errors.
- Accept the CLA, as described by the `google-cla` bot, otherwise we are unable to accept the contribution.

andreban:
@lennartkoester pinging to check if you'd like to follow-up on the PR and confirm the underlying issue.

lennartkoester:
@andreban I'll follow up soon - can't make time right now :(

Issue #457, opened 2021-02-08
Allow local config / download + read JDK path from the environment


toastal:
**Is your feature request related to a problem? Please describe.**
All of the docs suggest `npm install --global ...` and the config is read from user home. If you wanted to install this tool locally to keep versions pinned (both for the CLI tool as well as the optional automatic JDK) for a specific project, you can't really. Being able to access the local config would also make this more accessible towards using tools like `nix shell` to distribute and source the JDK version. Further useful would be the ability to use the env vars to point to the JDK version so something like `direnv` could run a command `which` or `nix-build --no-out-link '<nixpkgs>' -A jdk8_headless` to find the JDK version.

**Describe the solution you'd like**
I would like to see a first-local fallback for `.bubblewrap/config` file + something like `$JAVA_HOME` or `$JDK_HOME` that could be read could override the current tooling. Alternatively, allowing the ability to pass in the config file via command line arguments could be an easy solution. I think there's now too much magic in downloading Java (like assuming all Linuxes are x86_64, etc.) that you made it too difficult to override if using a different setup.

**Describe alternatives you've considered**
I'm currently trying to pull apart `@bubblewrap/core` and override the bits I need in a custom command. This is far from ideal as I'm already starting at a bad place with the node v14.15.4 REPL 

```js
var core = require("@bubblewrap/core")
```

outputting 

```sh
Uncaught Error: libuuid.so.1: cannot open shared object file: No such file or directory
    at Object.Module._extensions..node (internal/modules/cjs/loader.js:1122:18)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
```

_EDIT:_ I actually get the same error output as above just by running the `node_modules/.bin/bubblewrap`. It would seem there's a mess of missing dependencies for Node canvas ... which we need why?

**Additional context**
Thanks for putting the tool together. I think it's nice for a lot of people that ya'll are handling the JDK version management on generic Ubuntu user's laptops. There's instructions for how to override and manually install the dependencies, but a) they're set to global and b) they could/should be easier to override.

andreban:
[...]

Issue #401, opened 2020-11-16
Option to wrap app from local folder instead of URL


AlexDev404:
**Is your feature request related to a problem? Please describe.**

I think some people like me would enjoy the feature of being able to wrap your PWA from the source for testing purposes before putting it live on the web.

**Describe the solution you'd like**

As a end-user of this, I would appreciate the ability to be able to wrap an app from a local folder instead of URL for beta testing and the such, so that I know that my project works before putting it online. It would also be pretty useful too, and will allow the app the work offline instead of always having to be online.

**Describe alternatives you've considered**

I have tried using `file://` URI scheme/method to locate my manifest locally, but unfortunately your application says it lacks the support for it.

andreban:
To confirm my understanding:

The request would be to allow `init` to generate an application from a Web Manifest located in the local filesystem. In order to generate a valid package, Bubblewrap would also the origin and the final URL where the Web Manifest will be available

Also, to test the app, you'll need the `assetlinks.json` file deployed in the domain.

Given the need for all of those, developers would need to deploy those files anyway, which makes me wonder if there's value on initialising from a Web Manifest in the local file system.

I'm not against it, but would like to hear arguments that would help us prioritise.

AlexDev404:
@andreban Yes that is the request
I'm trying to do something like an experimental PWA that I don't really want to upload to the web, I just want to wrap it up and that's it. I don't really need any fancy features, I just need it wrapped up, but your tool doesn't support wrapping files, and I haven't found any other tool *even close* to this, so I'm stuck.

andreban:
@AlexDev404 currently, the content of the Trusted Web Activity is required to be available on the open web. There is a feature-request [crbug.com/816798](https://crbug.com/816798) to allow an application to ship with a Web Package, but that would still require a counter-part on the open web to be available. #50 tracks support for that in Bubblewrap.

Have you considered publishing the PWA in a URL that's not linked anywhere? I'd be interested in understanding if this is not an option. If it was possible to load the PWA from an internal network, would it help?

PEConn:
I guess it doesn't necessarily need to be the *open* web - TWAs could work with sites on some company intranet (though they wouldn't have a graceful fallback when not attached to it).

Also, you can make a set up where you run a webserver on your computer and then port forward to Android, so for development you can have a TWA visiting a website hosted on your dev computer.

I think for me, the primary use case for this feature request is to make the development flow better, eg:

1. User updates their website locally.
[...]

PR #461, opened 2021-02-09
MVP for passing in custom config path

toastal:
I have never used TypeScript and am not familiar with the tooling or coding styles of this project so I did the bare minimum that got my code running and your tests green (with no new tests). I believe this can address some of this issues in #457 in conjunction with process substitution in Bash, Fish, etc. to make a JSON "file" on the fly.

### DIFF
--- a/packages/cli/src/lib/Cli.ts
+++ b/packages/cli/src/lib/Cli.ts
@@ -38,8 +38,7 @@ export class Cli {
     }
     const parsedArgs = minimist(args);
 
-    const config = await loadOrCreateConfig(
-      parsedArgs.config != null ? {path: parsedArgs.config}: {});
+    const config = await loadOrCreateConfig(undefined, undefined, parsedArgs.config);
 
     let command;
     if (parsedArgs._.length === 0) {
--- a/packages/cli/src/lib/Cli.ts
+++ b/packages/cli/src/lib/Cli.ts
@@ -36,9 +36,10 @@ export class Cli {
       throw new Error(`Current Node.js version is ${process.versions.node}.` +
           ' Node.js version 10 or above is required to run bubblewrap.');
     }
-    const config = await loadOrCreateConfig();
-
     const parsedArgs = minimist(args);
+
+    const config = await loadOrCreateConfig(
+      parsedArgs.config != null ? {path: parsedArgs.config}: {});
 
     let command;
     if (parsedArgs._.length === 0) {
--- a/packages/cli/src/lib/config.ts
+++ b/packages/cli/src/lib/config.ts
@@ -90,13 +90,20 @@ async function renameConfigIfNeeded(log: Log): Promise<void> {
   }
 }
 
-export async function loadOrCreateConfig(log: Log = new ConsoleLog('config'),
-    prompt: Prompt = new InquirerPrompt(), path = DEFAULT_CONFIG_FILE_PATH): Promise<Config> {
-  await renameConfigIfNeeded(log);
-  const existingConfig = await Config.loadConfig(path);
+export async function loadOrCreateConfig(
+    {log = new ConsoleLog('config'), prompt = new InquirerPrompt(), path}:
+    {log?: Log; prompt?: Prompt; path?: string}): Promise<Config> {
+  let configPath;
+  if (path === undefined) {
+    await renameConfigIfNeeded(log);
+    configPath = DEFAULT_CONFIG_FILE_PATH;
+  } else {
+    configPath = path;
+  }
+  const existingConfig = await Config.loadConfig(configPath);
   if (existingConfig) return existingConfig;
 
   const config = await createConfig(prompt);
-  await config.saveConfig(path);
+  await config.saveConfig(configPath);
   return config;
 }
--- a/packages/cli/src/lib/config.ts
+++ b/packages/cli/src/lib/config.ts
@@ -91,8 +91,9 @@ async function renameConfigIfNeeded(log: Log): Promise<void> {
 }
 
 export async function loadOrCreateConfig(
-    {log = new ConsoleLog('config'), prompt = new InquirerPrompt(), path}:
-    {log?: Log; prompt?: Prompt; path?: string}): Promise<Config> {
+    log: Log = new ConsoleLog('config'),
+    prompt: Prompt = new InquirerPrompt(),
+    path?: string): Promise<Config> {
   let configPath;
   if (path === undefined) {
     await renameConfigIfNeeded(log);
--- a/packages/cli/src/lib/cmds/doctor.ts
+++ b/packages/cli/src/lib/cmds/doctor.ts
@@ -44,7 +44,7 @@ async function androidSdkDoctor(config: Config, log: Log): Promise<boolean> {
 }
 
 export async function doctor(log: Log = new ConsoleLog('doctor')): Promise<boolean> {
-  const config = await loadOrCreateConfig({log});
+  const config = await loadOrCreateConfig(log);
   const jdkResult = await jdkDoctor(config, log);
   const androidSdkResult = await androidSdkDoctor(config, log);
   if (jdkResult && androidSdkResult) {
--- a/packages/cli/src/lib/cmds/doctor.ts
+++ b/packages/cli/src/lib/cmds/doctor.ts
@@ -44,7 +44,7 @@ async function androidSdkDoctor(config: Config, log: Log): Promise<boolean> {
 }
 
 export async function doctor(log: Log = new ConsoleLog('doctor')): Promise<boolean> {
-  const config = await loadOrCreateConfig(log);
+  const config = await loadOrCreateConfig({log});
   const jdkResult = await jdkDoctor(config, log);
   const androidSdkResult = await androidSdkDoctor(config, log);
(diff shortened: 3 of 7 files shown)


toastal:
Considering I got to here

```sh
[I] ~/I/b/f/pwa (twa *) bubblewrap init --config .bubblewrap/config.json --manifest (dhall-to-json --file manifest.dhall | psub)                                                                                                 bosthlm 
,-----.        ,--.  ,--.  ,--.
|  |) /_,--.,--|  |-.|  |-.|  |,---.,--.   ,--,--.--.,--,--.,---.
|  .-.  |  ||  | .-. | .-. |  | .-. |  |.'.|  |  .--' ,-.  | .-. |
|  '--' '  ''  | `-' | `-' |  \   --|   .'.   |  |  \ '-'  | '-' '
`------' `----' `---' `---'`--'`----'--'   '--`--'   `--`--|  |-'
                                                           `--'    
Initializing application from Web Manifest:
	-  /run/user/1000/.psub.8n72sG4Hcf


cli ERROR Only absolute URLs are supported
```

I would say it's working so far for me. I'll work on how to generate my own config file shortly.

toastal:
So far my `shell.nix` is for being able to install `bubblewrap` via `yarn` inside the shell (not globally)

```nix
{
  pkgs ? import (fetchTarball "https://github.com/NixOS/nixpkgs/archive/d1ebffd8297bd5eba83f66cae35ac8220ccdae77.tar.gz")
    { config = { android_sdk.accept_license = true; }; }
}:

with pkgs.lib;

let
  androidSdk = pkgs.androidenv.androidPkgs_9_0.androidsdk;

in pkgs.mkShell {
  name = "xxxxxx";

  basePackages = with pkgs; [
    gawk
    jq
    ripgrep
  ];

  nativeBuildInputs = with pkgs.buildPackages; [
    # node
    nodejs-14_x
    yarn

    # dhall
    dhall
    dhall-bash
    dhall-json

    # android
    androidSdk
    cairo
    giflib
    gnome2.pango
    jdk8_headless
    libjpeg
    libuuid
    librsvg
    pkg-config
  ];

  GRADLE_OPTS = "-Dorg.gradle.project.android.aapt2FromMavenOverride=${androidSdk}/libexec/android-sdk/build-tools/28.0.3/aapt2";

  LD_LIBRARY_PATH = "${makeLibraryPath (with pkgs; [
    giflib
    libjpeg
    librsvg
    libuuid
  ])}";
}
```

packages/cli/src/lib/config.ts
```diff
@@ -90,13 +90,20 @@ async function renameConfigIfNeeded(log: Log): Promise<void> {
   }
 }
 
-export async function loadOrCreateConfig(log: Log = new ConsoleLog('config'),
-    prompt: Prompt = new InquirerPrompt(), path = DEFAULT_CONFIG_FILE_PATH): Promise<Config> {
-  await renameConfigIfNeeded(log);
-  const existingConfig = await Config.loadConfig(path);
+export async function loadOrCreateConfig(
```
andreban, line 93: I'm not sure I understand this change - it seems the end result is the same as the previous version.

In the section below, `path` is only assigned  `DEFAULT_CONFIG_FILE_PATH` if the value passed is different from `null` and different from `undefined`:

```ts
export async function loadOrCreateConfig(
    prompt: Prompt = new InquirerPrompt(), path = DEFAULT_CONFIG_FILE_PATH): Promise<Config> {
```
so, generally analog to:

```ts
  if (path === undefined) {
    await renameConfigIfNeeded(log);
    configPath = DEFAULT_CONFIG_FILE_PATH;
  } else {
    configPath = path;
  }
```

the only difference being the call to `await renameConfigIfNeeded(log)`. But that call should be no-op in most cases (it's just about handling old config paths to the current one).

packages/cli/src/lib/Cli.ts
```diff
@@ -36,10 +36,11 @@ export class Cli {
       throw new Error(`Current Node.js version is ${process.versions.node}.` +
           ' Node.js version 10 or above is required to run bubblewrap.');
     }
-    const config = await loadOrCreateConfig();
-
     const parsedArgs = minimist(args);
 
+    const config = await loadOrCreateConfig(
```
andreban, line 41: If we revert `loadOrCreateConfig()`, we should be able to call `loadOrCreateConfig(null, null, parsedArgs.config`). So, the 1st and 2nd parameters get the default value and the last uses `parsedArgs.config`. If `parsedArgs.config` is null, it will use the default value.

andreban:
@toastal thank you very much for this! The contribution is much appreciated! I dropped a couple of comments.

andreban:
> Considering I got here in `fish`
> 
> ```fish
> $ bubblewrap init --config .bubblewrap/config.json --manifest (dhall-to-json --file manifest.dhall | psub)                                                                                                 bosthlm 
> ,-----.        ,--.  ,--.  ,--.
> |  |) /_,--.,--|  |-.|  |-.|  |,---.,--.   ,--,--.--.,--,--.,---.
> |  .-.  |  ||  | .-. | .-. |  | .-. |  |.'.|  |  .--' ,-.  | .-. |
> |  '--' '  ''  | `-' | `-' |  \   --|   .'.   |  |  \ '-'  | '-' '
> `------' `----' `---' `---'`--'`----'--'   '--`--'   `--`--|  |-'
>                                                            `--'    
> Initializing application from Web Manifest:
> 	-  /run/user/1000/.psub.8n72sG4Hcf
> 
> 
> cli ERROR Only absolute URLs are supported
> ```
> 
> I would say it's working so far for me. I'll work on how to generate my own config file shortly.

It seems to me you are bumping into #401 - we require an URL as the `--manifest` parameter.

toastal:
I was going to make a separate `twa-manifest.dhall` file deriving from the `manifest.dhall` file I already have where I could include some of these new attributes. I _REALLY_ which the differences in these manifests were documented somewhere else other than in the config TS code. I tried searching for documentation about the `twa-manifest.json` and its additional properties, but got nothing.

packages/cli/src/lib/Cli.ts
toastal: Well, since there aren't named arguments, and they were cases where some were used and some not, I figured a dictionary would be cleaner than `loadOrCreateConfig(undefined, undefined, parsedArgs.config)`, but I don't really care one way or another on the style.

packages/cli/src/lib/config.ts
toastal: Well it _is_ calling out to filesystem which is a slow and unnecessary operation if we are already sure we don't need to do it, right?

toastal: Not to mention you could be renaming a users input file which would be unexpected when rerunning a command.

andreban: Hitting the filesystem shouldn't be such a large overhead in this case and renaming will only happen for a very specific filename, so I'm generally happy with the solution. Nevertheless, those are good points.

It sounds to me `loadOrCreateConfig()` is doing too much. Maybe a better alternative is to rename this function to `loadOrCreateDefaultConfig()` and when `parsedArgs.config` is available just call [`Config.loadConfig()`](https://github.com/GoogleChromeLabs/bubblewrap/blob/master/packages/core/src/lib/Config.ts#L43-L52) instead of this function.

PS: It's also possible to call `Config.deserialize()` to parse a JSON or build a config with a `new Config(jdkPath, androidSdkPath`). Would passing those strings from the command line make it a better solution for your use-case?

andreban:
The best place to check for the `twa-manifest.json` values is [this bit of code](https://github.com/GoogleChromeLabs/bubblewrap/blob/master/packages/core/src/lib/TwaManifest.ts#L480-L521).

packages/cli/src/lib/config.ts
toastal: I don't want to be in charge of function naming.

I think this is a very fast solution that can work for a lot of odd use case as anyone can make a config on the fly but they can also check in and use a local config file (mine is at `project/workspace/.bubblewrap/config.json`). This mean a user could say put the config in `package.json` and use `jq` to read and pass in through process substitution.

I plan to use Dhall everywhere and `dhall-to-json` for my configuration because I want types for my configuration (makes me want my `twa-manifest.json` to be typed too [which I'm working on](https://gitlab.com/toastal/dhall-webmanifest/-/blob/twa/TWAManifest/TrustedWebActivityManifest.dhall)). Also with this solution I can to call `nix-build --no-out-link '<nixpkgs>' -A jdk8_headless` to get the current version of JDK (same for Android SDK) on the fly which currently is at `/nix/store/q839v7q4k8s958zs2smmwm6id72xjsva-openjdk-headless-8u272-b10`. I need reproducible builds ... including all building tools and libraries, so everything needs to be local to `nix-shell` which includes `bubblewrap` and all its dependencies. I may make a derivation if time allows that contains all the bits that make this work on Nix.

andreban (approved):
LGTM, Thank you! Shall I merge this then?

toastal:
Yeah, that be great.

toastal:
How often do you cut releases?

andreban:
We don't have a fixed schedule, due to development not moving at a constant rate. I want to get #463 merged and update the docs with the `--config` option for the next release.

Issue #413, opened 2020-11-30
Add a flag to bubblewrap init to skip key generation/signing


raymeskhoury:
We do key signing ourselves, so it would be good to be able to skip this step when running through `init`.

andreban:
Closed via #473 

@jdecked FYI, and thanks for writing the PR that closed this!

PR #462, opened 2021-02-11
Separate build and sign steps for buildApk and buildAppBundles

andreban:
- Developers reported the need to generate "unsigned" APKs and
  Bundles, as signing packages is sometimes generated as part of
  a CI or by different teams (see #413).
- This PR splits build and signing steps into two methods and paves
  the way for a next PR that will enable passing a parameter to
  skip the signing step.
- We also move constant file names to `consts`.

### DIFF
--- a/packages/cli/src/lib/cmds/build.ts
+++ b/packages/cli/src/lib/cmds/build.ts
@@ -25,6 +25,18 @@ import {printValidationResult} from '../pwaValidationHelper';
 import {ParsedArgs} from 'minimist';
 import {createValidateString} from '../inputHelpers';
 
+// Path to the file generated when building an app bundle file using gradle.
+const APP_BUNDLE_BUILD_OUTPUT_FILE_NAME = 'app/build/outputs/bundle/release/app-release.aab';
+const APP_BUNDLE_SIGNED_FILE_NAME = './app-release-bundle.aab'; // Final signed App Bundle file.
+
+// Path to the file generated when building an APK file using gradle.
+const APK_BUILD_OUTPUT_FILE_NAME = './app/build/outputs/apk/release/app-release-unsigned.apk';
+const APK_SIGNED_FILE_NAME = './app-release-signed.apk'; // Final aligned and signed APK.
+const APK_ALIGNED_FILE_NAME = './app-release-unsigned-aligned.apk'; // Output file for zipalign.
+
+const TWA_MANIFEST_FILE_NAME = './twa-manifest.json';
+const ASSETLINKS_OUTPUT_FILE = './assetlinks.json';
+
 interface SigningKeyPasswords {
   keystorePassword: string;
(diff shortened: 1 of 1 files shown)


packages/cli/src/lib/cmds/build.ts
```diff
@@ -25,6 +25,18 @@ import {printValidationResult} from '../pwaValidationHelper';
 import {ParsedArgs} from 'minimist';
 import {createValidateString} from '../inputHelpers';
 
+// Path to the file generated when building an app bundle file using gradle.
+const APP_BUNDLE_BUILD_OUTPUT_FILE_NAME = 'app/build/outputs/bundle/release/app-release.aab';
+const APP_BUNDLE_SIGNED_FILE_NAME = './app-release-bundle.aab'; // Final signed App Bundle file.
+
+// Path to the file generated when building an APK file using gradle.
+const APK_BUILD_OUTPUT_FILE_NAME = './app/build/outputs/apk/release/app-release-unsigned.apk';
```
PEConn, line 33: Is there any reason to leave this file around? I could see the user getting a little confused between the three files:

* a: `app-release-unsigned.apk`
* b: `app-release-unsigned-aligned.apk`
* c: `app-release-signed.apk`

I can see the users caring about whether or not the apk is signed, but I don't think anyone (other than us for development purposes) cares about the unaligned one (a).

Could we rename (a) to something else (`app-release-unaligned.apk`?) during the build process, and delete it afterwards. Then we can rename (b) to `app-release-unsigned.apk`, and then the user only sees:

* `app-release-unsigned.apk`
* `app-release-signed.apk`

andreban: I wonder if there's anyone relying on those files as part of a CI process nowadays? This might be a breaking change for them.

Anyway, my goal with this PR was to avoid introducing any behaviour changes and just move code around.

packages/cli/src/lib/cmds/build.ts
```diff
@@ -25,6 +25,18 @@ import {printValidationResult} from '../pwaValidationHelper';
 import {ParsedArgs} from 'minimist';
 import {createValidateString} from '../inputHelpers';
 
+// Path to the file generated when building an app bundle file using gradle.
+const APP_BUNDLE_BUILD_OUTPUT_FILE_NAME = 'app/build/outputs/bundle/release/app-release.aab';
```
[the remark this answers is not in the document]
andreban: Not really - just used the same name we had before. I'll try `./` for consistency and check if we get the proper behaviour.

packages/cli/src/lib/cmds/build.ts
PEConn: Fair enough - happy to not change things unnecessarily.

PEConn (approved):

PR #472, opened 2021-02-25
Add --manifest option for build script

jdecked:
**Problem:** The directory with our `twa-manifest.json` and the directory we're running the Bubblewrap commands from are different.

**Solution:** Add a `--manifest` option to the `bubblewrap build` command which allows the user to specify a directory where the client should look for `twa-manifest.json`, much like that option for the `bubblewrap update` command.


andreban (approved):
LGTM, Thank you very much for this!

PR #473, opened 2021-02-25
Add --skipSigning option to build command

jdecked:
**Problem:** In our use case, we have to upload our unsigned APK/App Bundle to another service to be signed by people with the correct internal permissions for signing Android apps.

**Solution:** Add a `--skipSigning` option to the `bubblewrap build` command so that we can get an unsigned, built APK/App Bundle as output with a successful exit code.


andreban (approved):
LGTM, Thank you!

Issue #486, opened 2021-03-17
v1.11.0 assetlinks.json not generated by`bubblewrap build` and `bubblewrap fingerprint generateAssetLinks` is empty


elgreg:
**Describe the bug**

Per the [Quick Start Guide](https://developer.chrome.com/docs/android/trusted-web-activity/quick-start/), an assetlinks.json should be generated when I run `bubblewrap build` but this seems to only be the case in v1.10.0 AND when I use the new fingerprint command, the file is empty (I think because it imports and maps over the empty `fingerprints` key in the generated twa-manifest.json).

**To Reproduce**

1. npm install @bubblewrap/cli@v1.11.1
2. `npx bubblewrap init --manifest=https://slate.com/media/sites/slate-com/web-app-manifest.json`
3. press enter at all prompts
4. `npx bubblewrap build --skipPwaValidation`
5. Note that there is not assetlinks.json file generated nor a message about it
6. `npx bubblewrap fingerprint generateAssetLinks`
7. Note that assetlinks.json is empty

**Expected behavior**
`npx bubblewrap fingerprint generateAssetLinks` should generate an assetlinks.json file with content similar to v1.10.0 e.g.

```
[{
        "relation": ["delegate_permission/common.handle_all_urls"],
        "target" : { "namespace": "android_app", "package_name": "com.slate.twa",
                     "sha256_cert_fingerprints": ["B8:C8:22:AF:6C:D2:5B:7A:68:F4:2D:F8:BB:D8:F5:10:21:21:E4:69:E2:5A:FA:79:0C:EA:2D:C2:69:4D:FF:19"] }
      }
]
```

**Additional context**
When I install v1.10.0, its `bubblewrap build` generates the assetlinks.json file, so this error seems likely related to addition of the fingerprint command in #478. Note also that #478 assume that there will be something in the fingerprints array of the twa-manifest.json file, but that when I do a `bubblewrap init` for my manifest that array is empty in twa-manifest.json. Both v1.10.0 and v1.11.1 generate the empty fingerprints array `"fingerprints": [],` - but only v1.11.1 seems to rely on it being populated with the fingerprint command. Did I miss something?

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser n/a
 - Version v.1.11.1

andreban:
Thanks for filing this, @elgreg. 

[...]

PR #487, opened 2021-03-17
Update help.ts for fingerprint generateAssetLinks

elgreg:
While investigating #486, I noticed that help text was missing the fingerprint command.


andreban (approved):
LGTM, Thank you for the fix.

PR #489, opened 2021-03-23
Allow skip version upgrade and setting version name on merge

anikarni:
In our current flow for updating the TWA, we ran into the issue that we couldn't run `bubblewrap merge && bubblewrap update` in a way that would allow us to increment the version (only once) and change the version name. `bubblewrap update` allows us to do both things, but `bubblewrap merge` doesn't allow us to name the version or to skip upgrading. So, I'm suggesting replicating the logic for `update` in `merge`.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -211,6 +211,8 @@ bubblewrap merge --ignore [fields-list]
 ```
 
 Options:
+  - `--appVersionName`: version name to be used on on the upgrade. Ignored if `--skipVersionUpgrade` is used.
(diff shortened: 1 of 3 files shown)


andreban:
@anikarni thanks, the pull-request looks good to me. Would you mind confirming if you signed the CLA, since the bot added the message, but the check is passing?

anikarni:
@andreban , thanks for the review. Yes, I amended the commit to use the company email, which is why it passed. Hope that clarifies!!

andreban (approved):
Thanks for the PR! LGTM.

PR #491, opened 2021-03-30
Install the GPP plugin into the template project.

nohe427:
This installs the GPP plugin into the template project. This will apply
the configuration files serviceAccountJsonFile key to the specified
files which will activate the GPP plugin.

### DIFF
--- a/packages/core/src/lib/TwaManifest.ts
+++ b/packages/core/src/lib/TwaManifest.ts
@@ -120,7 +120,7 @@ type alphaDependencies = {
  * splashScreenFadeOutDuration: 300
  * isChromeOSOnly: false, // Setting to true will enable a feature that prevents non-ChromeOS devices
  *  from installing the app.
- * serviceAccountJson: '<%= serviceAccountJson %>', //The service account used to communicate with
+ * serviceAccountJson: '<%= serviceAccountJson %>', // The service account used to communicate with
  *  Google Play.
  *
  */
--- a/packages/core/src/lib/TwaManifest.ts
+++ b/packages/core/src/lib/TwaManifest.ts
@@ -156,7 +156,7 @@ export class TwaManifest {
   shareTarget?: ShareTarget;
   orientation: Orientation;
   fingerprints: Fingerprint[];
-  serviceAccountJson: string | undefined;
+  serviceAccountJsonFile: string | undefined;
 
   private static log = new ConsoleLog('twa-manifest');
 
@@ -200,7 +200,7 @@ export class TwaManifest {
     this.shareTarget = data.shareTarget;
     this.orientation = data.orientation || DEFAULT_ORIENTATION;
     this.fingerprints = data.fingerprints || [];
-    this.serviceAccountJson = data.serviceAccountJson;
+    this.serviceAccountJsonFile = data.serviceAccountJsonFile;
   }
 
   /**
@@ -524,7 +524,7 @@ export interface TwaManifestJson {
   shareTarget?: ShareTarget;
   orientation?: Orientation;
   fingerprints?: Fingerprint[];
-  serviceAccountJson?: string;
+  serviceAccountJsonFile?: string;
 }
 
 export interface SigningKeyInfo {
--- a/packages/core/src/lib/TwaManifest.ts
+++ b/packages/core/src/lib/TwaManifest.ts
@@ -120,6 +120,8 @@ type alphaDependencies = {
  * splashScreenFadeOutDuration: 300
  * isChromeOSOnly: false, // Setting to true will enable a feature that prevents non-ChromeOS devices
  *  from installing the app.
+ * serviceAccountJson: '<%= serviceAccountJson %>', //The service account used to communicate with
+ *  Google Play.
  *
  */
 export class TwaManifest {
@@ -154,6 +156,7 @@ export class TwaManifest {
   shareTarget?: ShareTarget;
   orientation: Orientation;
   fingerprints: Fingerprint[];
+  serviceAccountJson: string | undefined;
 
   private static log = new ConsoleLog('twa-manifest');
 
@@ -197,6 +200,7 @@ export class TwaManifest {
     this.shareTarget = data.shareTarget;
     this.orientation = data.orientation || DEFAULT_ORIENTATION;
     this.fingerprints = data.fingerprints || [];
+    this.serviceAccountJson = data.serviceAccountJson;
   }
 
   /**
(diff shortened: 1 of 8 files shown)


nohe427:
There is an issue with doing zipalign when calling `bubblewrap build` after installing the GPP plugin. Some more investigation is needed before this is merged, but if anyone wants to give this a quick once over, they are more than welcome to 😄

nohe427:
Just tested with an updated gradle. `zipalign` may be broken on the newer gradle. More investigation required.

nohe427:
Fixed my zipalign issues. See the comment in 28f53cd

> This changes the zip align command from aligning the file to checking
alignment only. This is based off the information provided in :
https://b.corp.google.com/issues/162117652
and
microsoft/azure-pipelines-tasks#13863
The Android Gradle Plugin automatically aligns APKs now and is no longer
needed to align APKs (based on this comment):
https://b.corp.google.com/issues/162117652#comment12

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
```diff
@@ -141,11 +141,11 @@ export class AndroidSdkTools {
     const env = this.getEnv();
     const zipalignCmd = [
       `"${this.pathJoin(this.getAndroidHome(), `/build-tools/${BUILD_TOOLS_VERSION}/zipalign`)}"`,
-      '-v -f -p 4',
+      '-c -v -p 4',
```
PEConn, line 144: Could you add some documentation for what these flags are and why we need them?

PEConn (commented):
So I don't actually know what you're doing with this PR - maybe Andre does though? Regardless, two small comments.

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
andreban: per documentation, `-c` checks of a file was zipaligned correctly. I don't think that's what we want here. The `-f` flag we had there was to overwrite the output file, if it already existed.

If we need a command to check alignment, I'd rather have a separate function so we keep backward compatibility with other tools using the core library (eg: PWA Builder)

packages/core/src/lib/TwaManifest.ts
```diff
@@ -154,6 +156,7 @@ export class TwaManifest {
   shareTarget?: ShareTarget;
   orientation: Orientation;
   fingerprints: Fingerprint[];
+  serviceAccountJson: string | undefined;
```
andreban, line 159: Do we expect a JSON string or is this a reference to a JSON file?

andreban: Looking at tests, it seems it's a reference to the JSON file. So, maybe we could call the field `serviceAccountJsonFile`?

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
andreban: Am I understanding correctly that, when using `com.github.triplet.play` Gradle will zipalign the file?

packages/core/src/lib/TwaManifest.ts
```diff
@@ -120,6 +120,8 @@ type alphaDependencies = {
  * splashScreenFadeOutDuration: 300
  * isChromeOSOnly: false, // Setting to true will enable a feature that prevents non-ChromeOS devices
  *  from installing the app.
+ * serviceAccountJson: '<%= serviceAccountJson %>', //The service account used to communicate with
```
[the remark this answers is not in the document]
nohe427: Done.

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
nohe427: So `com.github.triplet.play` will not zipalign the file, but the updates to the gradle version (which `com.github.triplet.play` requires) does automatically zipalign the file. The `-f` flag I believe is no longer required at this point and instead a `fs.copyFile ` change actually does the part that overwrites and copies the file after validation.

nohe427: @PEConn (https://developer.android.com/studio/command-line/zipalign) Did you want me to copy the flag definitions into the file as a comment?

packages/core/src/lib/TwaManifest.ts
nohe427: Done

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
PEConn: Verbose is probably self explanatory - explaining why we need `c` would probably be helpful though.

packages/core/template_project/app/build.gradle
```diff
@@ -158,6 +163,12 @@ android {
     }
 }
 
+<% if (serviceAccountJson) { %>
+play {
+    serviceAccountCredentials.set(file("<%= serviceAccountJson %>"))
```
andreban, line 168: I think this should read `serviceAccountJsonFile` now?
```suggestion
    serviceAccountCredentials.set(file("<%= serviceAccountJsonFile %>"))
```

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
andreban: @nohe427 I don't see a behaviour change regarding zipalign between versions 4.0.0 and 4.1.0 of the gradle plugin (https://developer.android.com/studio/releases/gradle-plugin#behavior_changes). I think this behaviour has been there [since 2.2.0](https://developer.android.com/studio/releases/gradle-plugin#2-2-0)? Is this changed needed for this PR to work?

If that's the case, the old zipalign call doesn't break anything, but also doesn't have any effect and should remove the `zipalign` call from the `build` command. If we want to double check if the file is properly aligned, let's add another function that verifies the alignment.

packages/core/template_project/app/build.gradle
nohe427: Yes it should

nohe427: There were a few other places that it looks like I missed with this change.

packages/core/src/lib/androidSdk/AndroidSdkTools.ts
nohe427: @PEConn - I added info about all the flags (no flag left behind!) + a documentation link. My reasoning is that it might be worthwhile for folks unfamiliar with zipalign to see the docs themselves without searching too much. LMK if I should trim it down.

nohe427: @andreban - I added a command called zipalignOnlyVerification. I think this is what you were interested in having instead?

andreban (approved):
LGTM, Thank you!

nohe427:
Thanks for the review(s)!

Issue #488, opened 2021-03-23
Issues with "splash screen" on android app built by PWABuilder (using bubblewrap)


getify:
I've built my android app using [PWA Builder](https://pwabuilder.com), which as I understand uses bubblewrap to produce its Android app exports. I successfully deployed my app to the play store. I originally filed this bug there, but they redirected me to file here.

When I launch the app from my phone, I see a few weird visual issues during the splash screen and loading. These same issues aren't really as pronounced when launching the installed-PWA on the same phone, so I think these are specifically related to how the Android app is generated.

----

The first issue I have is there's inconsistency (a sort of super quick flashing) of the colorization applied to the top status bar during load. See these 2 screenshots, which go in order.

In this one, the top status bar is green per my configuration:

<img width="300" src="https://user-images.githubusercontent.com/150330/112035892-0420e600-8b0e-11eb-986a-9e08081bef16.jpg">

In this one, which happens a split second later, but still while loading and showing my splash screen, the top bar is now black (my phone's dark mode setting overriding, maybe?):

<img width="300" src="https://user-images.githubusercontent.com/150330/112035893-04b97c80-8b0e-11eb-8772-2b0188ee0f64.jpg">

Any idea why this happens? Any way to fix it? I would *prefer* my status bar not be overridden and to stay the same green color I configured, and as it starts out with. Is this likely something specific with my phone/settings, or is this something that would affect most or all other Android phones?

----

Relatedly, the "splash" color being a single flat color is something I wish I could override. Is there anyway to specify a gradient instead of a single color?

As you can see in this screenshot, once my app fully loads, it's using a gradient background, which is intentionally designed so as to match up to the same "green" color on top (if the status bar stayed the intended color) and the same "gray" color I'm specifying for my bottom navigation bar:

<img width="300" src="https://user-images.githubusercontent.com/150330/112035894-04b97c80-8b0e-11eb-8363-8b7a95ce2162.jpg">

In the first two screenshots, that bottom navigation bar sticks out more like a sore thumb because of lack of ability to use my gradient in my splash screen.

I guess this is a little more of a feature request than bug report, but is it possible for bubblewrap to support gradient background on the splash screen? If so, that would eliminate one of the bugs/quirks I'm experiencing. I feel like other native Android apps I've seen (not PWAs) do have more customized splash/loading screens.

----

[...]

PR #494, opened 2021-04-07
Fixes artifacts when upscaling images for the Splash Screen

andreban:
[UPDATED]

Upscaling images using Jimp can generate artifacts when transparent pixels are blended with opaque pixels.  See #488 for more details on the issue.

This happens because, even though the alpha channel sets the pixel to fully transparent, those pixels still have an RGB part to their color. 

As example, if a transparent pixel is `#000000FF` is a transparent pixel that has a black RGB part,  merging with an opaque white like`#FFFFFF00` will result in a semi-transparent grey (eg: `#7F7F7F7F`). However, what we want in this case is for the color to be a semi-transparent white (`#FFFFFF7F`).

Since we know the background color the splash icon is going to be rendered against we can assign the RGB part of the background color to all fully transparent pixels, so that the opaque pixels will be blended against the background color, avoiding artifacts on the border of transparent regions of the image.

This is what an image upscaled with this workaround looks like:
splash

### DIFF
--- a/packages/core/src/lib/TwaGenerator.ts
+++ b/packages/core/src/lib/TwaGenerator.ts
@@ -16,6 +16,7 @@
 
 import * as path from 'path';
 import * as fs from 'fs';
+import * as Color from 'color';
 import fetch from 'node-fetch';
 import {template} from 'lodash';
 import {promisify} from 'util';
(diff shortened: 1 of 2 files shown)


andreban:
Leaving this as a draft for now. I'm still not convinced this is the way to go, as in some cases the output image will be of a much lower quality. Example:

Original:
shortcut_1

Resized:
shortcut_1

packages/core/src/lib/TwaGenerator.ts
```diff
@@ -69,17 +70,20 @@ const DELETE_FILE_LIST = [
   'app/src/main/res/mipmap-anydpi-v26/ic_launcher.xml',
 ];
 
+const SPLASH_IMAGES: IconDefinition[] = [
+  {dest: 'app/src/main/res/drawable-hdpi/splash.png', size: 450},
```
glennhartmann, line 74: currently these are sorted by |dest| alphabetically. IMHO, it would make more sense to sort by |size|.

andreban: Good point. Changed.

PEConn (approved):

PR #495, opened 2021-04-07
Create a Play library to communicate with GPP.

nohe427:
This is the foundation of the Google Play Publisher integration I want. Making some large changes with regards to adding a GooglePlay.ts file and changing the GradleWrapper to accept user defined args.

### DIFF
--- a/packages/core/src/lib/GooglePlay.ts
+++ b/packages/core/src/lib/GooglePlay.ts
@@ -57,7 +57,7 @@ export class GooglePlay {
     // selected track at 100% rollout
     if (track !== Track.Internal) {
       this.gradleWrapper.executeGradleCommand(
-          ['promoteArtifact', '--from-track', 'internal', '--promote-track', Track[track],
+          ['promoteArtifact', '--from-track', 'internal', '--promote-track', Track[track].toLowerCase(),
             '--release-status', 'completed']);
     }
   }
--- a/packages/core/src/lib/GooglePlay.ts
+++ b/packages/core/src/lib/GooglePlay.ts
@@ -14,9 +14,39 @@
  *  limitations under the License.
  */
 
-export interface GooglePlay {
-    initPlay(): Promise<void>;
-    publishBundle(track: Track): Promise<void>;
+import {GradleWrapper} from '..';
+
+export class GooglePlay {
+  /**
+   * Constructs a Google Play object with the gradleWrapper so we can use a
+   *   gradle plugin to communicate with Google Play.
+   *
+   * @param gradleWrapper This is the gradle wrapper object that supplies
+   *   hooks into Gradle.
+   */
+  constructor(private gradleWrapper: GradleWrapper) {}
+
+  /**
+   * Initialized Google Play and loads the existing configruation from Google Play.
+   * The resulting files are stored in the play folder in the src directory.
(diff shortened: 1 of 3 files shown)


nohe427:
@andreban @PEConn - Making some changes to the Play integration and the gradle wrapper and wanted to gather some early thoughts before I continue.

packages/core/src/lib/GooglePlay.ts
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2019 Google Inc. All Rights Reserved.
```
PEConn, line 2: nit: 2021

packages/core/src/lib/GradleWrapper.ts
```diff
@@ -62,4 +62,14 @@ export class GradleWrapper {
     await executeFile(
         this.gradleCmd, ['assembleRelease', '--stacktrace'], env, undefined, this.projectLocation);
   }
+
+  /**
+   * Executes gradle commands with custom arguments.
+   * @param args - Arguments supplied to gradle, also considered gradle tasks.
+   */
+  async executeGradleCommand(args: string[]): Promise<void> {
```
PEConn, line 70: We can now implement the above two methods (maybe more as well) just using this one.

Or maybe we wanted to keep all the raw gradle commands in this file (to make them easier to audit maybe), in which case we should make a `bootstrap` method or something.

nohe427: > We can now implement the above two methods (maybe more as well) just using this one.

Yup, that was my intention moving forward!

> Or maybe we wanted to keep all the raw gradle commands in this file (to make them easier to audit maybe), in which case we should make a bootstrap method or something.

I was torn between putting the bootstrap command in here or putting in its own Play.ts file. The reasoning behind the Play.ts file was that if we do decide to move away from the plugin in the future, it may make the API surface a bit more stable if we still have the Play.ts file hanging around that we can call. Might also make it a little more stable if we just make a Play interface with which the GradleWrapper can implement the methods and if we do decide to move away from the Gradle plugin, we can just implement the interface on another class?

Thoughts?

packages/core/src/lib/GooglePlay.ts
nohe427: Fixed.

PEConn:
This is looking good - are you planning on building on this change or landing this as a small part of a string of PRs?

andreban:
One thought here is that I'd probably try favouring composition over inheritance. Something like:

```typescript
export class GooglePlay {
    constructor(private gradleWrapper: GradleWrapper) {}

    async publishBundle(track: Track): Promise<void> {
        this.gradleWrapper.executeGradleCommand(...);
    }

    async initPlay(track: Track): Promise<void> {
        this.gradleWrapper.executeGradleCommand(...);
    }
}
```

so, we'd use with:
```typescript
const googlePlay = new GooglePlay(gradleWrapper);
googlePlay.initPlay(...);
googlePlay.publishBundler(...);
```

nohe427:
Ok, I can switch this to composition over inheritance.

nohe427:
@PEConn & @andreban - Sorry for the million commits, but I think I got this to a re-viewable state. Please let me know.

andreban (approved):
LGTM!

PEConn (approved):

PR #497, opened 2021-04-14
Add --signingKeyPath and --signingKeyAlias options to build command

### LINKED ISSUE #490: Support for providing signing key path & alias via command line flag in build command
I'm generating a non-production keystore to sign the APK with for testing purposes only. I don't want to put the non-production path and alias of this signing key into my TWA manifest because it isn't actually what's used to sign the production release of the APK, so it might confuse future developers. It would be great if Bubblewrap could use a keystore path and alias that aren't supplied in the TWA manifest to sign the APK during the build command.

Accept `--signingKeyPath` and `--signingKeyAlias` command line flags for the build command, which take precedence during the build command over what's specified in the TWA manifest.

**Key path and alias provided via environment variables:** Neither of these are particularly sensitive data, so environment variables don't seem necessary, but they could work, too.
**Putting the keystore path and alias into the TWA manifest:** Given that this keystore is generated on the fly for testing purposes, it may confuse future developers to put it in the TWA manifest when the actual (production) signing process involves working with another team to get the APK signed & released to prod.

**Additional context**
Happy to implement this myself if that'd be helpful!

jdecked:
Fixes #490.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -1,13 +1,13 @@
 <!---
 
   Copyright 2019 Google Inc. All Rights Reserved.
- 
+
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
    You may obtain a copy of the License at
- 
+
        http://www.apache.org/licenses/LICENSE-2.0
- 
+
    Unless required by applicable law or agreed to in writing, software
    distributed under the License is distributed on an "AS IS" BASIS,
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


packages/cli/README.md
```diff
@@ -113,12 +113,14 @@ Set `BUBBLEWRAP_KEYSTORE_PASSWORD` for the key store password and `BUBBLEWRAP_KE
...
 ```
-bubblewrap build [--skipPwaValidation] [--skipSigning] [--manifest="<path-twa-manifest>"]
+bubblewrap build [--skipPwaValidation] [--skipSigning] [--manifest="<path-twa-manifest>"] [--signingKeyPath="<path-to-signing-key>"] [--signingKeyAlias="<key-alias>"]
 ```
 
 Options: 
   - `--skipPwaValidation`: skips validating the wrapped PWA against the Quality Criteria.
-  - `--skipSigning`: skips signing the built APK and App Bundle.
+  - `--skipSigning`: skips signing the built APK and App Bundle. Incompatible with `--signingKeyPath` and `--signingKeyAlias`.
+  - `--signingKeyPath`: path to keystore to use for signing the built APK and App Bundle. Overrides signingKey.path specified in `twa-manifest.json`. Incompatible with `--skipSigning`.
```
PEConn, line 122: I think potentially we're overcommunicating here. I can't think of an example where a command line parameter to a tool is overridden by the value in the config file - the parameter always takes precedence. Also, personally I'd take the "incompatible with `--skipSigning` as implied.

I'd be fine with just:

`--signingKeyPath`: path to keystore to use for signing the built APK and App Bundle.

And assume the user can figure out the other two parts.

PEConn (approved):
Code looks good - got a comment on the README docs, but that's more of a loose opinion. Let's see what Andre thinks.

packages/cli/README.md
andreban: What I understand is that when `--skipSigning` is used, `--signingKeyPath` and `--signingKeyAlias` are ignored. I think incompatible makes it look like command will not work if those are used together, which is not the case? 

I agree that the user can probably figure out implications of `--skipSigning` , but it doesn't hurt to be explicit about it.  I'd just say other parameters related to signing are ignored when this is used. I wouldn't add extra information on this on `--signingKeyPath` and only have it on `--skipSigning`.

andreban (approved):
LGTM, with nit on incompatible vs ignored.

jdecked:
Could one of y'all please do a release once this is merged? Waiting for the release to merge an internal branch at Twitter. 🙏

andreban:
> Could one of y'all please do a release once this is merged? Waiting for the release to merge an internal branch at Twitter. 🙏

I'm planning a release on Tue (I'm OOO today and Monday)

PR #505, opened 2021-04-20
Ensure replacement color is positive

andreban:


nohe427 (approved):

packages/core/src/lib/ImageHelper.ts
```diff
@@ -47,7 +47,7 @@ export class ImageHelper {
     // See https://github.com/GoogleChromeLabs/bubblewrap/issues/488#issuecomment-806560923.
     if (backgroundColor && image.hasAlpha()) {
       // The replacement colour is the same as the background colour, but fully transparent.
-      const replacementColor = (backgroundColor.rgbNumber() << 8) & 0xFFFFFF00;
+      const replacementColor = ((backgroundColor.rgbNumber() << 8) & 0xFFFFFF00) >>> 0;
```
PEConn, line 50: This reads like magic to me (combined with there being no commit message). I'm assuming it's something to do with promoting the integer type?

andreban: TL;DR - bitten by the lack of unsigned integers :D. 

- `0xFFFFFF00` is `4294967040`, right?
- However, if you do `0xFFFFFF << 8`, you get `-256`.
- So, we force it to be interpreted as a positive int with `(0xFFFFFF << 8) >>> 0`

PEConn: What is the data type before and after the `>>> 0`?

andreban: `number`...

andreban: Adding more detailed comments on the next PR

PR #508, opened 2021-04-23
Add play command to the Bubblewrap CLI.

nohe427:


nohe427:
This is still in a draft state. Uploading for early feedback discussion on some points later on.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (this.args.appBundleLocation) {
+      // Check this is a directory that contains our specified file name.
+    }
+
+    
+      // Make tmp directory copy file over signed APK then cleanup.
+      // await this.googlePlay.publishBundle(Track.Internal,)
+  }
+
+  private validServiceAccountJsonFile(path: string | undefined): Boolean { // Return an error or boolean? Log a message?
```
PEConn, line 69: Can we use `string?` instead of `string | undefined`? (I can never remember this myself.)

packages/cli/src/lib/cmds/play.ts
```diff
@@ -0,0 +1,123 @@
...
+      // Add service account to the TWA-Manifest
+      const serviceAccountFile = this.args.serviceAccount;
+      twaManifest.serviceAccountJsonFile = serviceAccountFile;
+      // Then we need to call bubblewrap update so the gradle plugin has the appropriate file.
+    }
+    // Check if service account file exists on disk.
+
+    // Need to validate that the service account file exists in TWA-Manifest
+    // and/or on disk. (Thinking about CI/CD scenarios)
+    if (this.validServiceAccountJsonFile(twaManifest.serviceAccountJsonFile)) {
```
PEConn, line 97: I think there should be a "not" in this line.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+export async function play(config: Config, parsedArgs: ParsedArgs,
+    log: Log = new ConsoleLog('play'), prompt: Prompt = new InquirerPrompt()): Promise<boolean> {
+  const jdkHelper = new JdkHelper(process, config);
+  const androidSdkTools = await AndroidSdkTools.create(process, config, jdkHelper, log);
+  const gradleWrapper = new GradleWrapper(process, androidSdkTools);
+  const googlePlay = new GooglePlay(gradleWrapper);
+  const play = new Play(parsedArgs, config, androidSdkTools, googlePlay);
+  play.run();
```
PEConn, line 121: Do you not need an `await` here?

PEConn (commented):
Left a bunch of nits (I know it's unfair of me since this is a WIP). Overall structure looks good.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -0,0 +1,123 @@
...
+    await this.googlePlay.initPlay();
+  }
+
+  async getLargestVersion(): Promise<void> { 
+    // TODO(nohe427): This doesn't exist in the Gradle Play Plugin. Might be wortwhile to build
+      // small integration to do this by getting the service account file?
+  }
+
+  private isInAvailableTracks(userSpecifiedTrack: string): Boolean {
+    const track: string = (userSpecifiedTrack).toLowerCase();
```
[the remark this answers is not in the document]
nohe427: I dont. Removed.

packages/cli/src/lib/cmds/play.ts
nohe427: I do not believe so. Tested and it made the compiler unhappy

packages/cli/src/lib/cmds/play.ts
nohe427: Correct. Added that in

packages/cli/src/lib/cmds/play.ts
nohe427: Yup. Added

packages/cli/src/lib/cmds/play.ts
```diff
@@ -0,0 +1,123 @@
...
+    const selectedTrack: Track = (<any>Track)[track];
```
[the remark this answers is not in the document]
nohe427: This allows me to check that the string track is able to be matched from the Track enum.

packages/core/src/lib/GooglePlay.ts
```diff
@@ -51,14 +51,6 @@ export class GooglePlay {
...
-        ['publishBundle', '--artifact-dir', filepath]);
-
-    // Uses the promote function to promote from the internal track (default upload) to the user
-    // selected track at 100% rollout
-    if (track !== Track.Internal) {
-      this.gradleWrapper.executeGradleCommand(
-          ['promoteArtifact', '--from-track', 'internal', '--promote-track',
-            Track[track].toLowerCase(), '--release-status', 'completed']);
-    }
+        ['publishBundle', '--artifact-dir', filepath, '--track', Track[track].toLowerCase()]);
```
PEConn, line 54: Do we need the `toLowerCase` anymore? The enums are all lower case now.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -0,0 +1,131 @@
...
+import {
+  Config, GradleWrapper, JdkHelper, AndroidSdkTools, ConsoleLog, Log, GooglePlay, TwaManifest,
+} from '@bubblewrap/core';
+import {Track} from '@bubblewrap/core/dist/lib/GooglePlay';
+import * as fs from 'fs';
+import * as path from 'path';
+import {ParsedArgs} from 'minimist';
+import {TWA_MANIFEST_FILE_NAME} from '../constants';
+import {Prompt, InquirerPrompt} from '../Prompt';
+/**
```
PEConn, line 26: nit: Empty line between the imports and the comments.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  async bootstrapPlay(): Promise<void> {
+    await this.googlePlay.initPlay();
+  }
+
+  // bubblewrap play --versionCheck can validate the largest version number vs twa-manifest.json and update to give x+1 version number.
+  async getLargestVersion(): Promise<Number> {
+    // Need to get an editId, then list all apks available. This should allow us to query the highest apk number.
+    // This exists in Gradle play plugin but is not easily accessible over CLI.
+    // This should be completed in a future CL.
```
PEConn, line 45: Could you do a `throw new Error("Not implemented);` here?

packages/cli/src/lib/cmds/play.ts
```diff
@@ -0,0 +1,131 @@
...
+    return 0;
+  }
+
+  private isInAvailableTracks(userSpecifiedTrack: string): boolean {
+    const track: string = userSpecifiedTrack.toLowerCase();
+    const selectedTrack: Track = (Track as any)[track];
```
PEConn, line 51: So I think you could change this to `Track[track as keyof typeof Track]` (see the comments here [1]), which at least to me makes things seem a little clearer.

[1]: https://stackoverflow.com/a/17381004

packages/cli/src/lib/cmds/play.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+    // Make tmp directory copy file over signed APK then cleanup.
+    const publishDir = fs.mkdtempSync('bubblewrap');
+    fs.copyFileSync("DEFAULT_FILE_PATH", path.join(publishDir, "DEFAULT_FILE_NAME")); // Need help on default file path and default file name.
+    await this.googlePlay.publishBundle(Track.internal, publishDir);
+    
+    fs.rmdirSync(publishDir);
+  }
+
+  private validServiceAccountJsonFile(path: string | undefined): boolean { // Return an error or boolean? Log a message?
```
PEConn, line 79: We do have an Error type - https://github.com/GoogleChromeLabs/bubblewrap/blob/main/packages/core/src/lib/Result.ts

packages/cli/src/lib/cmds/play.ts
```diff
@@ -0,0 +1,131 @@
...
+import * as path from 'path';
+import {ParsedArgs} from 'minimist';
+import {TWA_MANIFEST_FILE_NAME} from '../constants';
+import {Prompt, InquirerPrompt} from '../Prompt';
+/**
+  * The Play class is the class that is used to communicate with the Google Play Store.
+  */
+class Play {
+  constructor(
+    private args: ParsedArgs,
```
andreban, line 31: We haven't been very consistent about this, but we may want to introduce a `PlayArgs` that helps us with type checking. We [have one for init](https://github.com/GoogleChromeLabs/bubblewrap/blob/main/packages/cli/src/lib/cmds/init.ts#L29-L34) and we [cast `ParsedArgs` to that type](https://github.com/GoogleChromeLabs/bubblewrap/blob/main/packages/cli/src/lib/Cli.ts#L70) when invoking the command.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -0,0 +1,131 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {
+  Config, GradleWrapper, JdkHelper, AndroidSdkTools, ConsoleLog, Log, GooglePlay, TwaManifest,
+} from '@bubblewrap/core';
+import {Track} from '@bubblewrap/core/dist/lib/GooglePlay';
```
andreban, line 20: This import doesn't look right. We probably want something like:

`import {Track} from `@bubblewrap/core`.

You'll probably need to modify `index.ts` in core for this.

packages/cli/src/lib/cmds/play.ts
andreban: We could implement `Track` [similar to how we implement `Orientation`](https://github.com/GoogleChromeLabs/bubblewrap/blob/main/packages/core/src/lib/TwaManifest.ts#L49-L61) and get rid of this cast.

andreban (commented):
Added some notes (I'm aware it's a WIP too)

packages/core/src/lib/GooglePlay.ts
nohe427: You're right.

packages/cli/src/lib/cmds/play.ts
nohe427: Done

packages/cli/src/lib/cmds/play.ts
nohe427: Done. Thanks for that!

packages/cli/src/lib/cmds/play.ts
nohe427: Done. Good Catch!

packages/cli/src/lib/cmds/play.ts
nohe427: I am starting down this path.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -39,40 +40,31 @@ class Play {
...
   }
 
   // bubblewrap play --publish="Internal"
   async publish(): Promise<void> {
     // if (this.args.publish) { // This argument should be validated in run()
     // }
     // Validate that the publish value is listed in the available Tracks.
-    const userSelectedTrack = this.args.publish.toLowerCase() || 'internal'; // If no value was supplied with publish we make it internal.
-    if (!this.isInAvailableTracks(userSelectedTrack)) {
+    const userSelectedTrack = asPlayStoreTrack(this.args.publish.toLowerCase() || 'internal'); // If no value was supplied with publish we make it internal.
```
PEConn, line 55: nit: Can you break the long lines up please?

packages/core/src/lib/GooglePlay.ts
```diff
@@ -16,11 +16,16 @@
...
+// Possible values for release tracks
+const TRACK_VALUES = ['alpha', 'beta', 'internal', 'production'];
+export type PlayStoreTrack = typeof TRACK_VALUES[number];
+export const PlayStoreTracks: PlayStoreTrack[] = [...TRACK_VALUES];
+
+export function asPlayStoreTrack(input?: string): PlayStoreTrack | null {
+  if(!input) {
+    return null;
+  }
+  return TRACK_VALUES.includes(input) ? input as PlayStoreTrack : null;
```
PEConn, line 28: This is much more readable, thanks!

PEConn (approved):
LGTM with nit.

packages/cli/src/lib/cmds/play.ts
nohe427: Think I got this one done.

packages/cli/src/lib/cmds/play.ts
nohe427: I think I got this resolved.

packages/cli/src/lib/cmds/play.ts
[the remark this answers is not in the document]
nohe427: This is resolved now as well i think

nohe427:
@PEConn & @andreban 

I think I got this to a pretty good state thus far (enough for an inital merge anyways).

Would you all take a peek and try kicking the tires on it as well?

PEConn:
LGTM

PR #510, opened 2021-04-26
Update twa-manifest.json after successful project generation

joycetoh8:
- Move writing new version code and version name to twa-manifest.json
until after project is successfully generated. 
- Also edited and added messages to remind user to run `bubblewrap build`
after successful `bubblewrap update`.


PEConn:
Could you add a little context on why we're doing this? Was there a bug? Why was the old behaviour sub-optimal?

andreban:
The issue this is fixing is that, if `update` failed to generate a new project, it would still increase the version number and version code. This change prevents this from happening, as it only saves the new `twa-manifest.json` after the project is generated successfully.

andreban (approved):
LGTM, Thank you!

PEConn:
OK, LGTM too.

PR #511, opened 2021-04-27
Additional checks for when Play Billing is enabled

joycetoh8:
- Check that enableNotifications is true
- Check that alphaDependencies are enabled

### DIFF
--- a/packages/cli/src/lib/cmds/build.ts
+++ b/packages/cli/src/lib/cmds/build.ts
@@ -143,6 +143,19 @@ class Build {
     const manifestFile = this.args.manifest || path.join(process.cwd(), TWA_MANIFEST_FILE_NAME);
     const twaManifest = await TwaManifest.fromFile(manifestFile);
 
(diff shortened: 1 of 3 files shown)


andreban (changes requested):
Code looks good. However, it sounds to me that this check belongs in `update`, not `build`. The first takes `twa-manifest.json` and generates the Android project, which is where these configurations matter, and the latter will just build the project.

joycetoh8:
> Code looks good. However, it sounds to me that this check belongs in `update`, not `build`. The first takes `twa-manifest.json` and generates the Android project, which is where these configurations matter, and the latter will just build the project.

Yes, you're absolutely right! Thanks for pointing it out. It'd be too late to point out in `build` that the setup isn't consistent because `update` would've already propagated those changes into the Android project. I've moved the checks to `update` instead. Thanks!

andreban (approved):
LGTM, Thank you!

PR #521, opened 2021-05-06
Adds documentation for fingerprint

andreban:


packages/cli/README.md
```diff
@@ -218,6 +218,60 @@ Options:
...
+Adds a fingerprint to the project configuration.
+
+Usage:
+
+```
+bubblewrap fingerprint add [SHA-256 fingerprint] <flags>
+```
+
+Flags:
+ - `--name=<name>`: he name for the fingerprint.
```
PEConn, line 246: nit: s/he/the/

packages/cli/README.md
```diff
@@ -218,6 +218,60 @@ Options:
...
+
+#### `remove`
+Removes a fingerprint from the project configuration.
+
+Usage:
+```
+bubblewrap fingerprint remove [SHA-256 fingerprint] <flags>
```
PEConn, line 253: For the above command you mention the flags below the example, but you don't do it here.

packages/cli/README.md
PEConn, line 246: Could you add something saying what the `name` does?

packages/cli/README.md
andreban: There are no flags beyond the global for this command.

PEConn: Ah, OK, in which case, can you rename "Global Options" to "Global Flags" above, and rename Flags on line 245 to "Additional Flags"?

packages/cli/README.md
andreban: updated.

andreban: done.

PEConn:
LGTM, thanks

Issue #516, opened 2021-05-03
Bubblewrap fails on sites that use HTTP/2


JudahGabriel:
When generating a TWA for a site that uses HTTP/2, Bubblewrap fails with the following error:

> 021-05-03T18:34:11.907964631Z: [ERROR]  Responded with status 500
2021-05-03T18:34:11.907985832Z: [ERROR]      at TwaGenerator.writeWebManifest (/app/node_modules/@bubblewrap/core/dist/lib/TwaGenerator.js:218:19)
2021-05-03T18:34:11.907993432Z: [ERROR]      at runMicrotasks (<anonymous>)
2021-05-03T18:34:11.907998132Z: [ERROR]      at processTicksAndRejections (internal/process/task_queues.js:97:5)
2021-05-03T18:34:11.908002332Z: [ERROR]      at TwaGenerator.createTwaProject (/app/node_modules/@bubblewrap/core/dist/lib/TwaGenerator.js:344:13)

Run Bubblewrap on a site that serves HTTP/2 content, such as https://jued.ru. This can be done from PWABuilder as well. You'll receive the above error.

I think the underlying issue is that Bubblewrap uses node-fetch, and node-fetch doesn't support HTTP/2. We've had several bugs opened on PWABuilder where the user gets the above error when creating a TWA package. Upon investigation, the issue is the site is using HTTP/2. [Original bug report here](https://github.com/pwa-builder/PWABuilder/issues/1675).

Most of the times we see this error, it's because the developer enabled Cloudflare, which allows HTTP/2 (and HTTP/3!) with the flip of a switch.

andreban:
Hey @JudahGabriel 

In the stacktrace above, we can see `021-05-03T18:34:11.907964631Z: [ERROR] Responded with status 500` - this indicates the server returned an [Internal Server Error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500), not an HTTP/2 error. Unfortunately, it seems the error is on the developers server.

Here's the output of `curl -v https://jued.ru/manifest.json`, which returns a 403 error:

<details>

<summary>
Expand to see the output:
</summary>

```
❯ curl -v https://jued.ru/manifest.json
*   Trying 104.26.11.228...
* TCP_NODELAY set
* Connected to jued.ru (104.26.11.228) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
[...]

PR #522, opened 2021-05-07
Use `fetch-h2` by default with option to fallback to `node-fetch`

andreban:
- `node-fetch` doesn't support HTTP/2 and has had an issue for
  support open since 2017: https://github.com/node-fetch/node-fetch/issues/342
- `fetch-h2` supports HTTP1.1 and HTTP/2 with upgrade.
- Lack of HTTP/2 has been causing issues with developers with server with only HTTP/2 enabled. See #516.
- Changes the code to use `fetch-h2` by default, but allows changing the default from `FetchUtils.setFetchEngine()`.


PEConn (approved):
Great, thanks!

andreban:
I'm thinking about using `node-fetch` by default but having a flag that optionally enables `fetch-h2`. @PEConn WDYT?

PEConn:
What's the advantage of supporting both `node-fetch` and `fetch-h2`? Is it just incase `fetch-h2` has some bugs?

andreban:
Yes. `node-fetch` has 21 million downloads / week and `fetch-h2` around 6k.

PEConn:
In that case, I suggest we use `fetch-h2` by default and allow an option to use `node-fetch`. That way we'll know if there are any bugs, and if there aren't we can remove `node-fetch`.

PEConn (approved):

andreban:
> Aren't we going to make one of them an optional dependency?

I don't think there's a big overhead in having both and this allow deciding which one to use when executing the app. Otherwise, users need to decide at install time.

PEConn:
Gotcha, LGTM

PR #528, opened 2021-05-11
Auto apply twa manifest changes

joycetoh8:
Using checksum to track changes to `twa-manifest.json`, if a user tries to build a outdated version of the project, they will be prompted if they would like to apply the newest changes before building.

### DIFF
--- a/packages/cli/src/lib/cmds/build.ts
+++ b/packages/cli/src/lib/cmds/build.ts
@@ -25,9 +25,8 @@ import {PwaValidator, PwaValidationResult} from '@bubblewrap/validator';
 import {printValidationResult} from '../pwaValidationHelper';
 import {ParsedArgs} from 'minimist';
 import {createValidateString} from '../inputHelpers';
-import {update} from './update';
+import {updateProject} from './shared';
 import {TWA_MANIFEST_FILE_NAME} from '../constants';
-import minimist = require('minimist');
 
 // Path to the file generated when building an app bundle file using gradle.
 const APP_BUNDLE_BUILD_OUTPUT_FILE_NAME = './app/build/outputs/bundle/release/app-release.aab';
@@ -62,22 +61,14 @@ class Build {
    * Checks if the twa-manifest.json file has been changed since the last time the project was generated.
    */
   async checkManifest(manifestFile: string): Promise<boolean> {
-    const prompt = this.prompt;
     const manifestContents = fs.readFileSync(manifestFile);
     const checksumFile = path.join(process.cwd(), 'manifest-checksum.txt');
     const prevChecksum = (await fs.promises.readFile(checksumFile)).toString();
     const currChecksum = crypto.createHash('sha1').update(manifestContents).digest('hex');
     if (currChecksum != prevChecksum) {
-      const applyChanges = await prompt.promptConfirm(messages.promptUpdateProject, true);
-      if (applyChanges) {
-        const updated = await update(minimist([]));
-        if (updated) {
-          prompt.printMessage(messages.messageContinueBuildingApp);
-        }
-        return updated;
-      }
+      return true;
     }
-    return true;
+    return false;
   }
 
   /**
@@ -170,8 +161,16 @@ class Build {
 
     const manifestCheck = await this.checkManifest(manifestFile);
 
-    if (!manifestCheck) {
-      return false;
+    if (manifestCheck) {
+      const applyChanges = await this.prompt.promptConfirm(messages.promptUpdateProject, true);
+      if (applyChanges) {
+        const updated = await updateProject(false, '', this.prompt, '', manifestFile);
+        if (!updated) {
+          return false;
+        }
+      } else {
+        this.prompt.printMessage(messages.messageProjectNotUpdated);
+      }
     }
 
     let passwords = null;
--- a/packages/cli/src/lib/cmds/build.ts
+++ b/packages/cli/src/lib/cmds/build.ts
@@ -65,8 +65,6 @@ class Build {
     const prevChecksum = (await fs.promises.readFile(checksumFile)).toString();
     const manifestContents = await fs.promises.readFile(manifestFile);
     const currChecksum = computeChecksum(manifestContents);
-    this.prompt.printMessage('New checksum found to be: ' + currChecksum);
-    this.prompt.printMessage('Old checksum is: ' + prevChecksum);
     return currChecksum != prevChecksum;
   }
 
@@ -162,7 +160,7 @@ class Build {
     if (hasManifestChanged) {
       const applyChanges = await this.prompt.promptConfirm(messages.promptUpdateProject, true);
       if (applyChanges) {
-        const updated = await updateProject(false, '', this.prompt,
+        const updated = await updateProject(false, null, this.prompt,
             this.args.directory, manifestFile);
         if (!updated) {
           return false;
--- a/packages/cli/src/lib/cmds/build.ts
+++ b/packages/cli/src/lib/cmds/build.ts
@@ -16,6 +16,7 @@
 
 import {AndroidSdkTools, Config, GradleWrapper, JdkHelper, KeyTool, Log,
   ConsoleLog, TwaManifest, JarSigner, SigningKeyInfo, Result} from '@bubblewrap/core';
+import * as crypto from 'crypto';
 import * as fs from 'fs';
 import * as path from 'path';
 import {enUS as messages} from '../strings';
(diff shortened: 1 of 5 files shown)


packages/cli/src/lib/cmds/build.ts
```diff
@@ -55,6 +57,20 @@ class Build {
       private prompt: Prompt = new InquirerPrompt()) {
   }
 
+  /**
+   * Checks if the twa-manifest.json file has been changed since the last time the project was generated.
+   */
+  async checkManifest(manifestFile: string): Promise<boolean> {
```
PEConn, line 63: Can we rename this to make it a bit clearer what the return type signifies? Something like `hasManifestChanged`?

packages/cli/src/lib/cmds/build.ts
```diff
@@ -55,6 +57,20 @@ class Build {
...
+    const manifestContents = fs.readFileSync(manifestFile);
+    const checksumFile = path.join(process.cwd(), 'manifest-checksum.txt');
+    const prevChecksum = (await fs.promises.readFile(checksumFile)).toString();
+    const currChecksum = crypto.createHash('sha1').update(manifestContents).digest('hex');
+    if (currChecksum != prevChecksum) {
+      return true;
+    }
+    return false;
```
PEConn, line 71: These lines can just be `return currChecksum != prevChecksum`.

packages/cli/src/lib/cmds/build.ts
```diff
@@ -143,6 +159,20 @@ class Build {
     const manifestFile = this.args.manifest || path.join(process.cwd(), TWA_MANIFEST_FILE_NAME);
     const twaManifest = await TwaManifest.fromFile(manifestFile);
 
+    const manifestCheck = await this.checkManifest(manifestFile);
```
PEConn, line 162: Similarly, could this variable be `hasManifestChanged`?

packages/cli/src/lib/cmds/shared.ts
```diff
@@ -84,3 +88,58 @@ export async function updateVersions(
     appVersionName: appVersionName,
   };
 }
+
+export async function generateManifestChecksumFile(manifestFile: string,
+    prompt: Prompt): Promise<void> {
+  fs.readFile(manifestFile, async function(err, data) {
```
PEConn, line 94: Could you use the promise version of `readFile` here please?

packages/cli/src/lib/cmds/shared.ts
```diff
@@ -84,3 +88,58 @@ export async function updateVersions(
...
+    if (err) {
+      prompt.printMessage(err.toString());
+      return;
+    }
+    const csFile = path.join(process.cwd(), 'manifest-checksum.txt');
```
PEConn, line 99: Could `csFile` be `checksumFile`?

packages/cli/src/lib/cmds/shared.ts
```diff
@@ -84,3 +88,58 @@ export async function updateVersions(
...
+    const sum = crypto.createHash('sha1').update(data).digest('hex');
```
PEConn, line 100: You've got this line in two places - here and in build.ts line 67, and both lines need to stay consistent. Could you create some helper function (`computeChecksum`?) so we don't risk someone accidentally changing one but not the other?

packages/cli/src/lib/cmds/shared.ts
PEConn: Actually, I think the current version has a bug - `await generateManifestChecksumFile(...)` won't actually wait for the completion of the callback in this promise.

packages/cli/src/lib/cmds/build.ts
```diff
@@ -55,6 +57,20 @@ class Build {
       private prompt: Prompt = new InquirerPrompt()) {
   }
 
+  /**
+   * Checks if the twa-manifest.json file has been changed since the last time the project was generated.
+   */
+  async checkManifest(manifestFile: string): Promise<boolean> {
+    const manifestContents = fs.readFileSync(manifestFile);
+    const checksumFile = path.join(process.cwd(), 'manifest-checksum.txt');
```
PEConn, line 65: We should put the checksum file in the same directory as the other files (`args.directory`).

packages/cli/src/lib/cmds/shared.ts
```diff
@@ -84,3 +88,58 @@ export async function updateVersions(
...
+    prompt: Prompt = new InquirerPrompt(),
+    directory: string,
+    manifest: string): Promise<boolean> {
+  const targetDirectory = directory || process.cwd();
+  const manifestFile = manifest || path.join(process.cwd(), 'twa-manifest.json');
+  const twaManifest = await TwaManifest.fromFile(manifestFile);
+  twaManifest.generatorApp = APP_NAME;
+
+  const features = twaManifest.features;
+  // Check that if Play Billing is enabled, enableNotifications must also be true
```
PEConn, line 117: I know you didn't introduce these comments, but do you mind cleaning them up a bit? (Adding full stops, changing "billing" to "Play Billing" in the one below)

packages/cli/src/lib/cmds/build.ts
```diff
@@ -143,6 +159,20 @@ class Build {
     const manifestFile = this.args.manifest || path.join(process.cwd(), TWA_MANIFEST_FILE_NAME);
     const twaManifest = await TwaManifest.fromFile(manifestFile);
 
+    const manifestCheck = await this.checkManifest(manifestFile);
+
+    if (manifestCheck) {
+      const applyChanges = await this.prompt.promptConfirm(messages.promptUpdateProject, true);
+      if (applyChanges) {
+        const updated = await updateProject(false, '', this.prompt, '', manifestFile);
```
PEConn, line 167: Shouldn't the `appVersionName` parameter here be not null? (Also I'm a bit dubious about the directory being null as well.)

PEConn (commented):
Hey Joyce,

Thanks for the change, sorry there are so many comments!

Peter C

packages/cli/src/lib/cmds/build.ts
joycetoh8: done.

packages/cli/src/lib/cmds/build.ts
joycetoh8: done.

packages/cli/src/lib/cmds/build.ts
joycetoh8: done.

packages/cli/src/lib/cmds/shared.ts
joycetoh8: Yep, you're absolutely right. Changed it to use the promise version.

packages/cli/src/lib/cmds/shared.ts
joycetoh8: done.

packages/cli/src/lib/cmds/shared.ts
joycetoh8: Good call! Added the helper function in.

packages/cli/src/lib/cmds/build.ts
joycetoh8: Changed it to use `args.directory` if it's there. Same thing in other parts of the code. The README should probably be updated to also have the `--directory` flag.

packages/cli/src/lib/cmds/shared.ts
joycetoh8: done.

packages/cli/src/lib/cmds/build.ts
joycetoh8: I changed the directory so it'll use `args.directory`. I'm not sure what to do about `appVersionName` though, because this update is happening when they run `build` so the user won't be passing in `appVersionName`.

joycetoh8:
> Hey Joyce,
> 
> Thanks for the change, sorry there are so many comments!
> 
> Peter C

Thanks for the review, Peter! The more comments the better :) I think I've resolved all except the one about `appVersionName`.

packages/cli/src/lib/cmds/build.ts
PEConn: OK, so it looks like the code will continue to work if `appVersionName` is empty - the `updateVersions` method should deal with it.

I'm not too fond of passing empty strings around - if anything we should make `appVersionName` nullable and pass null around to make it really clear. Also, the documentation for `updateProject` should make it clear that `appVersionName` can be null, and if so a default will be used (if needed).

Do you mind changing appVersion name to a `string?` and adding such documentation to `updateProject`?

joycetoh8: I made `appVersionName` nullable and also updated documentation for `updateProject`. Let me know if this is okay or if it makes more sense to make `appVersionName` an optional parameter instead. Thanks!

andreban (approved):
Looks good to me!

joycetoh8:
@PEConn, I wanted to see if you had anymore thoughts before I merge this? Thanks!

PEConn (approved):

PR #536, opened 2021-05-15
Adds support for additional origins

andreban:
- Adds the `additionalOrigins` field to the `TwaManifest`.
- Adds `res/values/strings.xml`.
- Moves the `assetStatements` declaration to the resource file.
- Introduces additional origins resource and adds the meta-tags
  and intent-filters in `AndroidManifest.xml`.

CC @jdecked

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,15247 @@
 {
   "name": "root",
   "version": "0.4.3",
-  "lockfileVersion": 1,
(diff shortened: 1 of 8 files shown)


PEConn (approved):

jdecked:
Sweet, this works! Thanks folks. Shared with you the TWA build with this feature in use.

PR #556, opened 2021-06-29
Escapes `name` and `launcherName` when writing to Gradle

### LINKED ISSUE #555: `name` and `launcherName` in `build.gradle` are not escaped
See https://github.com/pwa-builder/PWABuilder/issues/1798#issuecomment-869759239

andreban:
- Strings need to be escaped when passed to AAPT. We introduce
a method that processes strings and output what those need to look
like in the Gradle file.

Fixes #555


packages/core/src/spec/lib/utilSpec.ts
```diff
@@ -285,4 +285,16 @@ describe('util', () => {
...
     });
   });
+
+  describe('#escapeGradleString', () => {
+    it('Escapes single quotes', () => {
+      expect(util.escapeGradleString('Single quote \''))
+          .toEqual('Single quote \\\\\\\'');
+    });
+
+    it('Escapes forward slashes', () => {
```
PEConn, line 295: That's a backwards slash, not a forwards slash.

PEConn (approved):
LGTM % comment.

packages/core/src/spec/lib/utilSpec.ts
andreban: done!

PR #559, opened 2021-07-15
Checks for existing JDK and AndroidSDK installs.

### LINKED ISSUE #547: Prevent reinstalling of JDK
**Describe the bug**
I'm running through `bubblewrap init` for the first time. The wizard first prompts for a JDK path and then an Android SDK path.

I decided to install the JDK but not the Android SDK. However, I couldn't find the Android SDK path and figured it'd be easier to just install through the tool, so I exited the wizard.

I reran `bubblewrap init` and confirmed it to download the JDK. It started downloading the JDK again even though it had already installed the JDK.

Is there a way to 

**To Reproduce**
1. Run a clean `bubblewrap init`
2. Install JDK
3. Exit binary after it installs
4. Run `bubblewrap init` again and install JDK

**Expected behavior**
I expect that the JDK, already being installed, will just no-op to save time and bandwidth.

**Screenshots**
image

**Desktop:**
 - OS: [e.g. iOS] Ubuntu
 - Browser [e.g. chrome, safari] Chrome
 - Version [e.g. 22] Bubblewrap v1.13.1

**Additional context**
This seems to work as intended once both SDKs are installed. So maybe add a second check for just JDK?

nohe427:
This change checks for existing Android SDK and JDK installs so it does
not overwrite existing installations on disk. The check is done by
seeing if the existing bubblewrap config has a value set for JDK and
AndroidSDK fields. This change aims to resolve issue #547 where the user
exits the installation process halfway through and then resumes again
and is prompted to reinstall the JDK.

Closes #547

### DIFF
--- a/packages/cli/src/lib/config.ts
+++ b/packages/cli/src/lib/config.ts
@@ -34,19 +34,7 @@ const LEGACY_CONFIG_FILE_PATH = join(LEGACY_CONFIG_FOLDER, LEGACY_CONFIG_NAME);
 const DEFAULT_JDK_FOLDER = join(DEFAULT_CONFIG_FOLDER, 'jdk');
 const DEFAULT_SDK_FOLDER = join(DEFAULT_CONFIG_FOLDER, 'android_sdk');
 
-async function createConfig(prompt: Prompt = new InquirerPrompt()): Promise<Config> {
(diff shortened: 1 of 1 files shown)


PEConn (approved):
Thanks, LGTM

andreban (approved):
LGTM, Thanks

PR #560, opened 2021-07-16
getLargestVersion implementation.

nohe427:
This implements the getLargestVersion check from the Google Play
Publisher API. This should help our community query the largest version
of an uploaded bundle to the Play Store and ensure that they can update
their package to have a larger version than that uploaded.

### DIFF
--- a/packages/cli/src/lib/cmds/play.ts
+++ b/packages/cli/src/lib/cmds/play.ts
@@ -59,7 +59,7 @@ class Play {
   * @return {number} The largest version number found in the play console.
   */
   async getLargestVersion(twaManifest: TwaManifest): Promise<number> {
-    const versionNumber = await this.googlePlay.getLargestVersion(
+    const versionNumber = await this.googlePlay.getLargestVersionCode(
         twaManifest.packageId, twaManifest.serviceAccountJsonFile!);
     return versionNumber;
   }
@@ -150,13 +150,11 @@ class Play {
     // bubblewrap play --versionCheck
     if (this.args.versionCheck) {
       const version = await this.getLargestVersion(twaManifest);
-      this.prompt.printMessage(version.toString());
-      this.prompt.printMessage(twaManifest.appVersionCode.toString());
       if (version >= twaManifest.appVersionCode) {
         const updateVersion = await this.prompt.promptConfirm(enUS.promptVersionMismatch(
             twaManifest.appVersionCode.toString(), version.toString()), true);
         if (updateVersion) {
-          twaManifest.appVersionCode = version +1;
+          twaManifest.appVersionCode = version + 1;
           await twaManifest.saveToFile(manifestFile);
           await this.updateProjectAndWarn(manifestFile);
         }
--- a/packages/cli/src/lib/cmds/play.ts
+++ b/packages/cli/src/lib/cmds/play.ts
@@ -61,7 +61,7 @@ class Play {
   */
   async getLargestVersion(twaManifest: TwaManifest): Promise<number> {
     const versionNumber = await this.googlePlay.getLargestVersion(
-      twaManifest.packageId, twaManifest.serviceAccountJsonFile!);
+        twaManifest.packageId, twaManifest.serviceAccountJsonFile!);
     return versionNumber;
   }
 
--- a/packages/cli/src/lib/cmds/play.ts
+++ b/packages/cli/src/lib/cmds/play.ts
@@ -32,6 +32,7 @@ export interface PlayArgs {
   manifest?: string;
   appBundleLocation?: string;
   targetDirectory?: string;
+  versionCheck?: boolean;
   // versionCheck?: boolean; // Uncomment when getLargetVersion is implemented.
 }
 
@@ -58,12 +59,10 @@ class Play {
   * to give x+1 version number.
   * @return {number} The largest version number found in the play console.
   */
-  async getLargestVersion(): Promise<number> {
-    // Need to get an editId, then list all apks available.
-    // This should allow us to query the highest apk number.
-    // This exists in Gradle play plugin but is not easily accessible over CLI.
-    // This should be completed in a future CL.
-    throw new Error('Not Implemented');
+  async getLargestVersion(twaManifest: TwaManifest): Promise<number> {
+    const versionNumber = await this.googlePlay.getLargestVersion(
+      twaManifest.packageId, twaManifest.serviceAccountJsonFile!);
+    return versionNumber;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


nohe427:
I realize this will have a lot of linting errors. Some of them I am having trouble resolving.

packages/core/src/lib/GooglePlay.ts
```diff
@@ -58,4 +61,53 @@ export class GooglePlay {
     await this.gradleWrapper.executeGradleCommand(
         ['publishBundle', '--artifact-dir', filepath, '--track', track]);
   }
+
+  /**
+   * This calls the Google Play Console through the nodejs api and then calls bundles.list which
```
PEConn, line 66: I feel this doc is a little too conversational - do you mind changing it to something like "Connects to the Google Play Console, retrieves a list of all Android App Bundles for the given packageName and returns the max versionCode of those bundles. Considers both ChromeOS and Android releases".

packages/core/src/lib/GooglePlay.ts
```diff
@@ -58,4 +61,53 @@ export class GooglePlay {
...
+   * lists all current Android App Bundles of the app and edit which should help us narrow down the
+   * highest uploaded app bundle in the play console. This works irrespective of whether this is a
+   * Chrome OS or Android only release.
+   */
+  async getLargestVersion(
```
PEConn, line 71: Could this be called either `getLatestVersion` or `getLargestVersionCode`?

packages/core/src/lib/GooglePlay.ts
```diff
@@ -58,4 +61,53 @@ export class GooglePlay {
...
+      packageName: string,
+      serviceAccountJsonFilePath: string,
+  ): Promise<number> {
+    if (!this._googlePlayApi) {
+      this._googlePlayApi = this.getAndroidClient(serviceAccountJsonFilePath);
+    }
+    const edit = await this._googlePlayApi.edits.insert({packageName: packageName});
```
PEConn, line 78: So I don't know anything about the Google Play APIs, but it seems a bit surprising that we need to use `edits` to just get a list of packages. If we somehow crash (or the user terminates execution) between line 81 and line 89, have we left some permanent change on the server?

packages/core/src/lib/GooglePlay.ts
```diff
@@ -58,4 +61,53 @@ export class GooglePlay {
...
+    const editId = edit.data.id!;
+    const bundleResponse =
+      await this._googlePlayApi.edits.bundles.list({packageName: packageName, editId: editId});
+    let versionCode = 1;
```
PEConn, line 82: I think this could be simplified to:

```
Math.max(bundleResponse.data.bundles.map(bundle => bundle.versionCode))
```

packages/core/src/lib/GooglePlay.ts
```diff
@@ -15,6 +15,7 @@
  */
 
 import {GradleWrapper} from '..';
+import {androidpublisher_v3, google} from 'googleapis';
```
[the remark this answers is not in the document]
andreban: It seems both the `push ` and `pull_request` workflows were triggered here. This caused everything to run at least twice? Then, again for each Node version supported?

packages/cli/src/lib/cmds/play.ts
```diff
@@ -144,6 +146,23 @@ class Play {
     if (this.args.init) {
       await this.bootstrapPlay();
     }
+
+    // bubblewrap play --versionCheck
+    if (this.args.versionCheck) {
+      const version = await this.getLargestVersion(twaManifest);
+      this.prompt.printMessage(version.toString());
+      this.prompt.printMessage(twaManifest.appVersionCode.toString());
```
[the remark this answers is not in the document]
nohe427: We don't. This was me running a check  and forgetting to move it. Will remove.

packages/core/src/lib/GooglePlay.ts
```diff
@@ -58,4 +61,53 @@ export class GooglePlay {
...
+    // Initialize the Google API Client from service account credentials
+    const jwtClient = new google.auth.JWT({
+      keyFile: serviceAccountJsonFilePath,
+      scopes: ['https://www.googleapis.com/auth/androidpublisher'],
+    },
+    );
+
+    // Connect to the Google Play Developer API with JWT Client
+    return google.androidpublisher({
+      version: 'v3',
```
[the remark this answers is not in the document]
nohe427: The v3 comes from the [resource description](https://developers.google.com/android-publisher/api-ref/rest). AFAIK, old versions are left around for awhile so there is a migration period for us if need be. I think Play Developer API updates are generally far and few between.

packages/core/src/lib/GooglePlay.ts
nohe427: Yes, I found this a bit odd as well in my research. Fourtunately, edits automatically [come with a timeout](https://developers.google.com/android-publisher/api-ref/rest/v3/edits?hl=zh-TW#AppEdit), so if its not committed, it should just expire and not leave dangling edits on the server.

packages/core/src/lib/GooglePlay.ts
nohe427: This is a bad habit of mine. Changed it to be a little less conversational.

packages/core/src/lib/GooglePlay.ts
nohe427: Awesome. Thanks for that hint.

packages/core/src/lib/GooglePlay.ts
```diff
@@ -58,4 +61,47 @@ export class GooglePlay {
...
+   * This fetches the Android client using the bubblewrap configuration file.
+   */
+  private getAndroidClient(
+      serviceAccountJsonFilePath: string,
+  ): androidpublisher_v3.Androidpublisher {
+    // Initialize the Google API Client from service account credentials
+    const jwtClient = new google.auth.JWT({
+      keyFile: serviceAccountJsonFilePath,
+      scopes: ['https://www.googleapis.com/auth/androidpublisher'],
+    },
```
PEConn, line 98: nit: I don't think we need this comma.

PEConn (approved):
LGTM to me, let's see what Andre says!

packages/core/src/lib/GooglePlay.ts
nohe427: The linter wants a comma there.

andreban: Yep, our linter requires dangling commas. That makes diffs nicer when adding new items to the parameters.

andreban (approved):
LGTM

PR #561, opened 2021-07-16
Fix missing checksum

joycetoh8:
This is supposed to fix what happens when a checksum file doesn't currently exist yet to verify the state of twa-manifest.json. This will happen for users updating to the most recent version of Bubblewrap and run `bubblewrap build` without running `bubblewrap update` which will generate the checksum. When `bubblewrap build` doesn't find the checksum file, it will prompt the user whether they would like to generate their project again (which will create checksum) or build without regenerating the project.

### DIFF
--- a/packages/cli/src/lib/cmds/build.ts
+++ b/packages/cli/src/lib/cmds/build.ts
@@ -142,6 +142,26 @@ class Build {
         APP_BUNDLE_BUILD_OUTPUT_FILE_NAME, APP_BUNDLE_SIGNED_FILE_NAME);
   }
 
+  /**
+   *
+   * Based on the promptResponse to update the project or not, run an update or print the relevant warning message.
+   */
+  async runUpdate(
+      promptResponse: boolean,
+      manifestFile: string,
+      noUpdateMessage: string): Promise<boolean> {
+    if (promptResponse) {
+      const updated = await updateProject(false, null, this.prompt,
+          this.args.directory, manifestFile);
+      if (!updated) {
+        return false;
+      }
+    } else {
+      this.prompt.printMessage(noUpdateMessage);
+    }
+    return true;
+  }
+
   async build(): Promise<boolean> {
     if (!await this.androidSdkTools.checkBuildTools()) {
       this.prompt.printMessage(messages.messageInstallingBuildTools);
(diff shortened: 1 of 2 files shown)


packages/cli/src/lib/cmds/build.ts
```diff
@@ -142,6 +142,25 @@ class Build {
...
   }
 
+  /**
+   * Based on the promptResponse to update the project or not, run an update or print the relevant warning message.
+   */
+  async runUpdate(
+      promptResponse: boolean,
+      manifestFile: string,
+      noUpdateMessage: string): Promise<boolean> {
+    if (promptResponse) {
```
PEConn, line 152: nit: do you mind turning this around and making it a guard clause, so:

```
if (!promptResponse) {
  // print message
  return true;
}
```

It just means that the meat of the function is less indented.

packages/cli/src/lib/cmds/build.ts
```diff
@@ -142,6 +142,25 @@ class Build {
         APP_BUNDLE_BUILD_OUTPUT_FILE_NAME, APP_BUNDLE_SIGNED_FILE_NAME);
   }
 
+  /**
+   * Based on the promptResponse to update the project or not, run an update or print the relevant warning message.
```
PEConn, line 146: Could you explain what the return value signifies?

packages/cli/src/lib/cmds/build.ts
```diff
@@ -156,20 +175,31 @@ class Build {
...
-        const updated = await updateProject(false, null, this.prompt,
-            this.args.directory, manifestFile);
-        if (!updated) {
-          return false;
-        }
-      } else {
-        this.prompt.printMessage(messages.messageProjectNotUpdated);
+    const targetDirectory = this.args.directory || process.cwd();
+    const checksumFile = path.join(targetDirectory, 'manifest-checksum.txt');
+    let updateResult = true;
```
PEConn, line 180: Could this be "updateSuccessful"?

packages/cli/src/lib/cmds/build.ts
joycetoh8: good call out! done :)

packages/cli/src/lib/cmds/build.ts
joycetoh8: Added a @return to the method comment

packages/cli/src/lib/cmds/build.ts
joycetoh8: done!

PEConn (approved):

PR #562, opened 2021-07-16
Add a warning for the play family policy.

### LINKED ISSUE #545: Add a warning for the Play Family policy when building the application
A Trusted Web Activity application may be incompatible with the [Play Family policy](https://web.dev/using-a-pwa-in-your-android-app/#quality-criteria). Let's add this warning to the `build` command, so developers get this information even if missed on the docs.

nohe427:
This adds a warning for the Google Play Family Policy to let developers
know about being in compliance and also provides a link for them to
visit to find out more information.
Fixes #545

### DIFF
--- a/packages/cli/src/lib/strings.ts
+++ b/packages/cli/src/lib/strings.ts
@@ -369,7 +369,7 @@ the PWA:
   promptVersionCode: 'Starting version code for the new app version:',
   promptUpdateProject: 'There are changes in twa-manifest.json. ' +
       'Would you like to apply them to the project before building?',
-  warnFamilyPolicy: bold(red('WARNING: ')) + 'Trusted Web Activities are currently incompatible' +
+  warnFamilyPolicy: bold(yellow('WARNING: ')) + 'Trusted Web Activities are currently incompatible' +
       ' with applications\ntargeting children under the age of 13.' +
       ' Check out the Play for' +
       ' Families\npolicies to learn more.\n' +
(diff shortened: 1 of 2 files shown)


nohe427:
@andreban I took a shot at addressing #545 with this Cl. The only challenge is that when I run the tool from VSCode, it looks like the warning may get cropped off. See this screenshot which shows the warning:

Screenshot 2021-07-16 2 20 38 PM

and this is the screenshot of what it looks like in my terminal after running the tool:
Screenshot 2021-07-16 2 20 45 PM

I am wondering if its better to have some sort of acknowledge the message or something rather than just posting a warning that may get cropped by the terminal.

Is this similar to what you had in mind? Do you have a preferred link to use?

nohe427:
Penny was able to provide this link : https://support.google.com/googleplay/android-developer/answer/9893335 and with a little guidance the policy can be found under family policy requirements - app functionality regarding web views. What do you think of picking one link over the other?

packages/cli/src/lib/strings.ts
```diff
@@ -368,6 +369,11 @@ the PWA:
   promptVersionCode: 'Starting version code for the new app version:',
   promptUpdateProject: 'There are changes in twa-manifest.json. ' +
       'Would you like to apply them to the project before building?',
+  warnFamilyPolicy: bold(red('WARNING: ')) + 'Trusted Web Activities are currently incompatible' +
```
andreban, line 372: Looks good. I wonder if yellow is better for warning, as red can lead developers to think its an error/

nohe427:
Yellow WARNING text

What do you think of the yellow?

PEConn (approved):

andreban (approved):
LGTM

Issue #563, opened 2021-07-17
Random jasmine tests in packages/core are hanging


nohe427:
**Describe the bug**
Random jasmine tests in packages/core are hanging and timing out the github workflow runners after six hours.

**To Reproduce**
1. Download the bubblewrap repo on an Ubuntu machine
2. run `npm ci` and then `npm run test`
3. After a few test runs, it should hang. I put the random seed number in the context notes below for additional info to try to reporduce.

**Expected behavior**
Jasmine tests do not fail

**Desktop (please complete the following information):**
 - OS: Ubuntu-18.04
 - Browser n/a
 - Version HEAD

**Additional context**
I tried testing on a Mac machine but the svg2img dep is blocking me from running `npm ci` successfully at the moment.

Output from GitHub workflows and from my Ubuntu machine:

```
lerna info run Ran npm script 'test' in '@bubblewrap/core' in 8.4s:

> @bubblewrap/core@1.13.2 test /workspaces/bubblewrap/packages/core
> tsc && jasmine --config=jasmine.json

Randomized with seed 05494
Started
.............................................................................................................


109 specs, 0 failures
Finished in 0.175 seconds
Randomized with seed 05494 (jasmine --random=true --seed=05494)
```

andreban:
This is the first commit that is broken: https://github.com/GoogleChromeLabs/bubblewrap/commit/fcd433701552df943926021676c657e9d0dbf521

I used `nvm use 12` to use Node 12, followed by `npm run clean`, `npm run clean-modules`, `npm install` and `npm run test`.  The first 2 will remove modules so you can install the svg2img for the latest version.

Commenting out the tests in `configSpec.ts`. Commenting out the tests in `configSpec.ts` prevents things from hanging.

andreban:
Commenting out this test specifically https://github.com/GoogleChromeLabs/bubblewrap/blob/main/packages/cli/src/spec/configSpec.ts#L57-L91 stops the tests runner from hanging.

andreban:
Actually, it seems that any of the tests using `mock-fs` cause the tests to hang.

I'm using `jasmine --config=jasmine.json --random=false --seed=32234` to ensure it always hangs.

andreban:
I think the problem is in the two `if` statements here: https://github.com/GoogleChromeLabs/bubblewrap/blob/fcd433701552df943926021676c657e9d0dbf521/packages/cli/src/lib/config.ts#L113-L123.

[...]

PR #564, opened 2021-07-19
Fix the tests.

nohe427:
This adds a jdkPath and androidSdkPath to the config.json file. This
should stop the tests from crashing on the CI/CD bots.


nohe427:
Fixes #563

nohe427:
Ran and re-ran the tests on CI bots. Seems to work.

PEConn (approved):

Issue #534, opened 2021-05-13
Project generation should fail if `iconUrl` is not available


andreban:
It's not possible to generate a valid project if `iconUrl` is not a valid URL.

nohe427:
Do you want a regex validating that `iconUrl` is a url or do you want to run the icon check like you have done in a [previous CL](https://github.com/GoogleChromeLabs/bubblewrap/commit/e54a67ead0615720c7a900e1aa1651489ee76105) at project generation time?

andreban:
We can probably just [create a new URL](https://nodejs.org/api/url.html#url_new_url_input_base) and catch the error.

nohe427:
@andreban - This looks to already be implemented with the code that exists today:
![image](https://user-images.githubusercontent.com/3211589/126014225-d5e49456-f787-4734-aea8-fa9968fbdf02.png)

This is running from the init step. Is there another place or issue that you were thinking?

andreban:
Yes, init does the proper check. However, the developer can manually update twa-manifest.json and then run update. I'm not sure we do a proper check there.

andreban:
Closed via #565

PR #565, opened 2021-07-19
Validate iconUrl on update.

nohe427:
This adds a check to the updateProject command so it can validate the
iconUrl,


nohe427:
Related to #534

PEConn (approved):

PR #567, opened 2021-07-19
Disable the play command.

nohe427:
In discussions that we have had, we decided it was best to disable the
play command before we cut another release of bubblewrap. Once we are
more confident in our ability with the command, we will turn it on for
expiremental use.


andreban (approved):
LGTM, with only a note on commenting out the imports so we make the linter happy.

nohe427:
This seems to be hung on the CLA check for me.

PR #570, opened 2021-07-22
Await the saveConfig

nohe427:
This fixes a bug where the AndroidSDK could be downloaded twice because the config was not being saved if an invalid manifest was supplied. Adding the await will wait for the saveConfig method to finish before continuing execution.

### DIFF
--- a/packages/cli/src/lib/config.ts
+++ b/packages/cli/src/lib/config.ts
@@ -113,13 +113,13 @@ export async function loadOrCreateConfig(
   if (!config.jdkPath) {
     const jdkPath = await configureJdk(prompt);
     config.jdkPath = jdkPath;
-    config.saveConfig(configPath);
+    await config.saveConfig(configPath);
   }
 
   if (!config.androidSdkPath) {
     const androidSdkPath = await configAndroidSdk(prompt);
     config.androidSdkPath = androidSdkPath;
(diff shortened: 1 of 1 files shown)


PEConn (commented):
So I can see the "adding await" part to your solution, but I'm not sure how this prevents multiple downloads from happening?

andreban:
I think there may be a very short space of time where the JDK has been installed, the config hasn't been saved, and the application is killed before it is saved. Then, running the application again will reinstall the JDK. This should be quite uncommon though, and I'm not sure adding await would fix it. Adding `await` here is unlikely to have bad side effects, but I'm not sure it is needed too, since `saveConfig()` doesn't change the config in and persisting to disk can be happen asynchronously.

nohe427:
The issue for me wasn't the JDK getting installed twice, but the Android SDK. What I did to get this error was I went and called :  
`bubblewrap init --manifest <domain>/manifest.json` and did not supply the `https` protocol. After installing the Android SDK, the request failed for the manifest as it wasn't a valid URL and when I called `bubblewrap init...` with the right parameters, then the download restarted for the android sdk. I realize this is a bit of an edge case, but I happened to cut myself on this one last night.

andreban (approved):
LGTM

PR #571, opened 2021-07-22
Refactor from GradlePlayPlugin to API Directly

nohe427:
This is a refactor of the current GooglePlay integration so it calls the Play API directly rather than the gradle play plugin. This ensures that we are not excessively rebuilding the project or regenerating the gradle file. This update should have no affect on the end users experience of using the `play` command with the exception of removing the no longer needed `init` command.

### DIFF
--- a/packages/cli/src/lib/cmds/play.ts
+++ b/packages/cli/src/lib/cmds/play.ts
@@ -80,6 +80,10 @@ class Play {
     return await this.internalPublish(defaultPath, userSelectedTrack, twaManifest);
   }
 
+  /**
+  * Calls the internalPublish commands.
+  * @return {boolean} Whether the publish command completes successfully or not.
+  */
   private async internalPublish(
       bundleLocation: string,
       userSelectedTrack: string,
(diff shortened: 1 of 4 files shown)


nohe427:
I did not implement retain just yet, but plan on doing that in a future release.

packages/core/template_project/app/build.gradle
```diff
@@ -18,9 +18,6 @@ import groovy.xml.MarkupBuilder
 
 plugins {
     id 'com.android.application'
-    <% if (serviceAccountJsonFile) { %>
```
andreban, line 21: Does somehow change that gradle is signing the APKs (opposed to use apksigner?). Asking because I remember this was changed when we introduced this.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -39,36 +36,27 @@ export interface PlayArgs {
   * The Play class is the class that is used to communicate with the Google Play Store.
   */
 class Play {
+  private googlePlay?: GooglePlay;
```
andreban, line 39: I liked when this was initialised in the constructor, as it prevents errors. 

Maybe we can pull the bit that loads the TWA Manifest to before this is initialised, so we can build GooglePlay and pass it on the constructor?

nohe427: Done.

packages/core/template_project/app/build.gradle
nohe427: This change needed our gradle version to be updated, so [I updated it here](https://github.com/GoogleChromeLabs/bubblewrap/pull/491/files#diff-86b7a2a741060fd97a8c701a201b0db24189f86358593672cf3e02d33d0edd15R26). Moving forward, gradle will still do signing of the APKs and apksigner can still check for alignment I believe.

nohe427:
@PEConn - Yes, I can do that.

PR #574, opened 2021-07-28
Add publishing a bundle via API instead of Gradle.

nohe427:
This adds the capability to publish a bundle using the API directly
rather than depending on a gradle task.

### DIFF
--- a/packages/core/src/lib/GooglePlay.ts
+++ b/packages/core/src/lib/GooglePlay.ts
@@ -16,6 +16,7 @@
 
 import {GradleWrapper} from '..';
 import {androidpublisher_v3 as androidPublisher, google} from 'googleapis';
+import {createReadStream} from 'fs';
 
 // Possible values for release tracks
 const TRACK_VALUES = ['alpha', 'beta', 'internal', 'production'];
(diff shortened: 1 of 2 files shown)


nohe427:
This is still tracking the playExperiement branch so we can merge this into the main branch all at once

packages/core/src/lib/GooglePlay.ts
```diff
@@ -62,12 +63,104 @@ export class GooglePlay {
...
+   *   ChromeOS only releases.
+   */
+  async publishBundle(
+      track: PlayStoreTrack,
+      filepath: string,
+      packageName: string,
+      retainedBundles: number[],
+  ): Promise<void> {
+    // TODO(@nohe427): Remove this check when refactor is finished.
+    if (!this._googlePlayApi) {
```
andreban, line 95: nit: I know this is going to be removed, but maybe throw an error or a log message in the meantime?

packages/core/src/lib/GooglePlay.ts
```diff
@@ -62,12 +63,104 @@ export class GooglePlay {
...
+      return;
+    }
+    const edit = await this._googlePlayApi.edits.insert({packageName: packageName});
+    const editId = edit.data.id;
```
andreban, line 99: nit: it seems`edit.data.id` be null here, since we're adding the `!` a few lines below. Should we make a check for null/undefined and throw an error if that's the case? That would allow removing the `!` below.

packages/core/src/lib/GooglePlay.ts
```diff
@@ -62,12 +63,104 @@ export class GooglePlay {
...
+          packageName: packageName,
+          media: {
+            body: createReadStream(filepath),
+          },
+        },
+        {
+          timeout: 160000,
+        });
+
+    const versionCodeUploaded = result.data.versionCode;
```
andreban, line 113: nit: same as the note on `edit.data.id`

andreban (approved):
LGTM with a couple of notes that I usually try avoiding using `!` when possible and prefer doing a null check. If we know those fields are not null, but don't have a way to express it, them I'm fine, but wanted to make sure.

packages/core/src/lib/GooglePlay.ts
```diff
@@ -62,12 +63,104 @@ export class GooglePlay {
...
+   * Calls the following Play API commands in order:
+   * Edits.Insert
+   * Edits.Bunldes.Upload
+   * Edits.Tracks.Update
+   * Edits.Commit
+   * https://developers.google.com/android-publisher/api-ref/rest/v3/edits.bundles/upload
+   * https://developers.google.com/android-publisher/edits#workflow
+   * @param track - Specifies the track that the user would like to publish to.
+   * @param filepath - Filepath of the App bundle you would like to upload.
+   * @param packageName - packageName of the bundle.
```
PEConn, line 84: nit: `@param packageName - Package name of the bundle.`

packages/core/src/lib/GooglePlay.ts
```diff
@@ -62,12 +63,104 @@ export class GooglePlay {
...
+   * @param retainedBundles - all bundles that should be retained on upload. This is useful for
```
PEConn, line 85: nit: `All`

packages/core/src/lib/GooglePlay.ts
```diff
@@ -62,12 +63,104 @@ export class GooglePlay {
...
     await this.gradleWrapper.executeGradleCommand(
         ['publishBundle', '--artifact-dir', filepath, '--track', track]);
   }
 
+  /**
+   * This calls the publish bundle command and publishes an existing artifact to Google
+   * Play.
+   * Calls the following Play API commands in order:
+   * Edits.Insert
+   * Edits.Bunldes.Upload
```
PEConn, line 77: nit: typo

packages/core/src/lib/GooglePlay.ts
```diff
@@ -62,12 +63,104 @@ export class GooglePlay {
...
+        {
+          ackBundleInstallationWarning: false,
+          editId: editId!,
+          packageName: packageName,
+          media: {
+            body: createReadStream(filepath),
+          },
+        },
+        {
+          timeout: 160000,
```
PEConn, line 110: Do you mind moving this to a constant somewhere?

(Also, was there any logic behind picking 160s? 2 and 2/3 minutes seems like a bit of a random timeout.)

packages/core/src/lib/GooglePlay.ts
```diff
@@ -62,12 +63,104 @@ export class GooglePlay {
...
+        });
+
+    const versionCodeUploaded = result.data.versionCode;
+    const retainedBundlesStr = retainedBundles.map(((n) => n.toString()));
```
PEConn, line 114: I think you can get rid of some of the parentheses here (does `.map(n => n.toString())` not work?).

packages/core/src/lib/GooglePlay.ts
PEConn: I think throwing an error is fine here - it doesn't need to be tidy since users won't come across it, but it makes it obvious what went wrong if a developer hits it.

packages/core/src/lib/GooglePlay.ts
nohe427: Yes, my math was wrong. The documentation says to set it to a 2 to 3 minute timeout and I was aiming for three minutes. Updated to a const and the timeout.

packages/core/src/lib/GooglePlay.ts
nohe427: Yup this works.

nohe427:
@PEConn & @andreban - Do you both want to give it a once over again?

packages/core/src/lib/GooglePlay.ts
nohe427: Looks like the linter is adding in some extra `( )`

packages/core/src/lib/GooglePlay.ts
```diff
@@ -16,12 +16,15 @@
...
 import {GradleWrapper} from '..';
 import {androidpublisher_v3 as androidPublisher, google} from 'googleapis';
+import {createReadStream} from 'fs';
 
 // Possible values for release tracks
 const TRACK_VALUES = ['alpha', 'beta', 'internal', 'production'];
 export type PlayStoreTrack = typeof TRACK_VALUES[number];
 export const PlayStoreTracks: PlayStoreTrack[] = [...TRACK_VALUES];
 
+const PLAY_API_TIMEOUT = 180000;
```
PEConn, line 26: Could you add `_MS` to the end of the variable name?

packages/core/src/lib/GooglePlay.ts
```diff
@@ -62,12 +65,110 @@ export class GooglePlay {
...
+      retainedBundles: number[],
+  ): Promise<void> {
+    // TODO(@nohe427): Remove this check when refactor is finished.
+    if (!this._googlePlayApi) {
+      throw new Error('Service Account JSON file not set.');
+    }
+    const edit = await this._googlePlayApi.edits.insert({packageName: packageName});
+    const editId = edit.data.id;
+    if (!editId) {
+      throw new Error('Could not create a Google Play edit');
```
PEConn, line 103: For a follow up CL - do we know why this could fail? Should we retry or should we just tell the user "sorry, that didn't work, try again later"?

PEConn (approved):

packages/core/src/lib/GooglePlay.ts
nohe427: I am not sure why it would potentially fail just yet. I am going to add a TODO here.

PR #575, opened 2021-07-29
Remove init command.

nohe427:
The gradle bootstrapPlay command was extraneous. It synced the play
representation down in a file format that could be looked over. This
provided no benefit to the end user.


andreban (approved):
LGTM, Thank you

PEConn (approved):

PR #577, opened 2021-07-30
Adjust the CLI to not use any Gradle cmds.

nohe427:
This adjusts the CLI to not use any Gradle commands and only uses the
Play Publisher API commands. In a future CL, we will remove all the
gradle play publisher commands as they are no longer needed.

### DIFF
--- a/packages/cli/src/lib/cmds/play.ts
+++ b/packages/cli/src/lib/cmds/play.ts
@@ -14,10 +14,7 @@
  *  limitations under the License.
  */
 
-import {
-  Config, GradleWrapper, JdkHelper, AndroidSdkTools, ConsoleLog, Log, GooglePlay, TwaManifest,
-  asPlayStoreTrack,
-} from '@bubblewrap/core';
+import {GooglePlay, TwaManifest, asPlayStoreTrack} from '@bubblewrap/core';
 import * as fs from 'fs';
 import * as path from 'path';
 import {TWA_MANIFEST_FILE_NAME} from '../constants';
@@ -50,16 +47,14 @@ class Play {
   * @return {number} The largest version number found in the play console.
   */
   async getLargestVersion(twaManifest: TwaManifest): Promise<number> {
-    // TODO(@nohe427): This will return 0 until refactor finished.
-    const versionNumber = await this.googlePlay.getLargestVersionCode(twaManifest.packageId);
-    return versionNumber;
+    return await this.googlePlay.getLargestVersionCode(twaManifest.packageId);
   }
 
   /**
   * Publishes the Android App Bundle to the specified from the {@link PlayArgs}.
   * @return {boolean} Whether the publish command completes successfully or not.
   */
-  async publish(): Promise<boolean> {
+  async publish(twaManifest: TwaManifest): Promise<boolean> {
     // Validate that the publish value is listed in the available Tracks.
     // If no value was supplied with publish we make it internal.
     const userSelectedTrack = asPlayStoreTrack(this.args.publish?.toLowerCase() || 'internal');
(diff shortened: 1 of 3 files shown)


nohe427:
This is going to the `playExpirement` branch and once this is merged, I will remove all the gradle play publisher bits.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -50,62 +47,54 @@ class Play {
...
       this.prompt.printMessage(enUS.messageInvalidTrack);
       return false;
     }
-    // appbundlelocation is an option argument.
-    if (this.args.appBundleLocation && fs.existsSync(this.args.appBundleLocation!!)) {
-      await this.googlePlay.publishBundleWithGradle(userSelectedTrack, this.args.appBundleLocation);
-      return true;
+
+    // Default file path
+    const defaultSignedAppBundleFileName = 'app-release-bundle.aab';
```
andreban, line 67: can we move this to a constant on the top of the file?

packages/cli/src/lib/cmds/play.ts
```diff
@@ -161,12 +137,43 @@ class Play {
...
-  const gradleWrapper = new GradleWrapper(process, androidSdkTools);
-  const googlePlay = new GooglePlay(gradleWrapper);
+/**
+  * Validates that the service account JSON file exists.
+  * @param {string | undefined} path - The path the the JSON file.
+  * @return {boolean} Whether or not the JSON file exists.
+  */
+function validServiceAccountJsonFile(path: string | undefined): boolean {
+  if (path == undefined) {
+    // Log an error
```
andreban, line 147: Are we missing a TODO here?
```suggestion
    //TODO: Log an error
```

packages/cli/src/lib/cmds/play.ts
```diff
@@ -161,12 +137,43 @@ class Play {
...
+    return false;
+  }
+  if (!fs.existsSync(path)) {
+    // path doesn't exist log an error
```
andreban, line 151: Same as above

packages/core/src/lib/GooglePlay.ts
```diff
@@ -44,7 +44,7 @@ export class GooglePlay {
    * @param serviceAccountJsonFilePath This is the service account file to communicate with the
    *   play publisher API.
    */
-  constructor(private gradleWrapper: GradleWrapper, serviceAccountJsonFilePath?: string) {
+  constructor(private gradleWrapper?: GradleWrapper, serviceAccountJsonFilePath?: string) {
```
andreban, line 47: Why is `gradleWrapper` now optional?

andreban (changes requested):

packages/cli/src/lib/cmds/play.ts
nohe427: Done

packages/core/src/lib/GooglePlay.ts
nohe427: I made the `gradleWrapper` optional because in the next CL I plan on removing all the gradle changes. This is no longer needed after that change.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -50,62 +50,52 @@ class Play {
...
   */
   async getLargestVersion(twaManifest: TwaManifest): Promise<number> {
-    // TODO(@nohe427): This will return 0 until refactor finished.
-    const versionNumber = await this.googlePlay.getLargestVersionCode(twaManifest.packageId);
-    return versionNumber;
+    return await this.googlePlay.getLargestVersionCode(twaManifest.packageId);
   }
 
   /**
   * Publishes the Android App Bundle to the specified from the {@link PlayArgs}.
```
PEConn, line 57: nit: there's a typo in this line - to the specified what?

packages/cli/src/lib/cmds/play.ts
```diff
@@ -50,62 +50,52 @@ class Play {
...
     if (userSelectedTrack == null) {
       this.prompt.printMessage(enUS.messageInvalidTrack);
       return false;
     }
-    // appbundlelocation is an option argument.
-    if (this.args.appBundleLocation && fs.existsSync(this.args.appBundleLocation!!)) {
-      await this.googlePlay.publishBundleWithGradle(userSelectedTrack, this.args.appBundleLocation);
-      return true;
+
+    // Where we should find our default output file
```
PEConn, line 69: nit: I don't think this comment adds anything - the variable already has a pretty clear name.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -50,62 +50,52 @@ class Play {
...
+    const defaultPath = path.join(process.cwd(), defaultSignedAppBundleFileName);
+
+    const publishFilePath = this.args.appBundleLocation || defaultPath;
+
+    if (!fs.existsSync(publishFilePath)) {
+      throw new Error(`App Bundle not found on disk at location: ${publishFilePath}`);
```
PEConn, line 75: nit: "App Bundle not found at", the "on disk" bit doesn't really add much.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -50,62 +50,52 @@ class Play {
...
-    }
-    return true;
-  }
-
-  private async updateProjectAndWarn(manifestFile: string): Promise<void> {
+  private async updateProjectAndWarn(manifestFile: string, appVersionName?: string): Promise<void> {
     await updateProject(
-        true, null, this.prompt, this.args.targetDirectory || process.cwd(), manifestFile);
+        true,
+        appVersionName || null,
```
PEConn, line 95: Do we need `appVersionName || null`? `appVersionName` is either going to have a value or be `null` anyway. We can just pass `appVersionName`.

PEConn (approved):
LGTM with some nits.

packages/cli/src/lib/cmds/play.ts
nohe427: Yes because appVersionName in updateProject takes `string | null`. appVersionName is `string | undefined`. Would it be better to update project here to take `string | undefined`?

packages/cli/src/lib/cmds/play.ts
nohe427: Think I fixed this

PEConn (approved):

PR #586, opened 2021-08-12
Removed gradle and TripleT Dependency for play.

nohe427:
This removes the gradle and triple-t dependency for Google Play
Publishing. This implementation now directly communicates with the
Google Play API. This allows us to have more fine grained control on
communicating with Play and build in more features to bubblewrap as
needed.

### DIFF
--- a/packages/cli/src/lib/cmds/play.ts
+++ b/packages/cli/src/lib/cmds/play.ts
@@ -145,11 +145,9 @@ class Play {
   */
 function validServiceAccountJsonFile(path: string | undefined): boolean {
   if (path == undefined) {
-    // TODO(@nohe427): Log an error
     return false;
   }
   if (!fs.existsSync(path)) {
-    // TODO(@nohe427): path doesn't exist log an error
     return false;
   }
   return true;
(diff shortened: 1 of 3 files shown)


nohe427:
The next CL should be the retain bundles command and should wrap up this feature.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -145,11 +145,9 @@ class Play {
   */
 function validServiceAccountJsonFile(path: string | undefined): boolean {
   if (path == undefined) {
-    // TODO(@nohe427): Log an error
```
nohe427, line 148: **FYI: ** Error handling for this is handled on line 168 (of this file). I can throw errors here if needed.

packages/cli/src/lib/cmds/play.ts
```diff
@@ -165,11 +163,12 @@ async function setupGooglePlay(args: PlayArgs): Promise<GooglePlay> {
     twaManifest.serviceAccountJsonFile = args.serviceAccountFile;
     await twaManifest.saveToFile(manifestFile);
   }
-  if (!validServiceAccountJsonFile(twaManifest.serviceAccountJsonFile)) {
-    throw new Error('enUS.messageServiceAccountJSONMissing');
+  if (!validServiceAccountJsonFile(twaManifest.serviceAccountJsonFile) ||
+    !twaManifest.serviceAccountJsonFile) {
```
andreban, line 167: I think we can reverse the condition order here so that if `twaManifest.serviceAccountJsonFile` is `null` or `undefined` then `validServiceAccountJsonFile()` is not even called.
```suggestion
  if (!twaManifest.serviceAccountJsonFile ||
      !validServiceAccountJsonFile(twaManifest.serviceAccountJsonFile)) {
```

packages/cli/src/lib/cmds/play.ts
andreban: I'm happy with returning false if `path` is `null` or `undefined.

andreban (approved):
LGTM, once the `if` order is fixed.

packages/cli/src/lib/cmds/play.ts
nohe427: Ahh, awesome. Thanks!

PR #588, opened 2021-08-17
Add retain bundles feature

nohe427:
This adds the retain bundles feature to the Google Play integration.
This allows users when trying to publish an Android App bundle the
ability to select which bundles they would like to retain. This is
important for when a developer wants to release a ChromeOS only release
alongside an Android release.

### DIFF
--- a/packages/cli/src/lib/cmds/play.ts
+++ b/packages/cli/src/lib/cmds/play.ts
@@ -21,7 +21,6 @@ import {TWA_MANIFEST_FILE_NAME} from '../constants';
 import {Prompt, InquirerPrompt} from '../Prompt';
 import {updateProject} from './shared';
 import {enUS} from '../strings';
-import {PlayOperationResult} from '@bubblewrap/core/dist/lib/GooglePlay';
 
 export interface PlayArgs {
   publish?: string;
@@ -54,14 +53,11 @@ class Play {
   * @return {number} The largest version number found in the play console.
   */
   async getLargestVersion(twaManifest: TwaManifest): Promise<number> {
-    const result = await this.googlePlay.performPlayOperation(
-        twaManifest.packageId, async (editId: string): Promise<PlayOperationResult> => {
+    const versionCode = await this.googlePlay.performPlayOperation(
+        twaManifest.packageId, async (editId: string): Promise<number> => {
           return await this.googlePlay.getLargestVersionCode(twaManifest.packageId, editId);
         });
-    if (result.getLargestVersionCodeResult) {
-      return result.getLargestVersionCodeResult;
-    }
-    throw new Error('Play operation failed. Could not find largest version code.');
+    return versionCode;
   }
 
   /**
@@ -88,7 +84,7 @@ class Play {
 
     const retainedBundles = twaManifest.retainedBundles;
     await this.googlePlay.performPlayOperation(twaManifest.packageId,
-        async (editId: string): Promise<PlayOperationResult> => {
+        async (editId: string): Promise<void> => {
           return await this.googlePlay.publishBundle(
               userSelectedTrack, publishFilePath, twaManifest.packageId, retainedBundles, editId);
         });
@@ -151,12 +147,12 @@ class Play {
                 twaManifest.appVersionCode, versionToRetain));
       }
       // Validate that the version exists on the Play Servers.
-      const result = await this.googlePlay.performPlayOperation(twaManifest.packageId,
-          async (editId: string): Promise<PlayOperationResult> => {
+      const exists = await this.googlePlay.performPlayOperation(twaManifest.packageId,
+          async (editId: string): Promise<boolean> => {
             return await this.googlePlay.versionExists(
                 twaManifest.packageId, versionToRetain, editId);
           });
-      if (!result.versionExistsResult) {
+      if (!exists) {
         throw new Error(enUS.versionDoesNotExistOnServer);
       }
 
--- a/packages/cli/src/lib/cmds/play.ts
+++ b/packages/cli/src/lib/cmds/play.ts
@@ -53,7 +53,12 @@ class Play {
   * @return {number} The largest version number found in the play console.
   */
   async getLargestVersion(twaManifest: TwaManifest): Promise<number> {
-    return await this.googlePlay.getLargestVersionCode(twaManifest.packageId);
+    const result = await this.googlePlay.performPlayOperation(
+        twaManifest.packageId, await this.googlePlay.getLargestVersionCode(twaManifest.packageId));
+    if (result.getLargestVersionCodeResult) {
+      return result.getLargestVersionCodeResult;
+    }
+    throw new Error('Play operation failed. Could not find largest version code.');
   }
 
   /**
@@ -79,8 +84,10 @@ class Play {
     }
 
     const retainedBundles = twaManifest.retainedBundles;
-    await this.googlePlay.publishBundle(
-        userSelectedTrack, publishFilePath, twaManifest.packageId, retainedBundles);
+    await this.googlePlay.performPlayOperation(twaManifest.packageId,
+        await this.googlePlay.publishBundle(
+            userSelectedTrack, publishFilePath, twaManifest.packageId, retainedBundles),
+    );
 
     return true;
   }
@@ -140,7 +147,9 @@ class Play {
                 twaManifest.appVersionCode, versionToRetain));
       }
       // Validate that the version exists on the Play Servers.
-      if (!this.googlePlay.versionExists(twaManifest.packageId, versionToRetain)) {
+      const result = await this.googlePlay.performPlayOperation(twaManifest.packageId,
+          await this.googlePlay.versionExists(twaManifest.packageId, versionToRetain));
+      if (!result.versionExistsResult) {
         throw new Error(enUS.versionDoesNotExistOnServer);
       }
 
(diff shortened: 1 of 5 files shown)


nohe427:
I was also considering add a retained bundle command rather than include it with the Play API. I wanted to know your thoughts on the retain command vs. adding it into the play command as well.

nohe427:
After this is merged, documentation updates will be the next thing I tackle

packages/cli/src/lib/strings.ts
```diff
@@ -406,4 +410,12 @@ the PWA:
       'command to update it:\nbubblewrap updateConfig --androidSdkPath <path-to-sdk>, such that ' +
       'the folder of the path contains the folder "build". Then run bubblewrap doctor again.',
   bothPathsAreValid: 'Your jdkpath and androidSdkPath are valid.',
+  versionDoesNotExistOnServer: 'The supplied version code does not exist on Google Plays Servers.',
```
PEConn, line 413: nit: "on the Google Play servers."

packages/cli/src/lib/cmds/play.ts
```diff
@@ -124,6 +126,49 @@ class Play {
...
 
+    // bubblewrap play --retain 86
+    if (this.args.retain) {
+      const versionToRetain = this.args.retain;
+      // Validate an integer was supplied.
+      if (!Number.isInteger(versionToRetain)) {
+        throw new Error(enUS.versionRetainedNotAnInteger);
+      }
+      if (versionToRetain > twaManifest.appVersionCode) {
+        // Cannot retain a higher version as that would take presendence.
```
PEConn, line 137: nit: precedence

packages/cli/src/lib/cmds/play.ts
```diff
@@ -124,6 +126,49 @@ class Play {
...
+        const approveVersionDifference =
```
PEConn, line 138: I'm not quite sure I understand this - if we *can't* retain a higher version, then what does the user accepting the prompt do?

packages/core/src/lib/GooglePlay.ts
```diff
@@ -70,12 +70,7 @@ export class GooglePlay {
       packageName: string,
       retainedBundles: number[],
   ): Promise<void> {
-    const edit = await this._googlePlayApi.edits.insert({packageName: packageName});
-    const editId = edit.data.id;
-    if (!editId) {
-      // TODO(@nohe427): Try to recover and understand instances where this might fail.
-      throw new Error('Could not create a Google Play edit');
-    }
+    const editId = await this.startPlayOperation(packageName);
```
PEConn, line 73: Do we not need a corresponding `endPlayOperation` call?

packages/core/src/lib/GooglePlay.ts
```diff
@@ -146,24 +141,76 @@ export class GooglePlay {
...
+    }
+    return editId;
+  }
+
+  /**
+   * Cancels the edit in progress on the Play server.
+   * @param packageName - The packageName of the app we are interacting with.
+   * @param editId - The editId that is currently in progress.
+   */
+  private async endPlayOperation(packageName: string, editId: string): Promise<void> {
```
PEConn, line 177: Do we need an `endPlayOperation` for every `startPlayOperation`? If so, we could create a method like `performPlayOperation` that takes a callback, calls `startPlayOperation`, executes the callback (passing in the `editId` and then calls `endPlayOperation`.

That way it's easier for people to not forget the `end` part.

packages/cli/src/lib/cmds/play.ts
nohe427: Yeah good point

packages/core/src/lib/GooglePlay.ts
nohe427: I don't believe so, because these edits are committed in the end.

packages/core/src/lib/GooglePlay.ts
nohe427: I don't think we want to delete the edit on edits that we make a commit call. The delete documentation isn't super clear on this : https://developers.google.com/android-publisher/edits

nohe427: @PEConn - I am a little confused on how to do this in typescript. Each method that I would want to call has a different return type (one is void, one is number, another is bool). I am also interested in knowing how I would append an editId to a callback function that is being passed in.

nohe427: I spoke with someone here, I think i got it.

nohe427:
@PEConn & @andreban - I think I got what one of the reviews was hinting at. Can you all let me know what you think?

andreban:
I was wondering if `performPlayOperation` could work like the following:
```ts
    await this.googlePlay.performPlayOperation(twaManifest.packageId, async (editId: string) => {
      return this.googlePlay.publishBundle(editId, userSelectedTrack, publishFilePath,
          twaManifest.packageId, retainedBundles);
    });
```

So, the signature of `publishBundle()` would be simplified to:
```ts
publishBundle(editId: string track: PlayStoreTrack, filepath: string, packageName: string, retainedBundles: number[]): Promise<PlayOperationResult>>;
```

And `performPlayOperation` looks the same:

```ts
  async performPlayOperation(packageName: string,
      operation: (editId: string) => Promise<PlayOperationResult>): Promise<PlayOperationResult> {
    const editId = await this.startPlayOperation(packageName);
    const result = await operation(editId);
    await this.endPlayOperation(packageName, editId);
    return result;
  }
```

nohe427:
made the suggested changes. lmk what you think.

packages/core/src/lib/GooglePlay.ts
```diff
@@ -145,23 +161,75 @@ export class GooglePlay {
...
     await this._googlePlayApi.edits.delete({editId: editId, packageName: packageName});
+  }
+
+  /**
+   * Checks to see if the version that we want to retain already exists within the Play Store.
+   * @param packageName - The packageName of the versionCode we are looking up.
+   * @param versionCode - The version code of the APK / Bundle we want to retain.
+   *
```
PEConn, line 205: nit: random newline.

packages/core/src/lib/GooglePlay.ts
```diff
@@ -145,23 +161,75 @@ export class GooglePlay {
...
+   */
+  async versionExists(packageName: string, versionCode: number, editId: string):
+      Promise<PlayOperationResult> {
+    const playOpsResult: PlayOperationResult = {};
+
+    const uploadedApks =
+        await this._googlePlayApi.edits.apks.list({packageName: packageName, editId: editId});
+
+    let found = uploadedApks.data.apks?.find(async (obj) => obj.versionCode == versionCode);
```
PEConn, line 214: What does the `async` here do?

packages/cli/src/lib/strings.ts
```diff
@@ -406,4 +410,12 @@ the PWA:
       'command to update it:\nbubblewrap updateConfig --androidSdkPath <path-to-sdk>, such that ' +
       'the folder of the path contains the folder "build". Then run bubblewrap doctor again.',
   bothPathsAreValid: 'Your jdkpath and androidSdkPath are valid.',
+  versionDoesNotExistOnServer: 'The supplied version code does not exist on the Google Play' +
+    'Servers.',
```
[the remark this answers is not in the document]
nohe427: Done

packages/core/src/lib/GooglePlay.ts
```diff
@@ -145,23 +161,75 @@ export class GooglePlay {
...
-    // cleanup
+
+    const playOpsResult: PlayOperationResult = {getLargestVersionCodeResult: versionCode};
+    return playOpsResult;
+  }
+
+  /**
+   * Starts an edit on the Play servers. This is the basis for any play publishing api operation.
+   * @param packageName - the packageName of the app we want to interact with.
+   *
```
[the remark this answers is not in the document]
nohe427: Done

packages/core/src/lib/GooglePlay.ts
nohe427: Done.

packages/core/src/lib/GooglePlay.ts
nohe427: I don't believe it actually does anything. Everything is loaded in from line 209/210. I will remove.

packages/core/src/lib/GooglePlay.ts
```diff
@@ -48,6 +53,20 @@ export class GooglePlay {
     }
   }
 
+  /**
+   * Generates an editId that should be used in all play operations.
+   * @param packageName - The package that will be worked upon.
+   * @param operation - The Play Operation which requires an editId injected
+   * @returns - A PlayOperationResult which contains the output of the operation in a field that was worked upon.
+   */
+  async performPlayOperation(packageName: string,
```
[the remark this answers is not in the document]
nohe427: I think I got this worked out the way we want it.

One note, rather than write the generic function invocation like this:
```
await this.googlePlay.performPlayOperation<number>(...
```

I wrote it like this
```
await this.googlePlay.performPlayOperation(...
```

This okay?

andreban (approved):
LGTM, Thank you!

packages/core/src/lib/GooglePlay.ts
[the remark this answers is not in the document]
PEConn: Yeah, if the compiler can figure it out and it's obvious to anyone reading the code, it's fine to leave this out.

PEConn (approved):

PR #596, opened 2021-09-24
Improves escaping for shell commands

### LINKED ISSUE #587: Keystore initialization fails: password is not escaped
**Describe the bug**

**To Reproduce**
1. `bubblewrap init` without having a keystore
2. when prompted for keystore password, use password which contains `$`, for example `hello$world`
4. The program errors:
```
Signing key creation

An existing key store could not be found at "/home/xxx/tmp/twa/android.keystore".

? Do you want to create one now? Yes
? First and Last names (eg: John Doe): xxxxx
? Organizational Unit (eg: Engineering Dept): xx
? Organization (eg: Company Name): xx
? Country (2 letter code): xx
? Password for the Key Store: ************
? Password for the Key: ************


cli ERROR Command failed: keytool -genkeypair -dname "cn=xxxxx, ou=xx, o=xx, c=xx" -alias "android" -keypass "hello$world" -keystore "/home/xxxxx/tmp/twa/android.keystore" -storepass "hello$world" -validity 20000 -keyalg RSA
```
**Expected behavior**
The `$` is escaped correctly.

**Additional context**
Instead of double quotes, single quotes should be used when supplying the password.

andreban:
The ", $, `, and \ characters still needs to be escaped in double
quoted parameters for shell commands. Those are better handled
in this PR.

Fixes #587

### DIFF
--- a/packages/core/src/lib/jdk/KeyTool.ts
+++ b/packages/core/src/lib/jdk/KeyTool.ts
@@ -15,7 +15,7 @@
  */
 
 import {existsSync, promises} from 'fs';
-import {execute} from '../util';
+import {execute, escapeDoubleQuotedShellString} from '../util';
 import {JdkHelper} from './JdkHelper';
 import {Log, ConsoleLog} from '../Log';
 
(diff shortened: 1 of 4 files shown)


nohe427 (approved):
LGTM

PR #599, opened 2021-09-30
Fixes a retained bundles bug.

nohe427:
This fixes a bug with removing retained bundles from the list of bundles
we want to retain. Previously removeRetained would be called and nothing
would occur to the list of bundles.


andreban (approved):
LGTM, thank you!

PR #601, opened 2021-09-30
Fixes versionExists issue and extra error message.

nohe427:
Removed the asynchronous find method as it was returning true when it
should have returned false. Additionally, added a flag to
performPlayOperation to prevent an error message from appearing when we
commit an edit rather than create a query edit. Also fix duplicate
retain versions from existing in retainBundles.


andreban (approved):
LGTM, thank you!

PR #616, opened 2021-11-05
Update README for Google play feature

nohe427:
These are documentation updates to explain the google play feature and how to use it.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -279,24 +279,51 @@ Flags:
 
 Manages the artifacts for your Google Play Project.
 
+**Note**:
+These feature requires a service account file to work correctly. Please see [this documentation](https://github.com/chromeos/pwa-play-billing#setup-a-service-account) for setting up a service acocunt.
+
+The following options can be applied to all of the features commands:
+  - `--serviceAccountJsonFile`: sets the service account json file location in the twa-manifest.
+  - `--manifest`: specifies the manifest file to use if not in the current directory.
+
+### Subcommands
+
+#### `playPublish`
+
 Usage:
 
 ```
-bubblewrap play --serviceAccountFile="/path/to/service/account.json"  --retain 86 --publish
+bubblewrap playPublish --serviceAccountFile="/path/to/service/account.json"  --track="beta" --appBundleLocation="/home/appBundle.aab"
 ```
 
-This tool requires a service account file to work correctly. Please see [this documentation](https://github.com/chromeos/pwa-play-billing#setup-a-service-account) for setting up a service acocunt.
+Options:
+  - `--track`: publishes the prebuilt file to the Google Play Store specificed track (defaults to internal track).
+  - `--appBundleLocation`: specifies the location of the appbundle to upload to Google Play.
+
+#### `playRetain`
+
+Usage:
+
+```
+bubblewrap playRetain --add=86
+```
 
 Options:
-  - `--publish`: publishes the prebuilt file to the Google Play Store specificed track (defaults to internal track).
-  - `--serviceAccountJsonFile`: sets the service account json file location in the twa-manifest.
-  - `--manifest`: specifies the manifest file to use if not in the current directory.
-  - `--appBundleLocation`: specifies the location of the appbundle to upload to Google Play.
-  - `--targetDirectory`: the directory that versionCheck should run in (defaults to the current directory).
-  - `--versionCheck`: checks that the version specified is higher than the one currently in the Google Play Store.
-  - `--retain`: specifies the bundles to retain for release (this would be Android only bundles if releasing a Chrome OS only release).
-  - `--removeRetained`: removes the specified bundle if no longer relevant.
-  - `--listRetained`: shows a list of existing retained bundles.
+  - `--add`: specifies the bundles to retain for release (this would be Android only bundles if releasing a Chrome OS only release).
+  - `--remove`: removes the specified bundle if no longer relevant.
+  - `--list`: shows a list of existing retained bundles in the twa-manifest.json, not what is listed as retained from play.
+
+#### `playVersionCheck`
+
+Usage:
+
+```
+bubblewrap playVersionCheck --serviceAccountFile="/path/to/service/account.json"  --track="beta"
+```
+
+Options:
+  - `--targetDirectory`: the directory that versionCheck should run in (defaults to the current directory). This should be your bubblewrap project directory.
+  - `--versionCheck`: checks that the version in your app is higher than the one currently in the Google Play Store.
 
 ## `twa-manifest.json` reference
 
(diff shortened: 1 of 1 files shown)


packages/cli/README.md
```diff
@@ -273,6 +273,30 @@ bubblewrap fingerprint generateAssetLinks <flags>
...
+Usage:
+
+```
+bubblewrap play --serivceAccountFile="/path/to/serivce/account"  --retain 86 --publish
+```
+
+This tool requires a service account file to work correctly. Please see [this documentation](https://github.com/chromeos/pwa-play-billing#setup-a-service-account) for setting up a service acocunt.
+
+Options:
+  - `--publish`: publishes the prebuilt file to the Google Play Store under the internal track or specificed track.
```
PEConn, line 291: I think this would read better as "publishes ... to the specified track (defaults to internal)."

packages/cli/README.md
```diff
@@ -273,6 +273,30 @@ bubblewrap fingerprint generateAssetLinks <flags>
...
+## `play`
+
+:warning: This is an experimental feature.
+
+Manages the artifacts for your Google Play Project.
+
+Usage:
+
+```
+bubblewrap play --serivceAccountFile="/path/to/serivce/account"  --retain 86 --publish
```
PEConn, line 285: nit: typo on service (twice).

packages/cli/README.md
```diff
@@ -273,6 +273,30 @@ bubblewrap fingerprint generateAssetLinks <flags>
...
+```
+
+This tool requires a service account file to work correctly. Please see [this documentation](https://github.com/chromeos/pwa-play-billing#setup-a-service-account) for setting up a service acocunt.
+
+Options:
+  - `--publish`: publishes the prebuilt file to the Google Play Store under the internal track or specificed track.
+  - `--serviceAccountJsonFile`: sets the service acocunt json file location in the twa-manifest.
```
PEConn, line 292: nit: typo on "account".

So this command does two things if `serviceAccountJsonFile` is present - it both publishes and saves something to the TWA manifest. Will we ever want to save the thing to the TWA manifest separately? I'm a little wary of commands having side effects of both "override this value" and "write the overwritten value to something".

PEConn: Also, are there multiple valid formats for the service account file? If not then we can probably get rid of the `json`.

packages/cli/README.md
PEConn: Could this also be `/path/to/service/account.json`?

packages/cli/README.md
```diff
@@ -273,6 +273,30 @@ bubblewrap fingerprint generateAssetLinks <flags>
...
+  - `--manifest`: specifies the manifest file to use if not in the current directory.
+  - `--appBundleLocation`: specifies the location of the appbundle to upload to Google Play.
+  - `--targetDirectory`: the directory that versionCheck should run in.
```
PEConn, line 295: Does this have a default?

packages/cli/README.md
```diff
@@ -273,6 +273,30 @@ bubblewrap fingerprint generateAssetLinks <flags>
...
+  - `--versionCheck`: checks that the version specified is higher than the one currently in the Google Play Store.
+  - `--retain`: specifies the bundles to retain for release (this would be Android only bundles if releasing a Chrome OS only release).
+  - `--removeRetained`: removes the specified bundle if no longer relevant.
+  - `--listRetained`: shows a list of existing retained bundles.
```
PEConn, line 299: Can we split this out into a different command?

packages/cli/README.md
nohe427: Done

packages/cli/README.md
nohe427: Done.

packages/cli/README.md
nohe427: Fixed the account typo.

This never overrides the value, but always sets the serviceAccountFile in the twa-manifest.json to be used by the command. Do you think it would be worthwhile to have a save command or something else to compliment this so we can overwrite in some instances and override in others?

Yes, p12 keys also exist for the serviceAccountFile. I did not implement any handling for that format.

packages/cli/README.md
nohe427: So something like `bubblewrap retained --${command}`?

PEConn: Thinking about it a bit more, it seems that we have one `bubblewrap play` command that can do a bunch of different things. In fact most of the bubblewrap commands are `bubblewrap <verb>` (such as `bubblewrap build`, `bubblewrap update`, `bubblewrap validate`) which I think makes sense. `bubblewrap play` on the other hand doesn't tell us what the command is going to do (it connects to Play and does what?).

It would be good to split out which of these options determines what `bubblewrap play` actually does, and which are just configuration options. I think maybe it would work better to have something like the following commands:

```
bubblewrap playPublish
bubblewrap playListRetained
bubblewrap playRemoveRetained
```

And then each of these commands would have `serviceAccountFile`, `manifest`, etc options.

Bear in mind that I can't quite remember 100% what each of these options does - so maybe `playRemoveRetained` actually doesn't do anything on it's own and having a `--removeRetained` option for `playPublish` makes sense. Similarly, maybe `versionCheck` could be a separate command instead of a flag for the `playPublish` command.

PEConn: I still think we should try to simplify the `playPublish` command if possible, but yeah, I think that looks like a good path forward.

nohe427: Ok, cool. I will go and update the docs in this CL and then do a few small CLs to cover the playAction commands using the docs as a guide.

packages/cli/README.md
```diff
@@ -279,24 +279,51 @@ Flags:
 
 Manages the artifacts for your Google Play Project.
 
+**Note**:
+These feature requires a service account file to work correctly. Please see [this documentation](https://github.com/chromeos/pwa-play-billing#setup-a-service-account) for setting up a service acocunt.
```
andreban, line 283:
```suggestion
These feature requires a service account file to work correctly. Please see [this documentation](https://github.com/chromeos/pwa-play-billing#setup-a-service-account) for setting up a service account.
```

andreban (changes requested):
Looking good. There's a typo to fix.

andreban (approved):
LGTM, Thank you!

packages/cli/README.md
```diff
@@ -273,6 +273,56 @@ bubblewrap fingerprint generateAssetLinks <flags>
...
+  - `--manifest`: specifies the manifest file to use if not in the current directory.
+
+### Subcommands
+
+#### `playPublish`
+
+Usage:
+
+```
+bubblewrap playPublish --serviceAccountFile="/path/to/service/account.json"  --track="beta" --appBundleLocation="/home/appBundle.aab"
```
PEConn, line 296: nit: random double space before `--track`.

PEConn (approved):
LGTM % nit

PR #620, opened 2021-11-11
Adds the ArCore feature

### LINKED ISSUE #515: Add the ARcore permission to the Android Manifest when using WebXR
I created a TWA with PWABuilder. Works so far as expected.

What I just realized is, that the permissions aren't set correctly. The location permission is picked up correctly, but as my app uses WebXR, camera and AR requirements aren't picked up. Reported this on PWABuilder repo but was sent here.

The problem with the missing permission is, that the Play store can't filter the app for devices not capable to run the app. This will likely be frustrating to the user, when the user is told that the app isn't running on the device after installation.

URL to app
sparcl.app

These requirements should likely lead to camera permission and AR feature added to the android manifest. According to ARCore docs, that would look like this:

<uses-permission android:name="android.permission.CAMERA" />

<uses-feature android:name="android.hardware.camera.ar" />

<application …>
    …

    
    <meta-data android:name="com.google.ar.core" android:value="required" />
</application>

Adding the permissions to the Android manifest can probably done when using bubble wrap directly. But would appreciate when this could be handled for PWABuilder, too.

**Additional context**
Camera and AR permissions not included in the generated app package

image

andreban:
- Adds a feature that appends the AR Core requirement to the
  AndroidManifest, ensuring the app is only deployed to devices that
  support ARCore.

Closes #515

### DIFF
--- a/packages/core/src/lib/TwaManifest.ts
+++ b/packages/core/src/lib/TwaManifest.ts
@@ -27,6 +27,7 @@ import {AppsFlyerConfig} from './features/AppsFlyerFeature';
 import {LocationDelegationConfig} from './features/LocationDelegationFeature';
 import {PlayBillingConfig} from './features/PlayBillingFeature';
 import {FirstRunFlagConfig} from './features/FirstRunFlagFeature';
+import {ArCoreConfig} from './features/ArCoreFeature';
 
 // The minimum size needed for the app icon.
 const MIN_ICON_SIZE = 512;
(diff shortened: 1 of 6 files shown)


PEConn (approved):
LGTM, thanks!

PEConn:
We probably need to document this feature somewhere.

pranjaldugar:
@andreban Hi, I have a question to ask,
Don't we need such feature for Camera, Microphone and other sensors?
like when we make traditional app we ```<uses-permission android:name="android.permission.CAMERA" />```,
my TWA does use Camera and other web apis however the they are not shown in App permissions section of Play Store.

Thanks!

andreban:
The Trusted Web Activity runs in the Browser, so the browser permission (and the permissions the browser manages for the sites) will apply instead.

Issue #619, opened 2021-11-09
Migrate play command to playPublish command


nohe427:
This migrates the existing play command in the playExpirement branch to have clearer naming and prevents one command from doing lots of things. In this instance, it will just run operations related to publishing to Google Play, not for version checking or retaining bundles.

Issue #618, opened 2021-11-09
Google Play Integration


nohe427:
**Is your feature request related to a problem? Please describe.**
An easier path for a completed Bubblewrap project to be uploaded to an existing Google Play Store listing.

**Describe the solution you'd like**
`bubblewrap play ${cmd}`

**Describe alternatives you've considered**
none

**Additional context**
already inflight on the `playExperiment` branch.

Current Work to be completed:

- [x] #619
- [x] #622
- [x] #623
- [ ] Fix retain command to ensure no duplicates exist within the manifest.

Once these are completed, the Google Play Integration could be considered for merging into the main branch.

andreban:
@nohe427 it seems the PR for this has been merged. Can this issue be resolved?

PR #621, opened 2021-11-15
Create playPublish command.

nohe427:
This creates the playPublish command rather than the all in one play
command. This is a series of changes to complete migrating the commands
out to a less confusing state for our users.
See Google Play Integration #618

Fixes: Migrate play command to playPublish command #619

### DIFF
--- a/packages/cli/src/lib/Cli.ts
+++ b/packages/cli/src/lib/Cli.ts
@@ -92,7 +92,7 @@ export class Cli {
       // case 'play':
       //   return await play(parsedArgs as unknown as PlayArgs);
       // case 'playPublish':
-      //   return await playPublish(parsedArgs as unknown as PlayArgs);
+      //   return await play(parsedArgs as unknown as PlayArgs, 'publish');
       default:
         throw new Error(
             `"${command}" is not a valid command! Use 'bubblewrap help' for a list of commands`);
--- a/packages/cli/src/lib/Cli.ts
+++ b/packages/cli/src/lib/Cli.ts
@@ -31,6 +31,7 @@ import {merge} from './cmds/merge';
 import {fingerprint} from './cmds/fingerprint';
 // import {play, PlayArgs} from './cmds/play';
 import {fetchUtils} from '@bubblewrap/core';
+// import { play, PlayArgs, playPublish } from './cmds/play';
 
 export class Cli {
   async run(args: string[]): Promise<boolean> {
@@ -90,6 +91,8 @@ export class Cli {
         return await fingerprint(parsedArgs);
       // case 'play':
       //   return await play(parsedArgs as unknown as PlayArgs);
+      // case 'playPublish':
+      //   return await playPublish(parsedArgs as unknown as PlayArgs);
       default:
         throw new Error(
             `"${command}" is not a valid command! Use 'bubblewrap help' for a list of commands`);
--- a/packages/cli/src/lib/cmds/play.ts
+++ b/packages/cli/src/lib/cmds/play.ts
@@ -37,6 +37,8 @@ export interface PlayArgs {
 // Default file path
 const defaultSignedAppBundleFileName = 'app-release-bundle.aab';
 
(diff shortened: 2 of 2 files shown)


packages/cli/src/lib/cmds/play.ts
```diff
@@ -232,10 +238,24 @@ async function setupGooglePlay(args: PlayArgs): Promise<GooglePlay> {
...
+  // TODO(@nohe427): Remove after experimental
+  if (!await prompt.promptConfirm(enUS.promptExperimentalFeature, false)) {
+    return true;
+  }
+  const googlePlay = await setupGooglePlay(parsedArgs);
+  const play = new Play(parsedArgs, googlePlay, prompt);
+  return await play.runPlayPublish();
+}
+
 export async function play(parsedArgs: PlayArgs,
```
nohe427, line 252: Would it be better to set this function to have an enum as an argument for if we want playX to run rather than having three methods? i.e., 
```
play(args, prompt, method) {
if method == publish {
  runPlayPublish()
}

File: packages/cli/src/lib/Cli.ts [hunk 2]
```diff
@@ -31,6 +31,7 @@ import {merge} from './cmds/merge';
 import {fingerprint} from './cmds/fingerprint';
 // import {play, PlayArgs} from './cmds/play';
```
andreban, line 32: Should this line be removed?

packages/cli/src/lib/Cli.ts
```diff
@@ -90,6 +91,8 @@ export class Cli {
         return await fingerprint(parsedArgs);
       // case 'play':
       //   return await play(parsedArgs as unknown as PlayArgs);
```
andreban, line 93: And those 2 as well?

packages/cli/src/lib/cmds/play.ts
```diff
@@ -68,7 +68,7 @@ class Play {
   async publish(twaManifest: TwaManifest): Promise<boolean> {
     // Validate that the publish value is listed in the available Tracks.
     // If no value was supplied with publish we make it internal.
-    const userSelectedTrack = asPlayStoreTrack(this.args.publish?.toLowerCase() || 'internal');
+    const userSelectedTrack = asPlayStoreTrack(this.args.track?.toLowerCase() || 'internal');
```
andreban, line 71: Is there a fixed list of values that can be passed as a track? Or can any string be used?

packages/cli/src/lib/cmds/play.ts
andreban: You can use something like

```ts
type PlayCommand = 'publish' | 'retain' | 'versionCheck';
```
then
```ts
function play(args: PlayArgs, prompt: Prompt = new InquirerPrompt(), command: PlayCommand) {
...
}
```

andreban (approved):
LGTM with a couple nits

packages/cli/src/lib/Cli.ts
nohe427: No, we are keeping it for now so we can just uncomment when we want to test the feature in the wild. This also lets anyone that wants to clone the repo and try out the feature do so early without too much modification to the code base.

packages/cli/src/lib/Cli.ts
nohe427: Same as above

packages/cli/src/lib/cmds/play.ts
nohe427: Its a list of values. They can be found [here](https://github.com/GoogleChromeLabs/bubblewrap/blob/069e7444cf8c71c6ba4f844eff731c3eb97726bc/packages/core/src/lib/GooglePlay.ts#L21)

packages/cli/src/lib/cmds/play.ts
nohe427: Think I got this all resolved now. I changed the positioning of the args though since I think the prompt is an optional arg.

nohe427:
@andreban - Want to take another look before I merge?

Issue #622, opened 2021-11-15
Create playRetain command


nohe427:

nohe427:
Part of #618 work

PR #630, opened 2021-12-01
Migrate play --retain to playRetain command.

nohe427:
This migrates the retaining capability into its own command.

Fixes Create playRetain command #622


G8342 (approved):

PEConn (approved):
Thanks!

Issue #552, opened 2021-06-17
PWA Builder CloudAPK (Bubblewrap) zipalign issue "Failed. Status code 500, Error: Internal Server Error, Details: Error generating app package"


bryantvu:
Hello,

I'm trying to get PWA Builder CloudAPK running.
https://github.com/pwa-builder/CloudAPK

I'm getting an error generating an Android package with just the default settings. PWA Builder team did note that I was using the correction SDK version (30.0.3) and suggested that I seek help from the Bubblewrap team instead. Any ideas on what's going on here?
``
Failed. Status code 500, Error: Internal Server Error, Details: Error generating app package: Error: Command failed: "/home/ubuntu/api/CloudAPK/android-sdk/build-tools/30.0.3/zipalign" -v -f -p 4 /tmp/pwabuilder-cloudapk--24177-thSWAhVabhJC/app/build/outputs/apk/release/app-release-unsigned.apk /tmp/pwabuilder-cloudapk--24177-thSWAhVabhJC/app-release-unsigned-aligned.apk
``

andreban:
Recent versions of the Gradle plugin will automatically zipalign the APK, and calling zipalign in those cases is redundant. See [this issue](https://b.corp.google.com/issues/162117652#comment12). To reflect this, the Bubblewrap CLI only [verifies the APK is zipaligned](https://github.com/GoogleChromeLabs/bubblewrap/blob/d8019d0aef42d64b0590bf223d22b4ad0f5e0575/packages/cli/src/lib/cmds/build.ts#L117-L123). ClouldAPK seems to be still calling `zipalign` [here](https://github.com/pwa-builder/CloudAPK/blob/dbe52dda058e808b9f6f694095ffd0c4b05ebf12/build/bubbleWrapper.ts#L136-L145)

bryantvu:
Thanks for getting back so quickly Andre, I've informed the PWAB team of this change.

JudahGabriel:
FYI, we've updated our side of this as well. I believe you can close this.

PR #639, opened 2022-01-13
Set a default user-agent

nohe427:
This changes the user agent that is being sent to sites to match that of
a browser. Some sites were blocking our requests because of the default
user agent. node-fetch has some issues when attempting to set the user
agent, so this was left alone for now, especially because in #552 there
is a mention of potentially removing the user agent.

### DIFF
--- a/packages/core/src/lib/FetchUtils.ts
+++ b/packages/core/src/lib/FetchUtils.ts
@@ -15,13 +15,17 @@
  */
 
 import * as fs from 'fs';
-import {fetch as fetchh2, Response as FetchH2Response} from 'fetch-h2';
+import {context, Response as FetchH2Response} from 'fetch-h2';
 import nodefetch, {Response as NodeFetchResponse} from 'node-fetch';
 
 export type FetchEngine = 'node-fetch' | 'fetch-h2';
 const DEFAULT_FETCH_ENGINE: FetchEngine = 'fetch-h2';
 
 export type NodeFetchOrFetchH2Response = FetchH2Response | NodeFetchResponse;
+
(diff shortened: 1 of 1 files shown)


packages/core/src/lib/FetchUtils.ts
```diff
@@ -15,14 +15,18 @@
...
 import nodefetch, {Response as NodeFetchResponse} from 'node-fetch';
 
 export type FetchEngine = 'node-fetch' | 'fetch-h2';
 const DEFAULT_FETCH_ENGINE: FetchEngine = 'fetch-h2';
 
 export type NodeFetchOrFetchH2Response = FetchH2Response | NodeFetchResponse;
 
+const userAgent = 'Mozilla/5.0 (X11; Linux x86_64; rv:60.0) Gecko/20100101 Firefox/81.0';
+const fetchh2Ctx = context({userAgent: userAgent, overwriteUserAgent: true});
+const fetchh2 = fetchh2Ctx.fetch;
```
PEConn, line 28: Are `fetchh2Ctx` and `fetchh2` used anywhere?

nohe427: Yes, line 41 is where `fetchh2` is used and `fetchh2Ctx` is used to generate the fetch call. Maybe we should collapse line 27 and 28 together?

PEConn: Oh right, you were previously importing something as `fetchh2` but now you're creating a variable of the same name. Sorry I missed that

PEConn (approved):

PR #674, opened 2022-04-23
Escape gradle string in shareTarget / build.gradle

### LINKED ISSUE #673: Share Target with single quote throws an error
The [issue](https://github.com/pwa-builder/PWABuilder/issues/2812) was raised on the [PWABuilder](https://pwabuilder.com/) repo. We use bubblewrap to package TWAs. It seems to throw the following error every time we try to create a package, and it's due to the french characters being used in the apps manifest, specifically in the description:

> "description": "La Chorale de Cocody Philadelphie est une chorale l'Eglise Adventiste du 7ième jour de Cocody Philadelphie. Elle chante pour la gloire de DIEU et pour gagner des âmes à Christ.",

cli ERROR Command failed: gradlew.bat assembleRelease --stacktrace

FAILURE: Build failed with an exception.

* Where:
Build file 'C:\Users\jaylynbarbee\OneDrive - Microsoft\Desktop\bbwrap-test\app\build.gradle' line: 80

* What went wrong:
Could not compile build file 'C:\Users\jaylynbarbee\OneDrive - Microsoft\Desktop\bbwrap-test\app\build.gradle'.
> startup failed:
build file 'C:\Users\jaylynbarbee\OneDrive - Microsoft\Desktop\bbwrap-test\app\build.gradle': 80: unexpected char: '\' @ line 80, column 430.
pour gagner des �mes � Christ.\\",\\"url

**To Reproduce**
1. bubblewrap init --manifest https://adventiste-ccp.web.app/manifest.webmanifest
2. bubblewrap build

**Expected behavior**
Package generated with no error.

andreban:
Escapes the gradle string before escaping the JSON. This ensures single quotes
are escaped.

Fixes #673

### DIFF
--- a/packages/core/src/spec/lib/utilSpec.ts
+++ b/packages/core/src/spec/lib/utilSpec.ts
@@ -306,13 +306,14 @@ describe('util', () => {
   });
 
   describe('#escapeJsonString combined with #escapeGradleString returns the expected results',
-   () => {
-    it ('Escapes double and single quotes', () => {
-      // String.raw prevents escape characters from being applied, so we can simplify howe we write
-      // the expected value.
-      let expected = String.raw`\\"Hello Worl\\\'d\\"`;
-      expect(util.escapeJsonString(util.escapeGradleString(`"Hello Worl'd"`)))
-          .toEqual(expected);
-    });
-  });
+      () => {
+        it('Escapes double and single quotes', () => {
+          // String.raw prevents escape characters from being applied, so we can simplify howe we write
+          // the expected value.
+          const input = String.raw`"Hello Worl'd"`;
+          const expected = String.raw`\\"Hello Worl\\\'d\\"`;
+          expect(util.escapeJsonString(util.escapeGradleString(input)))
+              .toEqual(expected);
+        });
+      });
 });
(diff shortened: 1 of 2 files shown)


PEConn (commented):
Where does this method come from? I can't find the docs for it.

I feel uneasy about escaping a string twice - how do we know this won't do something crazy?

andreban:
It's set here: https://github.com/GoogleChromeLabs/bubblewrap/blob/main/packages/core/src/lib/TwaGenerator.ts#L404

PEConn:
Yeah, that would be good, thanks.

andreban:
> Yeah, that would be good, thanks.

Added a test. PTAL

packages/core/src/spec/lib/utilSpec.ts
```diff
@@ -304,4 +304,16 @@ describe('util', () => {
           .toEqual('\\"\\$Hello\\\\Wo\\`eld\\"');
     });
   });
+
+  describe('#escapeJsonString combined with #escapeGradleString returns the expected results',
+      () => {
+        it('Escapes double and single quotes', () => {
+          // String.raw prevents escape characters from being applied, so we can simplify howe we write
```
PEConn, line 311: nit: typo on "how"

PEConn (approved):
LGTM % nit thanks

packages/core/src/spec/lib/utilSpec.ts
andreban: fixed

PR #675, opened 2022-04-23
Remove criteria validation from the build step

andreban:
Removes the criteria validation from the build step for two reasons:

- The TWA criteria has changed, so the validation criteria doesn't
  reflect the truth anymore.
- Additionally, the PSI/Lighthouse endpoint frequently fails to return
  results.

We should also consider deprecating the validation command and package.


PEConn (approved):
LGTM % cleaning up the import (that GitHub Actions reported).

andreban:
Cleaned the import. Github Actions is happy :D

Issue #672, opened 2022-04-21
Package ID with "new" keyword throws an error


amrutha95:
**Describe the bug**
The [issue](https://github.com/pwa-builder/PWABuilder/issues/2805) was raised on the [PWABuilder](https://pwabuilder.com) repo. We use bubblewrap to package TWAs. It seems to throw the following error every time we try to create a package, and it's due to the "new" keyword being part of the name: 

Execution failed for task ':app:processReleaseResources'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.Workers$ActionFacade
   > **Package 'at.homeschooling4kids.new.twa' from AndroidManifest.xml is not a valid Java package name as 'new' is a Java keyword.**

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':app:processReleaseResources'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:207)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:263)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:205)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:186)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:114)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
        at ......

* Get more help at https://help.gradle.org

BUILD FAILED in 3m 58s

**To Reproduce**
1. bubblewrap init --manifest=https://new.homeschooling4kids.at/_nuxt/manifest.ca354309.json
2. bubblewrap build

**Expected behavior**
Package generated with no error.

PEConn:
We should add some checks in the `bubblewrap init` flow to catch this type of thing.

You can work around it by overriding the default `Application ID` in the `bubblewrap init` flow to be something that doesn't contain "new" (or any other Java keyword).

andreban:
[...]

PR #676, opened 2022-04-23
Handle Java keywords for applicatoinIds

andreban:
- Check for Java keywords when testing.
- Add _ before a keyword when generating.

See #672 for details

### DIFF
--- a/packages/core/src/lib/util.ts
+++ b/packages/core/src/lib/util.ts
@@ -37,12 +37,12 @@ const VALID_PACKAGE_ID_SEGMENT_REGEX = /^[a-zA-Z][A-Za-z0-9_]*$/;
 // List of keywords for Java 11, as listed at
 // https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-3.9.
 const JAVA_KEYWORDS = [
-  'abstract', 'continue', 'for', 'new', 'switch', 'assert', 'default', 'if', 'package',
-  'synchronized', 'boolean', 'do', 'goto', 'private', 'this', 'break', 'double', 'implements',
-  'protected', 'throw', 'byte', 'else', 'import', 'public', 'throws', 'case', 'enum', 'instanceof',
-  'return', 'transient', 'catch', 'extends', 'int', 'short', 'try', 'char', 'final', 'interface',
-  'static', 'void', 'class', 'finally', 'long', 'strictfp', 'volatile', 'const', 'float', 'native',
-  'super', 'while',
+  'abstract', 'assert', 'boolean', 'break', 'byte', 'case', 'catch', 'char', 'class', 'const',
+  'continue', 'default', 'do', 'double', 'else', 'enum', 'extends', 'final', 'finally', 'float',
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/util.ts
```diff
@@ -34,6 +34,17 @@ const extractZipPromise = promisify(extractZip);
 const DISALLOWED_ANDROID_PACKAGE_CHARS_REGEX = /[^a-zA-Z0-9_\.]/g;
 const VALID_PACKAGE_ID_SEGMENT_REGEX = /^[a-zA-Z][A-Za-z0-9_]*$/;
 
+// List of keywords for Java 11, as listed at
+// https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-3.9.
+const JAVA_KEYWORDS = [
```
PEConn, line 39: Sorry, but do you mind sorting these?

packages/core/src/lib/util.ts
```diff
@@ -146,6 +157,14 @@ export function generatePackageId(host: string): string | null {
     if (part.trim().length === 0) {
       continue;
     }
+
+    // Package names cannot contain Java keywords. The recommendation is adding an '_' before the
+    // keyword. See https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html.
+    if (JAVA_KEYWORDS.indexOf(part) >= 0) {
```
PEConn, line 163: Can't we just use `JAVA_KEYWORDS.includes(part)`?

packages/core/src/lib/util.ts
```diff
@@ -195,6 +214,13 @@ export function validatePackageId(input: string): string | null {
   }
 
   for (const part of parts) {
+    // Package names cannot contain Java keywords. The recommendation is adding an '_' before the
+    // keyword. See https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html.
+    if (JAVA_KEYWORDS.indexOf(part) >= 0) {
```
PEConn, line 219: Same as above.

PEConn (approved):
LGTM % nits

packages/core/src/lib/util.ts
andreban: Oh, silly me... I remembered there was a handy method in JS Array, but couldn't remember the name. I checked the docs, but couldn't find what I was looking for (I think I was looking for something like `contains`), and missed `includes`... I'll change that.

packages/core/src/lib/util.ts
andreban: LOL. I actually thought they were sorted, because I copied from the docs... but the docs use column-first sorting and this is using row first... I'll go change that.

andreban: done

packages/core/src/lib/util.ts
andreban: done

packages/core/src/lib/util.ts
andreban: likewise

PR #677, opened 2022-04-23
Fix tests for Doctor

andreban:
The tests in DoctorSpec seemed to be causing the test suite to abort,
and broken tests weren't being properly reported. My guess is that
many of the tests in DoctorSpec have actually never worked.

This PR fixes DoctorSpec and other tests that were broken.


PEConn (approved):
Well that is embarrassing... LGTM

andreban:
FWIW, I find it annoying that the test suite silently fails.

PR #685, opened 2022-04-28
Skips BOM on Web Manifests

andreban:
- Downloads manifest as String and trims before parsing the JSON.


andreban:
Yeah, if I get more places where I need it I'll wrap in a function.

PR #704, opened 2022-06-14
Fix #615 Add Meta Quest support

### LINKED ISSUE #615: [Research] Generate Oculus Store compatible APKs
**Initial request**

https://twitter.com/ThePra9/status/1455899334365618181

**Current plan**

1. Check if APK generated by Oculus CLI uses TWA (Trusted Web Activity):
1.1. Generate APK using Oculus CLI:
```sh
./ovr-platform-util create-pwa
  -o output.apk
  --android-sdk /opt/android_sdk
  https://example.com/manifest.webmanifest
```
Or to use local manifest file:
```sh
--manifest-content-file manifest.webmanifest
  https://example.com/
```
2. Check if Oculus supports TWA (Trusted Web Activity):
2.1. Generate APK using Bubblewrap (PWABuilder), try to publish it to Oculus Store via Oculus CLI or web interface (see https://developer.oculus.com/resources/publish-reference-platform-command-line-utility/#upload-quest) and make some changes to APK source code to fix errors/warnings (see https://developer.oculus.com/resources/publish-mobile-manifest/#specs).
2.2. Compare APK generated by Oculus CLI with APK generated by Bubblewrap (`AndroidManifest.xml`, etc.).
2.3. Check APK with fixes published to Oculus Store on real Oculus device.

3. Create a pull request for Bubblewrap.

4. Create a pull request for PWABuilder.

5. Write docs. E.g. document new vendor-specific proprietary web app manifest members such as:

```json
"ovr_scope_extensions": [
  {"pattern": "*.example.com"}
]
```

and

```json
"ovr_multi_tab_enabled" : true
```

**Notes**

This should be an option to not affect APKs for other app stores.

FluorescentHallucinogen:
This pull request adds a `--metaquest` flag for generating _universal_ APKs compatible with Meta Quest devices that open Trusted Web Activity (TWA) on regular Android devices and open Meta Quest Browser on Meta Quest devices. This will allow to abandon `ovr-platform-util` in favor of Bubblewrap (e.g. in PWABuilder).

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -372,7 +372,8 @@ Fields:
 |startUrl|string|true|The start path for the TWA. Must be relative to the domain.|
 |themeColor|string|true|The color used for the status bar.|
 |webManifestUrl|string|false|Full URL to the PWA Web Manifest. Required for the application to be compatible with Chrome OS and Meta Quest devices.|
-|fullScopeUrl|string|false|The navigation scope that the browser considers to be within the app. If the user navigates outside the scope, it reverts to a normal web page inside a browser tab or window. Must be a full URL. Required only for Meta Quest devices.|
+|fullScopeUrl|string|false|The navigation scope that the browser considers to be within the app. If the user navigates outside the scope, it reverts to a normal web page inside a browser tab or window. Must be a full URL. Required and used only by Meta Quest devices.|
+|minSdkVersion|number|false|The minimum [Android API Level](https://developer.android.com/guide/topics/manifest/uses-sdk-element#ApiLevels) required for the application to run. Defaults to `23`, if `isMetaQuest` is `true`, and `19` otherwise.|
 
 ### Features
 
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -98,7 +98,7 @@ bubblewrap init --manifest="<web-manifest-url>" [--directory="<path-to-output-lo
 
 Options:
   - `--directory`: path where to generate the project. Defaults to the current directory.
-  - `--chromeosonly`: this flag specifies that the build will be used for Chrome OS only and prevents non-Chrome OS devices (except Meta Quest devices) from installing the app.
+  - `--chromeosonly`: this flag specifies that the build will be used for Chrome OS only and prevents non-Chrome OS devices from installing the app.
   - `--metaquest`: this flag specifies that the build will be compatible with Meta Quest devices.
   - `--alphaDependencies`: enables features that depend on upcoming version of the Android library
   for Trusted Web Activity or that are still unstable.
@@ -351,7 +351,7 @@ Fields:
 |generatorApp|string|false|Identifier for tool used to generate the Android project. Bubblewrap uses `bubblewrap-cli`. Should only be modified by generator apps.|
 |host|string|true|The origin that will be opened in the Trusted Web Activity.|
 |iconUrl|string|true|Full URL to an the icon used for the application launcher and splash screen. Must be at least 512x512 px.|
-|isChromeOSOnly|boolean|false|Generates an application that targets only ChromeOS devices (except Meta Quest devices). Defaults to `false`.|
+|isChromeOSOnly|boolean|false|Generates an application that targets only ChromeOS devices. Defaults to `false`.|
 |isMetaQuest|boolean|false|Generates an application that compatible with Meta Quest devices. Defaults to `false`.|
 |launcherName|string|false|A short name for the Android application, displayed on the Android launcher|
 |maskableIconUrl|string|false|Full URL to an the icon used for maskable icons, when supported by the device.|
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -372,7 +372,9 @@ Fields:
 |startUrl|string|true|The start path for the TWA. Must be relative to the domain.|
 |themeColor|string|true|The color used for the status bar.|
 |webManifestUrl|string|false|Full URL to the PWA Web Manifest. Required for the application to be compatible with Chrome OS and Meta Quest devices.|
- 
+|scope|string|false|The navigation scope that the browser considers to be within the app. If the user navigates outside the scope, it reverts to a normal web page inside a browser tab or window. Only used to generate the `fullScopeUrl`.|
+|fullScopeUrl|string|false|The navigation scope that the browser considers to be within the app. If the user navigates outside the scope, it reverts to a normal web page inside a browser tab or window. Must be full URL. Required only for Meta Quest devices.|
+
 ### Features
 
 Developers can enable additional features in their Android application. Some features may include more dependencies into the application and increase the binary size.
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -98,7 +98,7 @@ bubblewrap init --manifest="<web-manifest-url>" [--directory="<path-to-output-lo
 
 Options:
   - `--directory`: path where to generate the project. Defaults to the current directory.
-  - `--chromeosonly`: this flag specifies that the build will be used for Chrome OS only and prevents non-Chrome OS devices from installing the app.
+  - `--chromeosonly`: this flag specifies that the build will be used for Chrome OS only and prevents non-Chrome OS devices (except Meta Quest devices) from installing the app.
   - `--metaquest`: this flag specifies that the build will be compatible with Meta Quest devices.
   - `--alphaDependencies`: enables features that depend on upcoming version of the Android library
   for Trusted Web Activity or that are still unstable.
@@ -351,7 +351,7 @@ Fields:
 |generatorApp|string|false|Identifier for tool used to generate the Android project. Bubblewrap uses `bubblewrap-cli`. Should only be modified by generator apps.|
 |host|string|true|The origin that will be opened in the Trusted Web Activity.|
 |iconUrl|string|true|Full URL to an the icon used for the application launcher and splash screen. Must be at least 512x512 px.|
-|isChromeOSOnly|boolean|false|Generates an application that targets only ChromeOS devices. Defaults to `false`.|
+|isChromeOSOnly|boolean|false|Generates an application that targets only ChromeOS devices (except Meta Quest devices). Defaults to `false`.|
 |isMetaQuest|boolean|false|Generates an application that compatible with Meta Quest devices. Defaults to `false`.|
 |launcherName|string|false|A short name for the Android application, displayed on the Android launcher|
 |maskableIconUrl|string|false|Full URL to an the icon used for maskable icons, when supported by the device.|
@@ -371,7 +371,7 @@ Fields:
 |splashScreenFadeOutDuration|number|true|Duration for the splash screen fade out animation.|
 |startUrl|string|true|The start path for the TWA. Must be relative to the domain.|
 |themeColor|string|true|The color used for the status bar.|
-|webManifestUrl|string|false|Full URL to the PWA Web Manifest. Required for the application to be compatible with Chrome OS devices.|
+|webManifestUrl|string|false|Full URL to the PWA Web Manifest. Required for the application to be compatible with Chrome OS and Meta Quest devices.|
  
 ### Features
 
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -93,12 +93,13 @@ parse the Web manifest and generate default values for the Android project, wher
 will prompt the user to confirm or input values where one could not be generated.
 
 ```
-bubblewrap init --manifest="<web-manifest-url>" [--directory="<path-to-output-location>"] [--chromeosonly]
+bubblewrap init --manifest="<web-manifest-url>" [--directory="<path-to-output-location>"] [--chromeosonly] [--metaquest]
 ```
 
 Options:
   - `--directory`: path where to generate the project. Defaults to the current directory.
   - `--chromeosonly`: this flag specifies that the build will be used for Chrome OS only and prevents non-Chrome OS devices from installing the app.
+  - `--metaquest`: this flag specifies that the build will be compatible with Meta Quest devices.
   - `--alphaDependencies`: enables features that depend on upcoming version of the Android library
   for Trusted Web Activity or that are still unstable.
 
@@ -351,6 +352,7 @@ Fields:
 |host|string|true|The origin that will be opened in the Trusted Web Activity.|
 |iconUrl|string|true|Full URL to an the icon used for the application launcher and splash screen. Must be at least 512x512 px.|
 |isChromeOSOnly|boolean|false|Generates an application that targets only ChromeOS devices. Defaults to `false`.|
+|isMetaQuest|boolean|false|Generates an application that compatible with Meta Quest devices. Defaults to `false`.|
 |launcherName|string|false|A short name for the Android application, displayed on the Android launcher|
 |maskableIconUrl|string|false|Full URL to an the icon used for maskable icons, when supported by the device.|
 |monochromeIconUrl|string|false|Full URL to a monochrome icon, used when displaying notifications.|
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -372,8 +372,7 @@ Fields:
 |startUrl|string|true|The start path for the TWA. Must be relative to the domain.|
 |themeColor|string|true|The color used for the status bar.|
 |webManifestUrl|string|false|Full URL to the PWA Web Manifest. Required for the application to be compatible with Chrome OS and Meta Quest devices.|
-|scope|string|false|The navigation scope that the browser considers to be within the app. If the user navigates outside the scope, it reverts to a normal web page inside a browser tab or window. Only used to generate the `fullScopeUrl`.|
-|fullScopeUrl|string|false|The navigation scope that the browser considers to be within the app. If the user navigates outside the scope, it reverts to a normal web page inside a browser tab or window. Must be full URL. Required only for Meta Quest devices.|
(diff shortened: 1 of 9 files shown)


packages/cli/src/lib/cmds/init.ts
```diff
@@ -245,6 +246,10 @@ export async function init(
     twaManifest.isChromeOSOnly = true;
   }
 
+  if (args.metaquest) {
```
andreban, line 249: I'm wondering if passing `--chromeosonly` and `--metaquest` together should throw an error, or maybe a warning, since setting Chrome OS only will effectively make the APK incompatible with Meta Quest.

FluorescentHallucinogen: The `--chromeosonly` and `--metaquest` options should be fully compatible. Meta Quest uses only the `AndroidManifest.xml` from the APK file. All unknown things in `AndroidManifest.xml` are simply ignored. It seems to be the same for Chrome OS. But I can check again to be sure.

packages/core/template_project/app/src/main/AndroidManifest.xml
```diff
@@ -61,6 +76,20 @@
...
+            android:name="com.oculus.pwa.NAME"
+            android:value="@string/appName" />
+
+        <meta-data
+            android:name="com.oculus.pwa.START_URL"
+            android:value="@string/launchUrl" />
+
+        <meta-data
+            android:name="com.oculus.pwa.SCOPE"
+            android:value="@string/launchUrl" />
```
FluorescentHallucinogen, line 90: This is a temporary stub. A (new?) string representing the actual absolute scope is needed.

packages/core/template_project/app/src/main/AndroidManifest.xml
```diff
@@ -61,6 +76,20 @@
                 android:value="<%= metadata.value %>" />
         <% } %>
 
+        <% if (isMetaQuest) { %>
```
FluorescentHallucinogen, line 79: @andreban

This is the third `if (isMetaQuest)` in this file. Should I refactor this to a feature like https://github.com/GoogleChromeLabs/bubblewrap/blob/main/packages/core/src/lib/features/ArCoreFeature.ts?

Why `arcore` option is a "feature", but `chromeosonly` option is just an "if condition" (https://github.com/GoogleChromeLabs/bubblewrap/blob/main/packages/core/template_project/app/src/main/AndroidManifest.xml#L29), while both add only one string to `AndroidManifest.xml`?

When should an "if condition" be used and when should a "feature" be used?

andreban: I think it's down to timelines - Chrome OS only was added before the Feature system was a thing. We should probably migrate.

packages/cli/src/lib/cmds/init.ts
andreban: Chrome OS only adds `<uses-feature android:name="org.chromium.arc" android:required="true" />`. Since it contains `required=true`, in theory, app stores should only install the package on Chrome OS devices. No clue if other stores are handling this though

FluorescentHallucinogen: I've double-checked, APK built with `--chromeosonly` and `--metaquest` options together installs and works on Meta Quest like a charm.

Also I've tried `ovr-platform-util upload-quest-build` command on it. The validator doesn't throw any errors about `<uses-feature android:name="org.chromium.arc" android:required="true" />`. It seems App Lab and Oculus Quests just ignore it.

I think increasing `minSdkVersion` form 19 to 23 doesn't break Chrome OS compatibility. Right?

FluorescentHallucinogen:
> Is there a way to change android:minSdkVersion="19" to 23, if --metaquest flag is passed? If so, how?

@andreban It seems I've found a solution. PTAL.

packages/core/src/lib/TwaManifest.ts
```diff
@@ -326,6 +337,8 @@ export class TwaManifest {
       features: {},
       shareTarget: TwaManifest.verifyShareTarget(webManifestUrl, webManifest.share_target),
       orientation: asOrientation(webManifest.orientation) || DEFAULT_ORIENTATION,
+      scope: webManifest['scope'],
+      fullScopeUrl: new URL(webManifest['scope'] || '.', webManifestUrl).toString(),
```
FluorescentHallucinogen, line 341: @andreban If `webManifest['scope']` value is invalid (e.g. `//`) the following error will be thrown: `Uncaught TypeError: Failed to construct 'URL': Invalid URL`. Should we add `scope` value validation? If so, how to check if the `scope` value is valid or not?

packages/core/template_project/app/build.gradle
```diff
@@ -70,11 +70,14 @@ android {
 
         <% if (webManifestUrl) { %>
```
FluorescentHallucinogen, line 71: @andreban In what cases is this condition false?

The `webManifestUrl` is now required for Meta Quest too, not only Chrome OS.

packages/core/src/lib/TwaManifest.ts
```diff
@@ -528,6 +543,10 @@ export interface TwaManifestJson {
   };
   enableSiteSettingsShortcut?: boolean;
   isChromeOSOnly?: boolean;
+  isMetaQuest?: boolean; // Older Manifests may not have this field.
+  scope?: string; // Older Manifests may not have this field.
+  fullScopeUrl?: string; // Older Manifests may not have this field.
+  minSdkVersion?: number; // Older Manifests may not have this field.
```
FluorescentHallucinogen, line 549: @andreban Please check these 4 lines above. Do they really have to be optional?

packages/core/src/lib/TwaManifest.ts
```diff
@@ -473,6 +486,8 @@ export class TwaManifest {
           webManifest['name']?.substring(0, SHORT_NAME_MAX_SIZE)),
       display: this.getNewFieldValue('display', fieldsToIgnore, oldTwaManifest.display,
           asDisplayMode(webManifest['display']!)!),
+      scope: this.getNewFieldValue('scope', fieldsToIgnore, oldTwaManifest.scope,
```
FluorescentHallucinogen, line 489: @andreban Please check this carefully, because I don't understand fully what's happening here.

packages/core/src/lib/TwaManifest.ts
```diff
@@ -201,6 +207,10 @@ export class TwaManifest {
     this.enableSiteSettingsShortcut = data.enableSiteSettingsShortcut != undefined ?
       data.enableSiteSettingsShortcut : true;
     this.isChromeOSOnly = data.isChromeOSOnly != undefined ? data.isChromeOSOnly : false;
+    this.isMetaQuest = data.isMetaQuest != undefined ? data.isMetaQuest : false;
+    this.scope = data.scope;
+    this.fullScopeUrl = data.fullScopeUrl ? new URL(data.fullScopeUrl) : undefined;
+    this.minSdkVersion = data.minSdkVersion || DEFAULT_MIN_SDK_VERSION;
```
FluorescentHallucinogen, line 213: @andreban Please check these 4 lines above carefully.

packages/core/src/lib/TwaManifest.ts
```diff
@@ -155,6 +157,10 @@ export class TwaManifest {
   alphaDependencies: alphaDependencies;
   enableSiteSettingsShortcut: boolean;
   isChromeOSOnly: boolean;
+  isMetaQuest: boolean;
+  scope: string | undefined;
+  fullScopeUrl?: URL;
+  minSdkVersion: number;
```
FluorescentHallucinogen, line 163: @andreban Please check these 4 lines above carefully. What of them should be optional and what can be undefined?

packages/core/src/lib/TwaManifest.ts
```diff
@@ -201,6 +207,10 @@ export class TwaManifest {
     this.enableSiteSettingsShortcut = data.enableSiteSettingsShortcut != undefined ?
       data.enableSiteSettingsShortcut : true;
     this.isChromeOSOnly = data.isChromeOSOnly != undefined ? data.isChromeOSOnly : false;
+    this.isMetaQuest = data.isMetaQuest != undefined ? data.isMetaQuest : false;
+    this.scope = data.scope;
```
FluorescentHallucinogen, line 211: @andreban Should I add the `scope` at all? It is not actually used in TWA and is only needed to generate `fullScopeUrl` that is used only by Meta Quest.

packages/cli/src/lib/cmds/init.ts
andreban: In this case, `uses-feature` is used the application stores to figure out if the application is compatible with a given device. If it isn't it would be installable from the app store. I don't think increasing minSdkVersion will break compatibility.

packages/core/src/lib/TwaManifest.ts
andreban: It seems what we'd want is to catch the error and throw a friendlier message?

packages/core/src/lib/TwaManifest.ts
andreban: An option would be using the Manifest URL to generate a default scope and fullScopeUrl, if we want to prevent undefined values. But I'm happy either way.

packages/core/src/lib/TwaManifest.ts
andreban: `merge` would need to reflect adding / removing / changing a scope?

packages/core/src/lib/TwaManifest.ts
andreban: looks good.

packages/core/src/lib/TwaManifest.ts
andreban: Looks right to me - this if `scope` is in `fieldsToIgnore`, it will ignore the new value, basically.

packages/core/src/lib/TwaManifest.ts
andreban: They do, since those are new fields and won't exist in old manifests. This forces us to do proper checks for `undefined` later in the code.

packages/core/template_project/app/build.gradle
FluorescentHallucinogen: @andreban What about this?

BTW, in theory, can it be the case that the `chromeosonly` option is passed, but `webManifestUrl` is missing? 🤔

I mean, why is this wrapped in the "if condition" if it is always set. Or is it not always?

packages/core/src/lib/TwaManifest.ts
FluorescentHallucinogen: @andreban Is there a case when there is no Manifest URL?

packages/core/template_project/app/build.gradle
andreban: In the past, that field didn't exist in twa-manifest.json. So, older manifests may not have that information.

packages/core/src/lib/TwaManifest.ts
andreban: Yes, we didn't save that information in the past.

FluorescentHallucinogen: @andreban I mean are there any ways to generate `twa-manifest.json` file without CLI wizard? If we use CLI wizard we must provide `--manifest` parameter. So we cannot get into an invalid state.

As an example, what if I open the `twa-manifest.json` file in a text editor and edit it so that `"isMetaQuest": true`, but `"webManifestUrl"` is missing? Should we care about such cases?

Can we get into an invalid state by updating the old `twa-manifest.json` (using `merge` command on the old `twa-manifest.json`)?

FluorescentHallucinogen:
@andreban The pull request is ready to review and merge. :wink:

andreban:
@PEConn would you mind taking a look too?

packages/core/template_project/app/build.gradle
```diff
@@ -70,11 +70,16 @@ android {
...
             // The URL the Web Manifest for the Progressive Web App that the TWA points to. This
-            // is used by Chrome OS to open the Web version of the PWA instead of the TWA, as it
-            // will probably give a better user experience for non-mobile devices.
+            // is used by Chrome OS and Meta Quest to open the Web version of the PWA instead of
+            // the TWA, as it will probably give a better user experience for non-mobile devices.
             resValue "string", "webManifestUrl", '<%= webManifestUrl %>'
         <% } %>
 
+        <% if (fullScopeUrl) { %>
```
PEConn, line 78: if fullScopeUrl && isMetaQuest?

packages/cli/README.md
```diff
@@ -369,8 +371,9 @@ Fields:
 |splashScreenFadeOutDuration|number|true|Duration for the splash screen fade out animation.|
 |startUrl|string|true|The start path for the TWA. Must be relative to the domain.|
 |themeColor|string|true|The color used for the status bar.|
-|webManifestUrl|string|false|Full URL to the PWA Web Manifest. Required for the application to be compatible with Chrome OS devices.|
- 
+|webManifestUrl|string|false|Full URL to the PWA Web Manifest. Required for the application to be compatible with Chrome OS and Meta Quest devices.|
+|fullScopeUrl|string|false|The navigation scope that the browser considers to be within the app. If the user navigates outside the scope, it reverts to a normal web page inside a browser tab or window. Must be a full URL. Required only for Meta Quest devices.|
```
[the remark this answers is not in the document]
FluorescentHallucinogen: `fullScopeUrl` is not a flag, it's a regular field (like `startUrl` or `webManifestUrl`). This field is only read by the Meta Quest Browser on Meta Quest devices and just ignored by other devices as unknown.

packages/core/template_project/app/build.gradle
FluorescentHallucinogen: Why then don't we use `<% if (webManifestUrl && isChromeOSOnly { %>` instead of `<% if (webManifestUrl) { %>` in line 71?

packages/cli/README.md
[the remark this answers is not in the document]
PEConn: I'm wary that if the user uses `fullScopeUrl` but not `isMetaQuest` they're doing something wrong. It seems more likely they they meant to add Quest support, but missed the `isMetaQuest` flag, than they intentionally added `fullScopeUrl` for no reason.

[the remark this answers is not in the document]
FluorescentHallucinogen: @PEConn What exactly do you suggest?

We cannot use the presence of `fullScopeUrl` to throw an error.

The problem is that currently `fullScopeUrl` will be added to `twa-manifest.json` in any case, regardless of the `isMetaQuest` flag value. But if `isMetaQuest` is `false` or missing, the `fullScopeUrl` will not be used in any way.

That's why I've added `Required only for Meta Quest devices.` to the field description.

BTW, I didn't add the `minSdkVersion` here. My rationale: because it shouldn't be changed by hand. Should I add it?

This is the problem of related (computed) fields.

If `isChromeOSOnly` is `true`, then `webManifestUrl` must be present. But I can edit `twa-manifest.json` in text editor and remove `webManifestUrl` or just use the old `twa-manifest.json` (generated by the old version of Bubblewrap) that has no `webManifestUrl`. In this case, we don't throw any build errors.

E.g. what stops the user from setting `isMetaQuest` to `true` and e.g. changing `minSdkVersion` back from `23` to `19` or something else?

As an idea, I can suggest to improve the description. What about adding something like `To use this field, the 'isMetaQuest' flag must be 'true'` to the `fullScopeUrl` description? Will it solve the problem?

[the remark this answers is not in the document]
PEConn: Ah OK, I misunderstood where `fullScopeUrl` was coming from - yeah, if it's going to be in `twa-manifest.json` regardless there's not anything we can do. Things are fine as is.

packages/core/template_project/app/build.gradle
FluorescentHallucinogen: And also why then don't we use `<% if (webManifestUrl && isChromeOSOnly) { %>` instead of `<% if (webManifestUrl) { %>` in `packages/core/template_project/app/src/main/AndroidManifest.xml` file?

packages/cli/src/lib/cmds/init.ts
```diff
@@ -245,6 +246,11 @@ export async function init(
...
+    twaManifest.isMetaQuest = true;
+    twaManifest.minSdkVersion = 23;
```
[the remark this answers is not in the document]
FluorescentHallucinogen: Done.

packages/core/template_project/app/build.gradle
andreban: `webManifestUrl` is additional information that can be used by anyone and, nowadays, it's always saved on `init` - it doesn't have to be exclusive to Chrome OS. I think `fullScopeUrl` would fall in the same category - it won't cause any bad side effects to have it. The counter argument would be bloating the APK, but I'd guess it's small enough that it won't have significant impact.

andreban (approved):
LGTM

PEConn (approved):
This looks good, thanks for your work on it.

FluorescentHallucinogen:
@PEConn @andreban Could you please release a new version of Bubblewrap that contains this pull request? It's needed for further work on PWABuilder.

PEConn:
Done, the changes should be in version 1.18.0 of `core` and `cli`.

FluorescentHallucinogen:
@PEConn Thanks!

BTW, just noticed that version `1.18.0` is not listed at https://github.com/GoogleChromeLabs/bubblewrap/releases.

PR #714, opened 2022-07-13
feat: add docker image for cli

anbraten:
Add a docker image which ships the bubblewrap cli (with all needed dependencies like java & android-sdk). This docker image can be used to easily run the cli as container for local use or for example in CI systems.

To build the container image this PR introduces a new Github workflow pushing the container image to <https://ghcr.io>. The workflow will create a `:next` tagged image on pushes to the `main` branch and to git tags / releases it will push an image with the tag name without the `v` prefix.


andreban:
@anbraten there's not description to this PR or related issue. Could you provide an overview of what this is trying to achieve and why to the description?

anbraten:
I have added a short note in the docs of the cli package.

Issue #699, opened 2022-05-30
Unable to load the "svg2img" library. SVGs are not supported


nunoperalta:
I'm getting this warning every time I run a `bubblewrap build` (or any other `bubblewrap` command really).
What can I do to resolve this?

```
> bubblewrap build
Unable to load the "svg2img" library. SVGs are not supported
,-----.        ,--.  ,--.  ,--.
|  |) /_,--.,--|  |-.|  |-.|  |,---.,--.   ,--,--.--.,--,--.,---.
|  .-.  |  ||  | .-. | .-. |  | .-. |  |.'.|  |  .--' ,-.  | .-. |
|  '--' '  ''  | `-' | `-' |  \   --|   .'.   |  |  \ '-'  | '-' '
`------' `----' `---' `---'`--'`----'--'   '--`--'   `--`--|  |-'
                                                           `--'
Using passwords set in the BUBBLEWRAP_KEYSTORE_PASSWORD and BUBBLEWRAP_KEY_PASSWORD environmental variables.

Building the Android App...
```

I've already run:

`npm i -g svg2img`
`npm i svg2img` (in the folder where I run `bubblewrap`)

But I'm still getting the same issue.

In the `catch` block at: https://github.com/GoogleChromeLabs/bubblewrap/blob/830dbe695c1e690c81c0a6b29dbdd75693586f47/packages/core/src/lib/wrappers/svg2img.ts
It would probably help if the actual exception was printed too (or logged somewhere).

Thank you!

 - OS: Windows 11
 - Version: 1.17.2

andreban:
This is due to the [`canvg`](https://www.npmjs.com/package/canvg) dependency on native libraries. If those are not available, `canvg` fails to install when `npm install @bubblewrap/cli` runs.  Since many users don't need SVGs, Bubblewrap uses it a an optional dependency and tries to use it at runtime.

I just learned about [`@resvg/resvg-wasm`](https://www.npmjs.com/package/@resvg/resvg-wasm), which is based on pure Rust implementation called [`resvg`](https://crates.io/crates/resvg), which can be compiled to WASM and would remove any native dependencies. It's probably worth trying to replace `canvg` with `@resvg/resvg-wasm`.

@tropicadri FYI

andreban:
A new version using resvg was published.

PR #742, opened 2023-01-13
Replace svg2img with resvg-js

andreban:
** Needs broader testing**

svg2img depends on node-canvas which, as a platform-specific library frequently breaks when deployed to specific architectures.

resvg-js uses a WASM implementation, which makes it less likely to break.

Should improve on #699


PEConn (approved):
LGTM!

andreban:
Tested on a M1 MacOS. Working fine :)

andreban:
Tested on a ChromeOS Linux. Working fine.

PR #763, opened 2023-02-14
Fix #759 Drop support for Node 12

### LINKED ISSUE #759: [Discussion] Drop support for Node 12
One of the Bubblewrap, dependencies, lerna, dropped support for Node 12 and the project fails to build. We should drop support in Bubblewrap as well.

See:
https://github.com/lerna/lerna/blob/main/package.json#L15

FluorescentHallucinogen:


FluorescentHallucinogen:
@andreban We are increasing the minimum supported Node.js version only because of the `lerna` dev dependency in the root project. But `cli`, `core` and `validator` packages don't use `lerna` (directly). I mean that output npm packages still can work in Node.js 12.

Should we update `/packages/cli/README.md` and `/packages/core/README.md` at all?

Should we add `"engines": { "node": ">=14.15.0" }` to `/packages/cli/package.json` and `/packages/core/package.json`?

Should we update https://github.com/GoogleChromeLabs/bubblewrap/blob/main/packages/cli/src/lib/Cli.ts#L39-L42?

andreban:
By dropping support to Node 12 we'll stop testing the code against it and ensuring using APIs that work on Node 12.  Right now, the whole application will still be compatible but, over time, it's likely that it will stop working on Node 12. I'd rather clarify this to users in documentation / code now, rather than surprise people when it breaks.

andreban:
We'd probably want to update the CI too: https://github.com/GoogleChromeLabs/bubblewrap/blob/main/.github/workflows/nodejs.yml#L12

FluorescentHallucinogen:
@andreban BTW, what about the `validator` package? I see that `core` and `cli` packages was updated to `v1.19.1`, but the `validator` is still `v1.18.1`. 🤔

FluorescentHallucinogen:
@andreban And unfortunately you have not answered the question, should we add `"engines": { "node": ">=14.15.0" }` to `/packages/cli/package.json` and `/packages/core/package.json`? And what about `/packages/validator/package.json`?

If so, should we also create an `.npmrc` file with `engine-strict=true`? Without this flag, the `engines` field in `package.json` is advisory only and will only produce warnings when the package is installed as a dependency.

andreban (approved):
LGTM, Thank you!

PR #770, opened 2023-03-11
Update DigitalAssetLinks.ts

hslee2008:
# Problem (?) I had before

After using bubblewrap for the first time, I had 1 issue that took me days to fix.
It was the Digital Asset Links JSON file.

The problem was the fingerprints were spread out on one item in arrays.

```json
[
  {
    "relation": [
      "delegate_permission/common.handle_all_urls"
    ],
    "target": {
      "namespace": "android_app",
      "package_name": "app.web.test.twa",
      "sha256_cert_fingerprints": ["fingerprint1"]
    }
  },
  {
    "relation": [
      "delegate_permission/common.handle_all_urls"
    ],
    "target": {
      "namespace": "android_app",
      "package_name": "app.web.test.twa",
      "sha256_cert_fingerprints": ["fingerprint2"]
    }
  }
]
```

This was how the file was generated.

# Fix
After looking at few sources, I came to conclusion that the `assetlinks.json` file should be something like below.

1. [Example Format in AssetLinks docs - Google](https://github.com/google/digitalassetlinks/blob/master/well-known/details.md#assetdescriptor)
2. [Example File in AssetLinks docs - Google](https://github.com/google/digitalassetlinks/blob/master/well-known/specification.md#details)
3. Also, looking at few examples such as [Google Docs AssetLinks.json](https://0.docs.google.com/.well-known/assetlinks.json), we can see that the fingerprints are contained in arrays by package names.
4. Checking out why my site was not validated in Google Play Console

> Note: I could not recreate the error message Google Play Console told me to update my `asssetlinks.json` file to contain all the fingerprints in one array. I will edit this pull request again if I manage to do so.

```json
[
  {
    "relation": [
      "delegate_permission/common.handle_all_urls"
    ],
    "target": {
      "namespace": "android_app",
      "package_name": "app.web.test.twa",
      "sha256_cert_fingerprints": [
        "fingerprint1",
        "fingerprint2"
      ]
    }
  }
]
```

# Todos
- [x] Update corresponding testing files
- [ ] Recreate Google Play Console Error Message

### DIFF
--- a/packages/core/src/lib/DigitalAssetLinks.ts
+++ b/packages/core/src/lib/DigitalAssetLinks.ts
@@ -16,19 +16,17 @@
 
 export class DigitalAssetLinks {
   static generateAssetLinks(applicationId: string, ...sha256Fingerprints: string[]): string {
-    const assetlinks = new Array<string>();
-    assetlinks.push('[');
-    sha256Fingerprints.forEach((sha256Fingerprint, index) => {
-      if (index > 0) {
-        assetlinks.push(',');
+    if (sha256Fingerprints.length === 0) {
+      return '[]';
+    }
+
+    return `[{
+      "relation": ["delegate_permission/common.handle_all_urls"],
+      "target": {
+        "namespace": "android_app",
+        "package_name": "${applicationId}",
+        "sha256_cert_fingerprints": ["${sha256Fingerprints.join('\", \"')}"]
       }
-      assetlinks.push(`{
-        "relation": ["delegate_permission/common.handle_all_urls"],
-        "target" : { "namespace": "android_app", "package_name": "${applicationId}",
-                     "sha256_cert_fingerprints": ["${sha256Fingerprint}"] }
-      }\n`);
-    });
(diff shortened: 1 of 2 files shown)


andreban:
Looping in @PEConn, as he may remember the details better. IIRC, listing all the fingerprints in the array is the approach thought to be right initially, but then found that the current implementation, which is documented [here](https://developer.chrome.com/docs/android/trusted-web-activity/android-for-web-devs/#adding-more-keys), is the correct one.

PEConn:
It turns out we were originally correct and our "correction" was unnecessary. According to [this](https://developer.android.com/training/app-links/verify-android-applinks#web-assoc) - "This field supports multiple fingerprints, which can be used to support different versions of your app, such as debug and production builds.".

So, putting all the fingerprints in a single array should work. (However, my understanding is that having separate entries should also work.)

andreban:
Ah, thanks for clarifying @PEConn. PR looks good to me then.

andreban (approved):
LGTM, thank you!

PR #773, opened 2023-03-18
Update `short_name` and `shortName` key, type, test

hslee2008:
# Opinion
Assuming that [Example in Test File](https://github.com/GoogleChromeLabs/bubblewrap/blob/81adecf07002fbd55451558257bb86242d96066a/packages/core/src/spec/lib/TwaManifestSpec.ts#L219) is the most basic object to create shortcuts, there are a lot of codes/test/types (maybe) not updated.

But if `shortName` is the correct way to make short names rather than `short_name`, if the manifest is first fetched through the cli, it will create the following error as `shortName` doesn't exist.

```
C:\Users\USER\AppData\Roaming\npm\node_modules\@bubblewrap\cli\node_modules\@bubblewrap\core\dist\lib\util.js:299
    return input.replace(/[\\']/g, '\\\\\\$&');
                 ^

TypeError: Cannot read properties of undefined (reading 'replace')
    at Object.escapeGradleString (C:\Users\USER\AppData\Roaming\npm\node_modules\@bubblewrap\cli\node_modules\@bubblewrap\core\dist\lib\util.js:299:18)
    at ShortcutInfo.toString (C:\Users\USER\AppData\Roaming\npm\node_modules\@bubblewrap\cli\node_modules\@bubblewrap\core\dist\lib\ShortcutInfo.js:49:35)
    at C:\Users\USER\AppData\Roaming\npm\node_modules\@bubblewrap\cli\node_modules\@bubblewrap\core\dist\lib\TwaManifest.js:205:67
    at Array.map (<anonymous>)
    at Object.generateShortcuts (C:\Users\USER\AppData\Roaming\npm\node_modules\@bubblewrap\cli\node_modules\@bubblewrap\core\dist\lib\TwaManifest.js:205:37)
    at eval (lodash.templateSources[2]:34:11)
    at TwaGenerator.applyTemplate (C:\Users\USER\AppData\Roaming\npm\node_modules\@bubblewrap\cli\node_modules\@bubblewrap\core\dist\lib\TwaGenerator.js:179:55)

Node.js v18.13.0
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

So, I think the best approach is to make both handlers for `shortName` and `short_name` rather than using `shortName` only.

But then, some few tests and types files have either `shortName` or `short_name` which is really confusing while trying to understand how shortcuts work.

In conclusion, I think the best approach is to make `shortName` the default key for all, but make it possible to using `short_name`.

# Files Changed
`packages/core/src/lib/ShortcutInfo.ts`: Update the `.toString()` method, use both `short_name` and `shortName`.

`packages/core/src/lib/TwaManifest.ts`: Add warning message for skipping shortcuts items after 4.

`packages/core/src/lib/types/WebManifest.ts`: Add `short_name` type.

`packages/core/src/spec/lib/TwaManifestSpec.ts` & `packages/core/src/spec/lib/ShortcutInfoSpec.ts`: Update shortcut tests (+ Format icon key).

### DIFF
--- a/packages/core/src/lib/ShortcutInfo.ts
+++ b/packages/core/src/lib/ShortcutInfo.ts
@@ -48,9 +48,9 @@ export class ShortcutInfo {
 
   toString(index: number): string {
     return `[name:'${escapeGradleString(this.name)}', ` +
-      `short_name:'${escapeGradleString(this.shortName)}', ` +
+      `shortName:'${escapeGradleString(this.shortName)}', ` +
       `url:'${this.url}', ` +
-      `icon:'${this.assetName(index)}']`;
+      `chosenIconUrl:'${this.assetName(index)}']`;
   }
 
   assetName(index: number): string {
@@ -66,7 +66,8 @@ export class ShortcutInfo {
    * @returns {TwaManifest}
    */
   static fromShortcutJson(webManifestUrl: URL, shortcut: WebManifestShortcutJson): ShortcutInfo {
-    const name = shortcut.name || shortcut.short_name;
+    const manifestShortName = shortcut.shortName || shortcut.short_name;
+    const name = shortcut.name || manifestShortName;
 
     if (!shortcut.icons || !shortcut.url || !name) {
       throw new Error('missing metadata');
@@ -88,7 +89,7 @@ export class ShortcutInfo {
       return icon ? new URL(icon.src, webManifestUrl).toString() : undefined;
     }
 
-    const shortName = shortcut.short_name || shortcut.name!.substring(0, SHORT_NAME_MAX_SIZE);
+    const shortName = manifestShortName || shortcut.name!.substring(0, SHORT_NAME_MAX_SIZE);
     const url = new URL(shortcut.url, webManifestUrl).toString();
     const shortcutInfo = new ShortcutInfo(name!, shortName!, url, resolveIconUrl(suitableIcon),
         resolveIconUrl(suitableMaskableIcon), resolveIconUrl(suitableMonochromeIcon));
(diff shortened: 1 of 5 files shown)


packages/core/src/lib/ShortcutInfo.ts
```diff
@@ -48,9 +48,9 @@ export class ShortcutInfo {
 
   toString(index: number): string {
     return `[name:'${escapeGradleString(this.name)}', ` +
-      `short_name:'${escapeGradleString(this.shortName)}', ` +
+      `shortName:'${escapeGradleString(this.shortName)}', ` +
       `url:'${this.url}', ` +
-      `icon:'${this.assetName(index)}']`;
+      `chosenIconUrl:'${this.assetName(index)}']`;
```
andreban, line 53: Is there a reason to change `icon` -> `chosenIconUrl`? This will break the code [here](https://github.com/GoogleChromeLabs/bubblewrap/blob/81adecf07002fbd55451558257bb86242d96066a/packages/core/template_project/app/build.gradle#L179). I think this is fine to be called just `icon` at his point, as `chosenIconUrl` is more of an implementation detail then.

andreban (commented):
Hey @hslee2008, thank you for the PR. In short, `short_name` is the name defined on the [W3C Web Manifest](https://developer.mozilla.org/en-US/docs/Web/Manifest), which is a different thing than the TwaManifest, where a similar field exists and is called `shortName`. We use the Web Manifest to bootstrap a configuration and generate a TWA Manifest, so we should never encounter one with a `shortName` field - and if we do, that's technically an invalid Web Manifest.

I'd strongly recommend creating an issue instead of PRs for discussions/opinions, instead of PRs too. It's less work for us to review and for you to create the PR.

PR #781, opened 2023-04-17
Comment out flaky jdkDoctor test

andreban:
The mock filesystem used for tests seems to be broken, causing the test to fail locally. Commenting out the test until that's fixed.


andreban:
> Could we use [xit](https://jasmine.github.io/api/4.6/global.html#xit) instead of actually commenting the code out?

#TIL. Done :)

PEConn (approved):
Thanks!

Issue #793, opened 2023-05-09
Bubblewrap should tell me that it needs a specific version of the JDK BEFORE I give it the path


christianhujer:
```
$ bubblewrap init --manifest https://chatncharge.nelkinda.com/manifest.json
,-----.        ,--.  ,--.  ,--.
|  |) /_,--.,--|  |-.|  |-.|  |,---.,--.   ,--,--.--.,--,--.,---.
|  .-.  |  ||  | .-. | .-. |  | .-. |  |.'.|  |  .--' ,-.  | .-. |
|  '--' '  ''  | `-' | `-' |  \   --|   .'.   |  |  \ '-'  | '-' '
`------' `----' `---' `---'`--'`----'--'   '--`--'   `--`--|  |-'
                                                           `--'    
? Do you want Bubblewrap to install JDK?
  (Enter "No" to use your JDK installation) No
? Path to your existing JDK: /home/christian.hujer/.sdkman/candidates/java/current/
>> JDK version not supported. JDK version 11 is required.
```
(current was JDK 19)
It would've been great if Bubblewrap would've told me that it requires JDK 11 _before_ it wasted my time figuring out which PATH my currently used JDK is on.

**Is your feature request related to a problem? Please describe.**
When I started bubblewrap, it asked me for the JDK. Which is bad enough, as I have a JDK on `PATH` as well as several other JDKs managed by _SDKMAN!_. But what was worse is that once I figured what path my JDK is and entered it, bubblewrap told me it's not happy with my JDK, it needs JDK 11. If bubblewrap would've told me right away, I could've saved the hassle of giving it a JDK 19 path and would have given it a JDK 11 path right away.

**Describe the solution you'd like**
I'd like bubblewrap to tell me `Path to your existing JDK (JDK11, not older, not newer)` instead of `Path to your existing JDK` if bubblewrap is really so picky about the JDK.

andreban:
A quick comment on this - this particular requirement comes from the Android SDK (Bubblewrap doesn't use Java at all). The Android SDK is also tricky to get right, as specific versions can have different bugs. For this reason, we strongly recommend allowing Bubblewrap to install the JDK and Android SDK, as setting those manually is error prone. The manual setup and required JDK version is documented at https://github.com/GoogleChromeLabs/bubblewrap/tree/main/packages/cli#get-the-java-development-kit-jdk-11.

christianhujer:
[...]

PR #795, opened 2023-05-10
Update JDK/Android SDK prompt strings

andreban:
- Sets Bubblewrap handling installations as recommended.
- Clarifies JDK 11 when asking for paths.

See #793


PEConn (approved):
Thanks!

PR #815, opened 2023-07-19
Update ABH billing version to 1.0.0-alpha10

SayedElabady:
Update ABH billing dependency to point out to the new version 1.0.0-alpha10


andreban (approved):
LGTM, thank you!

PR #817, opened 2023-07-31
Bump core version to 1.20.1

SayedElabady:
Hey Andre, please advice if I need to change more files to bump the core version to v1.20.1.

### DIFF
--- a/packages/cli/package-lock.json
+++ b/packages/cli/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.20.0",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


andreban:
Hey @SayedElabady, thanks for the PR. The [latest version on NPM](https://www.npmjs.com/package/@bubblewrap/core?activeTab=versions) is `1.20.0` , so this will break the CLI. Also, we use [lerna](https://lerna.js.org/) to manage versions and deployments, so this should be handled automatically when publishing to NPM via `npm run publish`. Happy to grab time to discuss better.

andreban:
Closing as this was released via `lerna`

Issue #786, opened 2023-04-26
ERROR "playPublish" is not a valid command!


lovkyndig:
Screenclip of error:
<img width="457" alt="playPublish not valid command" src="https://user-images.githubusercontent.com/125480548/234495730-6cd12e52-9537-4a5e-9a6b-4c7cabcbdde5.PNG">

[`bubblewrap/cli/README.md`](https://github.com/GoogleChromeLabs/bubblewrap/tree/main/packages/cli#play) give this information about _play_ and _playPublish_:
<img width="734" alt="playPublish-command" src="https://user-images.githubusercontent.com/125480548/234496000-900aa8bb-f85b-4f24-830c-f6c957ac9fef.PNG">

## Why this error?
I am using windows and the latest version of everything.
Do I need to install `core` for this command?

## Source:
I was also reading about this command in [`#621`](https://github.com/GoogleChromeLabs/bubblewrap/commit/f5d32bb32032f97161eb6e8b6a1e3727af5d357b)

FRSgit:
Hey! I've just created a PR resolving [this one](https://github.com/GoogleChromeLabs/bubblewrap/pull/885) - hopefully it will get merged soon.

And until it does - in my repo I'm using [pnpm patch](https://pnpm.io/cli/patch) to [uncomment needed lines by myself](https://github.com/FRSOURCE/fast-angle/blob/main/patches/%40bubblewrap__cli.patch). You can do the same on your side natively in pnpm/yarn or  if you need npm then you can try [using this package](https://www.npmjs.com/package/patch-package).

PR #885, opened 2024-09-01
feat: publish play-related CLI command

FRSgit:
resolves https://github.com/GoogleChromeLabs/bubblewrap/issues/786

### Description

Even though available in the repository README, the play-related CLI commands (`play`, `playPublish`, `playVersionCheck`, `playRetain`) are not publicly available in the CLI itself.

It turns out, that the code is there already - just needs to be uncommented. They all seem to be working.

I personally use the `playPublish` in my repository to fully automate publishing of my app to the Google Play (see the source [here](https://github.com/FRSOURCE/fast-angle/blob/main/packages/android/package.json#L10) and [here](https://github.com/GoogleChromeLabs/bubblewrap/issues/786)).

### DIFF
--- a/packages/cli/src/lib/Cli.ts
+++ b/packages/cli/src/lib/Cli.ts
@@ -29,9 +29,8 @@ import {updateConfig} from './cmds/updateConfig';
 import {doctor} from './cmds/doctor';
 import {merge} from './cmds/merge';
 import {fingerprint} from './cmds/fingerprint';
-// import {play, PlayArgs} from './cmds/play';
 import {fetchUtils} from '@bubblewrap/core';
-// import { play, PlayArgs, playPublish } from './cmds/play';
+import { play, PlayArgs, playPublish } from './cmds/play';
 
 export class Cli {
   async run(args: string[]): Promise<boolean> {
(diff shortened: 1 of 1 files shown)


packages/cli/src/lib/Cli.ts
```diff
@@ -89,14 +88,14 @@ export class Cli {
...
-      // case 'play':
-      //   return await play(parsedArgs as unknown as PlayArgs);
-      // case 'playPublish':
-      //   return await play(parsedArgs as unknown as PlayArgs, 'publish');
-      // case 'playVersionCheck':
-      //     return await play(parsedArgs as unknown as PlayArgs, 'versionCheck');
-      // case 'playRetain':
-      //     return await play(parsedArgs as unknown as PlayArgs, 'retain');
+      case 'play':
+        return await play(parsedArgs as unknown as PlayArgs);
```
OlegNitz, line 92: Hi Jakub, did you test all 4 of these new commands? I doubt that the `play` command would work, because the `play()` function requires the second parameter `command`. I am not sure what the original author meant by this, but I suggest deleting the two lines above.

FRSgit: hey, you're right - I filed this PR rather lazily.
Check out new commit: I've move play commands into single namespace (so, all of the `publish`, `retain`, `versionCheck` commands are subcommands of `bubblewrap play`) and cleanup the implementation a bit - now it resembles other parts of the codebase more (mainly fingerprint/build commands).

I've also tested once again `publish` and `versionCheck` - they are working. Not sure about the retain - that's why I'm leaving it marked as "experimental" in the docs and in the implementation (there is an additional prompt, so the needs to confirm when they want to use an experimental feature).

What do you think?

OlegNitz (approved):
Thank you for putting more time into this change.

PR #910, opened 2025-02-24
Protocol Handler support

gstepniewski-google:


packages/core/template_project/app/src/main/AndroidManifest.xml
```diff
@@ -207,6 +207,15 @@
                         android:host="<%= additionalOrigin %>"/>
                 </intent-filter>
             <% } %>
+
+            <% for (const protocolHandler of protocolHandlers) { %>
+                <intent-filter>
```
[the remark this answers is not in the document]
gstepniewski-google: Not sure if that would work - the way I understand it is for verification to work you need

1. An android:host tag (because otherwise how does it know which domain to verify?)
2. A http/https scheme

https://developer.android.com/training/app-links/verify-android-applinks#add-intent-filters

packages/core/template_project/app/src/main/java/LauncherActivity.java
```diff
@@ -57,6 +66,15 @@ protected Uri getLaunchingUrl() {
             <%= code %>
         <% } %>
 
+        /* Protocol Handler support */
+        String scheme = uri.getScheme();
```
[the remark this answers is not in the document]
gstepniewski-google: I thought about this, but I'm not sure how to pass the map of handlers and the launchUrl to ABH in a way that will work with apps that use ABH but are not generated with bubblewrap... Any ideas?

packages/core/src/spec/lib/TwaManifestSpec.ts
```diff
@@ -354,6 +354,10 @@ describe('TwaManifest', () => {
           'purpose': 'any',
         },
         ],
+        'protocol_handlers': [{
+          'protocol': 'web+test-replace',
+          'url': 'test-format-web',
```
wzwonarz, line 359: this is invalid url, it always has to have %s formatter:
https://developer.chrome.com/docs/web-platform/best-practices/url-protocol-handler#how_to_use_url_protocol_handler_registration_for_pwas

packages/core/template_project/app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
ukaratkevich: Oh yeah, thanks for clarification. We had a problem in browser handler that lint failed for link verification with scheme other than http/s.

packages/core/src/lib/TwaManifest.ts
```diff
@@ -479,6 +482,19 @@ export class TwaManifest {
...
+      protocolHandlersMap.set(handler.protocol, handler.url);
+    }
+    if (!(fieldsToIgnore.includes('protocol_handlers'))) {
+      for (const handler of webManifest.protocol_handlers ?? []) {
+        protocolHandlersMap.set(handler.protocol, handler.url);
+      };
+    }
+    const protocolHandlers = Array.from(protocolHandlersMap.entries()).map(([protocol, url]) => {
+      return {protocol, url} as ProtocolHandler;
+    });
```
wzwonarz, line 496: I think we need to add the list of allowed protocols in the protocol_handlers entry. Also the format of the url should be verified:
https://developer.chrome.com/docs/web-platform/best-practices/url-protocol-handler#how_to_use_url_protocol_handler_registration_for_pwas

gstepniewski-google: @orbital17 how does validation of webmanifest.json currently work? what's the best place to add this?

packages/core/template_project/app/src/main/java/LauncherActivity.java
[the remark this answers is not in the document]
ukaratkevich: Just checked package manager and seems like we can't query Activity's intent filters, so the only way to get them is probably `protected Map<String, String> getProtocolHandlers()` that is exposed from ABH and change `getLaunchingUrl()` implementation to handle protocols.

packages/core/src/spec/lib/TwaManifestSpec.ts
gstepniewski-google: didn't think that would be needed for unit tests, but added just in case

packages/core/template_project/app/src/main/java/LauncherActivity.java
```diff
@@ -57,6 +66,15 @@ protected Uri getLaunchingUrl() {
...
+        if (mProtocolHandlers.containsKey(scheme)) {
+            String target = uri.toString().replace(scheme + "://", "");
+            String format = mProtocolHandlers.get(scheme);
+            String baseUrl = getString(R.string.launchUrl);
```
gstepniewski-google, line 74: baseUrl should already be prepended to format during manifest processing

PR #911, opened 2025-03-06
DO NOT MERGE YET Protocol Handler support

gstepniewski-google:
Waiting for a new version of android-browser-helper to become available and add the new protected method in LauncherActivity.

### DIFF
--- a/packages/core/src/lib/types/ProtocolHandler.ts
+++ b/packages/core/src/lib/types/ProtocolHandler.ts
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2025 Google Inc. All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+export interface ProtocolHandler {
+  protocol: string;
+  url: string;
+}
+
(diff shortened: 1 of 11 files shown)


packages/core/template_project/app/src/main/java/LauncherActivity.java
```diff
@@ -48,15 +51,25 @@ protected void onCreate(Bundle savedInstanceState) {
         }
     }
 
+    <% if (protocolHandlers.length > 0) { %>
```
orbital17, line 54: @gstepniewski-google I think this whole method should go into launcherActivity.methods variable. And probably it's worth adding a new feature for protocol handlers in src/lib/features

gstepniewski-google: Done!

packages/core/src/lib/types/ProtocolHandler.ts
```diff
@@ -0,0 +1,119 @@
...
+const ProtocolHandlerExtraScheme = /^web\+[a-z]+$/;
+const ProtcolHandlerAllowedSchemes = [
+  'bitcoin', 'ftp', 'ftps', 'geo', 'im', 'irc', 'ircs',
+  'magnet', 'mailto', 'matrix', 'news', 'nntp', 'openpgp4fpr',
+  'sftp', 'sip', 'ssh', 'urn', 'webcal', 'wtai', 'xmpp',
+];
+// 'mms', 'sms', 'smsto', and 'tel' are not supported!
+
+const ProtocolHandlerAllowedAbsoluteUrlSchemes = [
+  'http:', 'https:',
```
glennhartmann, line 31: We may want to insist on https and disallow http

gstepniewski-google: Should we do it for all websites or only the ones hosted on https? In other words, should we force the target url to have the same scheme as the base url?

gstepniewski-google: [This](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler) actually says "The handler URL must use the https scheme". I'm going to remove http.

PR #928, opened 2025-04-29
Add file handling support

orbital17:


packages/core/src/spec/lib/features/FileHandlingFeatureSpec.ts
```diff
@@ -0,0 +1,44 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {FileHandlingFeature} from '../../../lib/features/FileHandlingFeature';
+
+describe('FileHandlingFeature', () => {
+  describe('#constructor', () => {
+    // variables is the only field dynamically generated.
+    it('Generates correct variables for application', () => {
```
gstepniewski-google, line 22: I would probably consider adding a couple of tests for invalid/incomplete manifest entries (missing or invalid mime type, missing url, url with bad scheme/origin/scope, invalid url).

orbital17: Thanks, added more tests for web manifest processing to a new test file. This file just checks for a correct template generation from the twaManifest, so no need for checking validation here

gstepniewski-google (approved):

Issue #901, opened 2024-12-23
Display content behind Android notch


korywka:
Hello. There is a lot of recipes how to launch PWA in full screen mode.

Some meta tags, `?fullscreen=true` parameters for `start_url` and so on.

But they doesn't really work as expected.

The goal is to show PWA as Google maps, to take the view behind notch:

<img src="https://github.com/user-attachments/assets/b52a5bac-1e15-4682-96c0-986af5733132" width="400">

The `standalone` mode leaves space at the top and at the bottom:

<img src="https://github.com/user-attachments/assets/12e2a6af-76ce-4b76-8e6f-ede1a0f3640e" width="400">

The `fullscreen` mode removes the space at the bottom, but the top status bar becomes **black**.

All "[look ma, my PWA runs fullscreen](https://pwadisplaytest.tecdrop.com/fullscreen/)" examples on the net also do not stretch behind notch.

Does someone managed to do it on latests devices (pixel 7a, 8a) and bubblewrap?

arpu:
@ukaratkevich  anything we can do to get this in chrome?

I-am-Pritam-20:
@korywka you have mentioned that the bottom area gets stretched but it will not, try with 3 button navigation enabled on Android. You will see it will behave the same like the status bar.

PR #947, opened 2025-07-24
Add display override to TWA manifest

adamraine:
This handles the `display_override` web manifest fields and adds support for `window-controls-overlay` and `tabbed` display modes.

🔒 b/433357939

### DIFF
--- a/packages/core/src/lib/types/WebManifest.ts
+++ b/packages/core/src/lib/types/WebManifest.ts
@@ -35,6 +35,9 @@ export interface WebManifestShortcutJson {
 
 type WebManifestDisplayMode = 'browser' | 'minimal-ui' | 'standalone' | 'fullscreen';
 
+export type WebManifestDisplayOverrideValue = 'window-controls-overlay' | 'tabbed' | 'browser' |
+    'minimal-ui' | 'standalone' | 'fullscreen';
+
 export type OrientationLock = 'any' | 'natural' | 'landscape'| 'portrait' | 'portrait-primary'|
     'portrait-secondary' | 'landscape-primary' | 'landscape-secondary';
 
@@ -64,6 +67,7 @@ export interface WebManifestJson {
   start_url?: string;
   scope?: string;
   display?: WebManifestDisplayMode;
+  display_override?: WebManifestDisplayOverrideValue[];
   theme_color?: string;
   background_color?: string;
   icons?: Array<WebManifestIcon>;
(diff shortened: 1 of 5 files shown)


adamraine:
cc @PEConn

PEConn:
Code LGTM, but as this code is now owned by the KRK team you'll want to get an approval from one of them (eg, @OlegNitz or @orbital17 ).

packages/core/src/lib/types/WebManifest.ts
```diff
@@ -35,6 +35,9 @@ export interface WebManifestShortcutJson {
 
 type WebManifestDisplayMode = 'browser' | 'minimal-ui' | 'standalone' | 'fullscreen';
 
+export type WebManifestDisplayOverrideValue = 'window-controls-overlay' | 'tabbed' | 'browser' |
+    'minimal-ui' | 'standalone' | 'fullscreen';
```
gstepniewski-google, line 39: I may be missing something obvious, but why is 'fullscreen-sticky' not defined here? I know it's added in TwaManifest.ts but I can't fully understand the reason...

adamraine: It wasn't defined on `WebManifestDisplayMode` either, and it's not listed as a valid web manifest value for `display_override`:
https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/display_override

My understanding is that `fullscreen-sticky` is specific to TWAs and so we should define it in the TWA manifest but not the web manifest.

gstepniewski-google: I see. So how would someone use the `fullscreen-sticky` override in their TWA? Would they have to manually add it to the twamanifest.json?

gstepniewski-google:
@ukaratkevich mind taking a final look, since you were involved in the design of this?

ukaratkevich (approved):
LGTM.

korywka:
@adamraine Sorry for the ping. Should this PR fix this issue? https://github.com/GoogleChromeLabs/bubblewrap/issues/901

I updated bubblewrap and added this to the TWA:
```
"displayOverride": [
  "window-controls-overlay"
]
```
But the notch area is still black. Is `window-controls-overlay` related to the notch area?

korywka:
@ukaratkevich Thank you very much for the clarification 🙏

Issue #1047, opened 2026-07-21
App must target Android 16 (API level 36) or higher


Stevie-Ray:
First of all thanks for creating Bubblewrap, it's a shame its not as actively maintained as in was the past!

**Describe the bug**
From Aug 31, 2026, if your target API level is not within 1 year of the latest Android release, you won't be able to update your app.

Same for: App must use Google Play Billing Library version 8.0.0 or later

**To Reproduce**
1. Go to 'https://play.google.com/console'
2. Click on 'notifications'
3. See error

**Screenshots**

<img width="378" height="79" alt="Image" src="https://github.com/user-attachments/assets/2f2fef8e-036f-42f2-9fdb-b04f9b862fb4" />
<img width="387" height="64" alt="Image" src="https://github.com/user-attachments/assets/9a908408-1b88-4da2-a132-5371b52de2e4" />

tunapiq:
We have the same issue. We just got email notifications from Google.

EzekielaRakotoarijaona:
Hello same issue here. And we only have one month to upgrade.

JudahGabriel:
I can repro the issue with my own apps as well.

JudahGabriel:
For frame of reference, here's a previous PR that bumped the Android API level version. Pretty simple: https://github.com/GoogleChromeLabs/bubblewrap/pull/764/changes

Ubeydu:
I'm working on a PR for this. The scope on current main is smaller than #764: commit e0af5ad already bumped compileSdkVersion to 36 (with AGP 8.9.1, Gradle 8.11.1, JDK 17), so what remains is targetSdkVersion 35 → 36 in the template's app/build.gradle, plus the BUILD_TOOLS_VERSION constant in AndroidSdkTools.ts (34.0.0 → 36.1.0, same pairing as #764 and #862). The specs interpolate that constant, so no test changes are needed.

I'll validate with the repo's build/lint/test, then build a real production TWA with the patched CLI and upload the AAB to Google Play to confirm it accepts targetSdk 36. PR shortly.

One note for the thread: the pinned android-browser-helper (2.6.2) predates the edge-to-edge splash fixes in 2.7.x, but bumping it raises the library minSdk to 23, which is breaking for minSdk 21 projects. That looks like a separate PR; this one stays minimal.

thequailman:
Are there any plans to also bump the billing library up to 8.0?

Ubeydu:
[...]

PR #862, opened 2024-07-10
Support Android SDK version 35, with many consequential updates

A1kmm:
Fix npm install - dependency changes meant that bubblewrap wouldn't compile due to dependencies not supporting previous TypeScript version. To address this, updated TypeScript to 5.5.3.

Updating TypeScript necessitates updating inquirer. Inquirer has been re-written in version 10, and version 9 is ESM only, while other dependencies need it to be CommonJS. So inquirer was updated to 8.2.6 for now - the latest CommonJS supporting version with the legacy API. As part of this change, esModuleInterop has been turned on, and the module import style has been made consistent, avoiding type errors.

A few other minor changes (particularly to caught errors) were needed for TS5.5.3 compat.

Next, the main aim of this change updates the targetSdkVersion and compileSdkVersion in app/build.gradle in the template. However, this also requires adding the namespace (set to the package ID) to the template.

Updating the compileSdkVersion requires bumping the Android Gradle plugin to the latest 8.5.0.

This change makes it no longer compatible with JDK 11, so this requires another extensive change - updating to OpenJDK 17. This has a number of consequential documentation and string updates, including moving the installer to use Eclipse Temurin instead of AdoptOpenJDK.

Finally, updating JDK requires updating the Gradle wrapper. The outputs are from running `gradle wrapper --gradle-version 8.8`.

PR #764, opened 2023-02-14
Fix #750 Bump compileSdkVersion and targetSdkVersion to 33

FluorescentHallucinogen:

PR #1050, opened 2026-07-22
Bump targetSdkVersion to 36

A1kmm:
Everything tests okay with a simple PWA after upgrading, and passes Google Play's Quick Checks.

35 is going to be unsupported for publishing new versions to Google Play from the end of August according to https://support.google.com/googleplay/android-developer/answer/11926878.


Ubeydu:
I was working on the same bump when this PR appeared: I had claimed it in #1047 earlier today, and I opened #1051 with my version shortly after this one. Rather than compete, I want to add my test results here, because they verify the exact template change this PR makes.

My branch carries the identical targetSdkVersion 36 line, plus a BUILD_TOOLS_VERSION bump. With that branch:

- npm build, lint and test are green on all three packages.
- I generated the TWA for my production PWA (com.akluma.app, currently in closed testing on Google Play) with the patched CLI, and the generated project gets targetSdkVersion 36.
- aapt2 badging on the built APK reports targetSdkVersion 36, and the signature verifies against the app's existing upload key.
- The app installs and runs on an Android 16 (API 36) emulator: full-screen launch, with Digital Asset Links verification working.

One question for the maintainers: #764 and #862 both paired the target bump with a BUILD_TOOLS_VERSION bump (39d5972). Leaving it at 34.0.0 works, since it only drives the sdkmanager install and the zipalign/apksigner paths, but if the pairing is worth keeping, that companion change is in #1051.

JudahGabriel:
@gstepniewski-google Can we get a review on this? It's rather time-sensitive with the new Google Play requirements.

leeobrum:
I am eagerly awaiting this PA

gstepniewski-google:
Merged #1051

JudahGabriel:
Thanks Greg!

ghnp5:
Hi @gstepniewski-google 

Do you confirm that this change is really included in the last release?

When I use the latest Docker container (`ghcr.io/googlechromelabs/bubblewrap:latest`, after pulling it),  
and then run `bubblewrap init`,

the result is still `targetSdkVersion 35` in `build.gradle`.

I know I can update it manually, but I was expecting bubblewrap to do it for me.

Thanks!

ghnp5:
Maybe this is the reason: https://github.com/pwa-builder/PWABuilder/issues/6159#issuecomment-5185270338

But that means that whenever the NPM version is released, the docker image needs to be updated too, so it can pull the latest npm package.

JudahGabriel:
@ghnp5 No, these are separate issues. PWABuilder doesn't rely on the Docker container for Bubblewrap. We have our own container.

